Issue #40, opened 2021-02-18
NFT Metadata JSON


sam-at-luther:
A key accelerator to the adoption of the ERC721 [0] Non Fungible Tokens (NFT) standard for the Ethereum network is the "ERC721 Metadata JSON Schema" definition. This schema is part of the optional metadata extension of that standard, and provides a guidelines for the format of the data that served for a token's metadata URI. This standard is how the plethora of wallets in that ecosystem (for example OpenSea [1] and d'cent [2]) display asset attributes for many ERC721 implementations, including an image for each NFT. 

Building on the success of that standard, I propose _optionally_ adopting this same JSON schema standard for NFT metadata on HTS. Specifically,

```
{
    "title": "Asset Metadata",
    "type": "object",
    "properties": {
        "name": {
            "type": "string",
            "description": "Identifies the asset to which this NFT represents"
        },
        "description": {
            "type": "string",
            "description": "Describes the asset to which this NFT represents"
        },
        "image": {
            "type": "string",
            "description": "A URI pointing to a resource with mime type image/* representing the asset to which this NFT represents. Consider making any images at a width between 320 and 1080 pixels and aspect ratio between 1.91:1 and 4:5 inclusive."
        }
    }
}
```

I also would like to solicit feedback on how the community thinks is best to include URIs that serve metadata in the above format within HTS NFTs. 

Based on these discussions, I'm eager to draft a HIP and be a champion for such a standard.

[0] https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md#specification
[1] https://opensea.io/
[2] https://dcentwallet.com/

rahul-kothari:
for images/videos - probably use IPFS to ensure authenticity? Pinata is also another interesting idea!

cylon56:
Here is a potential add-on to this proposal or a separate HIP for NFTs. **HTS should support the creation of numerous non-fungible tokens of a single type**. 

Currently, it seems that NFTs can only work on HTS if you create a token with a supply of one and treat it as an NFT. This seems like a workaround for now and leaves a lot to be desired. In addition to having embedded metadata, HTS should also allow users to create a single class of NFTs that can each be recognized as the same type while still being individually distinct tokens.

rahul-kothari:
At this point, it might be easier to just create an HNFTS (Hedera Non-Fungible Token Service).

sam-at-luther:
[...]

PR #41, opened 2021-02-18
Add HTS NFT Metadata specification

sam-at-luther:
This specification provides a standard way to interrogate Non-fungible Tokens (NFT) on HTS for associated metadata. For example, it provides a standard way for HTS token explorers to display an image associated with a NFT on HTS, in the same way ERC721 images are displayed on the Ethereum network.

See https://github.com/hashgraph/hedera-improvement-proposal/issues/40 for related discussion.

### DIFF
--- a/hip-0000-hts-nft-metadata.md
+++ b/hip-0000-hts-nft-metadata.md
@@ -1,0 +1,117 @@
+---
+hip: 0000
+title: NFT Metadata JSON Schema
+author: Sam Wood sam.wood@luthersystems.com, Susan Chan susan.chan@luthersystems.com, Stephanie Yi stephanie.yi@luthersystems.com, Khoa Luong khoa.luong@luthersystems.com
+type: Standards Track
+category: Application
+status: Draft
+created: 2020-2-18
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/issues/40
+updated:
+requires:
+replaces:
+superseded-by:
+---
+
+## Abstract
+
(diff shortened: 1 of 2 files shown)


kenthejr:
To be clear, this proposed schema can be applied today with no changes to the underlying HTS, correct? It's my understanding that you are defining two things: 1) the inclusion of a URI (in the Name or a new URI field), and 2) the schema for the meta data being pointed to. Is that correct?

sam-at-luther:
Correct. This proposal is purely at the application level and does not suggest a change in HTS.

kenthejr (dismissed):

donaldthibeau (commented):
In the upcoming mainnet release we are including a new memo field on the token entity. This field would seem a better fit for the reference data than the name. Could we update the HIP to reference that memo field as an option as well?

https://github.com/hashgraph/hedera-services/releases

sam-at-luther:
Agreed that memo sounds like a more appropriate field for this data. I will update the proposal and the example to use memo.

To clarify, is this the same memo field that is set via:
https://github.com/hashgraph/hedera-sdk-go/blob/master/token_create_transaction.go#L416
`SetTransactionMemo` on CreateTokenTransaction?

donaldthibeau:
> Agreed that memo sounds like a more appropriate field for this data. I will update the proposal and the example to use memo.
> 
> To clarify, is this the same memo field that is set via:
> https://github.com/hashgraph/hedera-sdk-go/blob/master/token_create_transaction.go#L416
> `SetTransactionMemo` on CreateTokenTransaction?

No, this would be a memo on the token entity vs on the transaction.

donaldthibeau (approved):

kenthejr (approved):

donaldthibeau (approved):

PR #45, opened 2021-03-10
Zero Knowledge Proofs to Hedera

rahul-kothari:
Since there are a LOT of coding/design discussions that must be had (involving cryptographers within Hedera), the HIP doesn't have all technical details yet.

I look forward to starting the discussion around this. 

Some of the use cases:
* It would be awesome to see people create SNARKS not on solidity but on Java/Rust etc. using the Hedera SDKs and still get superior speeds!
* Have private tokens/topics
* Ability for appnets to prove they are using HCS properly without having to reveal all their data

### DIFF
--- a/HIP/hip-12.md
+++ b/HIP/hip-12.md
@@ -1,16 +1,12 @@
 ---
-hip: <HIP number (this is determined by the HIP editor)>
+hip: 12
 title: Zero Knowledge Proofs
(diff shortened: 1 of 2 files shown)


donaldthibeau (dismissed):

kenthejr (approved):
This

PR #55, opened 2021-03-13
hip-0000-hedera-name-service-draft

H-Bart:
HIP Draft for the Hedera Name Service (HNS)


kenthejr (changes requested):
Please rename the file to `hip-13.md` and we should be good to merge.

kenthejr (approved):

PR #60, opened 2021-03-25
HIP: Token Holds / Held Balance

Ferparishuertas:
Added a new HIP hip-0000-holds.md

This HIP defines and discusses a mechanism, new way of executing transaction on a held balance, ensuring that pre-agreed settlement scenarios, have the needed collateral.


Ferparishuertas:
@kenthejr I've recreated the PR but the checkbox hasn't shown up. Strange..

PR #73, opened 2021-04-23
HIP 17 - HTS NFT support 

Daniel-K-Ivanov:
**Details**
- Proposal for extending the HTS service with support for NFTs
- Consolidates IWA specification and HTS API
- The proposed API aligns with the pending NFT design and specification referring to `MerkleInvertibleMap`

### DIFF
--- a/HIP/hip-0000.md
+++ b/HIP/hip-0000.md
@@ -70,6 +70,12 @@ The following protobuf messages to be changed.
 #### TokenCreateTransactionBody
 
 ```diff
++//Fungible or Non-Fungible Token Base. Cannot be updated using admin key
++enum TokenType {
++	FUNGIBLE = 0;
++	NON_FUNGIBLE = 1;
++}
+
 message TokenCreateTransactionBody {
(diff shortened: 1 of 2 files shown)


corviato1:
There is a hip-17 in the pipeline that predates this one.

donaldthibeau (approved):

Daniel-K-Ivanov:
> There is a hip-17 in the pipeline that predates this one.

I wasn't sure whether we are counting only the merged ones or the pending ones as well. So should it be 17 or 18?

kenthejr:
> There is a hip-17 in the pipeline that predates this one.

HIP numbers are assigned by editors when merged. This one will most likely become HIP 17. I added a comment to your proposal.

kenthejr (commented):
This looks great. Can we get the meta-data updated? The `type` property should be "Standards Track" & `category` should be "Service".

Other than that, it looks ready to merge.

kenthejr (approved):
Thank you. It looks good.

PR #76, opened 2021-04-27
Improved HIP-1 and added table of HIPs to README

kenthejr:
HIP-1 Updates:

- Clarified the desire for granular HIPs.
- Clarified the use of Discussions to vet ideas before generating HIPs
- Require spec & reference implementation for Final consideration

### DIFF
--- a/HIP/hip-1.md
+++ b/HIP/hip-1.md
@@ -22,7 +22,7 @@ HIPs are not meant to address *bugs* in implemented code. Bugs should be address
 
 There are three kinds of HIP:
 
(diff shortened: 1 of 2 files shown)


kenthejr:
@rahul-kothari does this address your feedback?

coltsfanatic07:
I like it. Large HIPS will result in never getting ideal requirements or anything near consensus for a path forward. Being agile is another benefit for the overall solution.

rahul-kothari:
Yea looks good!

donaldthibeau (approved):

PR #115, opened 2021-07-13
Proposal for Transaction Signing Protocol

0xJepsen:


HIP/hip-0000-Transaction-Signing-Protocol.md
```diff
@@ -0,0 +1,163 @@
...
+    //required publickey = 3; //potential auth, would also need a ECDSA
+    //signed object
+    optional Capabilities = 4; // What can this wallet do? Limitations 
+    optional Memo = 5;
+}
+
+message LimitResponse{
+    required ResponseCode response = 1;
+    required MaxTransactionFee = 2;
+    required PreAuthorizationLimit = 3;
```
steven-sheehy, line 53: This looks to be written with proto2. Hedera uses proto3 for all of its protobufs.

HIP/hip-0000-Transaction-Signing-Protocol.md
```diff
@@ -0,0 +1,163 @@
...
+    required int32 PreAuthorizationLimit = 2; //maybe we make these types bigger to prevent overflow
+    //required publickey = 3; //potential auth, would also need a ECDSA
+    //signed object
+    optional Capabilities = 4; // What can this wallet do? Limitations 
+    optional Memo = 5;
+}
+
+message LimitResponse{
+    required ResponseCode response = 1;
+    required MaxTransactionFee = 2;
```
steven-sheehy, line 52: Most fields are missing the field name

HIP/hip-0000-Transaction-Signing-Protocol.md
```diff
@@ -0,0 +1,163 @@
...
+* `Submit` provides the requesting entity with a response containing information of a signed transaction. 
+* `SignAndSubmit` allows for both the sign and submit actions to be carried out in half of the number of responses and requests. 
+
+These calls allow for a single transaction to be signed by multiple keys for multi signature protocols.
+
+## Specification
+
+```
+message LimitsRequest{
+    required int32 MaxTransactionFee  = 1; //maybe we make these types bigger to prevent overflow
```
bugbytesinc, line 42: Perhaps this type should match what a TX max transaction fee is for a `TransactionBody`, that would make it `uint64`?

HIP/hip-0000-Transaction-Signing-Protocol.md
```diff
@@ -0,0 +1,163 @@
...
+    required int32 PreAuthorizationLimit = 2; //maybe we make these types bigger to prevent overflow
```
bugbytesinc, line 43: This could end up being more complex if we were to consider custom tokens as a Gas instead of hBars

PR #116, opened 2021-07-14
Transaction-Signing-Protocol.md

0xJepsen:
Open to community feedback

### DIFF
--- a/HIP/Transaction-Signing-Protocol.md
+++ b/HIP/Transaction-Signing-Protocol.md
@@ -1,0 +1,163 @@
+- hip: XX
+- title: Hedera Signing Protocol
+- author: [0xJepsen](https://github.com/0xJepsen), [rocketmay](https://github.com/rocketmay), [bugBytes](https://github.com/bugbytesinc)
+- type: Standards Track
+- category: Application
+- status: Draft 
+- created: <07/03/21>
+- discussions-to: [here](https://github.com/hashgraph/hedera-improvement-proposal/discussions/98)
+- updated: <07/12/21> 
+- requires: <HIP number(s)>
+- replaces: <HIP number(s)>
+- superseded-by: <HIP number(s)>
+
+## Abstract
+
+This specification proposes a protocol for exchanging unsigned transactions and signatures.
+
+
+## Motivation
+
+Developers are currently creating web authentication mechanisms from scratch for Hedera-based web apps. This limits consumer adoption and results in a poor user and developer experience. A standard protocol for decentralized applications to communicate with clients and allow clients to sign transactions would significantly improve the developer and user experience. The Hedera development space is still young, with many new projects developing promising ideas. These projects will need a standardized method of communication, much like what exists in the Ethereum space.
+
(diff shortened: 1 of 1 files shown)


0xJepsen (commented):
Was discussing this with @bugbytesinc and I think he has a good point here. 

'thinking the protocol for Send and SignAndSend should return the TransctionGetReceiptResponse --- because if really high thruput is required, all the "originating system" is going to ask for is a signature, then they'd manage the rest on their side -- so forget that SendForPrecheckOnly idea.'
'When I say return TransactionGetReceiptResponse I'm really meaning the protobuf encoded bytes of it.  But still parsing out the Response Code might be nice.
It is also possible the transaction would not pass precheck, in which case the bytes would be empty. Because the wallet would have not gotten to the point where it could ask for a receipt because none would be created. '

rocketmay:
Agree, the receipt should be sent back as the response for calls which result in the receiving entity submitting anything to HAPI. I don't think we need to specifically parse out the Response Code as part of the protocol.

bugbytesinc:
> Agree, the receipt should be sent back as the response for calls which result in the receiving entity submitting anything to HAPI. I don't think we need to specifically parse out the Response Code as part of the protocol.

Perhaps "Response Code" is not a good name, its more of a response for this protocol (not the HAPI response code)...like "I refuse to sign" or "You exceeded my limits" or "LGTM".

sergmetelin:
> What is the difference between this and scheduled transactions. Rather, why are scheduled transactions not able to be used to solve this issue?

As far as I understand this HIP, it's more so for local communication between an application and a wallet, rather than a network-level communication. Using scheduled transactions would make the flow unnecessarily more complex (each transaction would require a co-signature from the app key) and more expensive.

0xJepsen:
@danielakhterov What Sergey said is correct. If you were to use scheduled transactions for this, the decentralized application would always have to pay for the scheduled transaction without guaranteeing that its users would fulfill it. When a user uses a decentralized protocol, they are supposed to be responsible for that transaction's transaction fees. The new smart contract service may change this. Also, I should rewrite it in JSON.

cacampbell:
You can already use an asynchronous transaction signer for local devices (with the SDK)

gregscullard:
Note this other PR which proposes something similar based on existing standards
https://github.com/hashgraph/hedera-improvement-proposal/pull/179

gregscullard:
I'm not sure it's up to the application to suggest to the browser wallet what limits should be. This could open the door to malicious behaviour where an application acquires privileges from an unwary user who may not read the message properly and approve by mistake.

I would propose a separate HIP for wallets to implement "convenience features" independently, e.g. allow a user to specify that they approve an app to spend up to x amount over a period of time, or approve the next 5 requests (up to an amount) by an app automatically.
These approvals would not be known to the app which would not be able to exploit them, they would merely be enabled by the user who doesn't want to be nagged every time they click on a link triggering a micropayment while reading article for example.

sergmetelin:
@0xJepsen do you think we should merge this one with https://github.com/hashgraph/hedera-improvement-proposal/pull/179? Do you mind adding relevant feedback there and potentially closing this one?

0xJepsen:
@sergmetelin absolutely, that makes the most sense to me.

PR #134, opened 2021-09-08
Various formatting updates to HIP-23

sergmetelin:
This PR includes various formatting updates

### DIFF
--- a/HIP/hip-23.md
+++ b/HIP/hip-23.md
@@ -1,7 +1,7 @@
 * hip: 23
 * title: Make the need for token association opt-in
 * author: Gerbert Vandenberghe (gerbert-vandenberghe)
-* type: Standard Track - Service
+* type: Standards Track - Service
 * status: Draft
 * created: 2021-06-22
 * discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/107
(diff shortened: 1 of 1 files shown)


gregscullard (changes requested):
`In addition, I can also decrease the number of associations, and that call will succeed if the new (decreased) number that I have specified is greater than or equal to the current number of automatic associations that I have already used.`

I can think of a corner case where I’ve allowed 100 associations, used them (associationCount=100) and then manually associate with 2 more. If I want to decrease the number of automatic association, how will dissociating from a token reduce the association count ? Will it only do so when I have removed two tokens (which may have been associated automatically) and start dissociating other tokens (could be the last 2 I associated with). Or do we keep track of which tokens were automatically associated and only dissociating from these decrements the counter ?

gregscullard:
The intro (and some of the text) refers to wallet or wallets where I think account would be preferable (happy to proceed without change)

sergmetelin:
> `In addition, I can also decrease the number of associations, and that call will succeed if the new (decreased) number that I have specified is greater than or equal to the current number of automatic associations that I have already used.`
> 
> I can think of a corner case where I’ve allowed 100 associations, used them (associationCount=100) and then manually associate with 2 more. If I want to decrease the number of automatic association, how will dissociating from a token reduce the association count ? Will it only do so when I have removed two tokens (which may have been associated automatically) and start dissociating other tokens (could be the last 2 I associated with). Or do we keep track of which tokens were automatically associated and only dissociating from these decrements the counter ?

My understanding is that the number of auto-associations will be tracked independently of the number of manual associations.

gregscullard (approved):

gregscullard (approved):

PR #213, opened 2021-11-09
hip-applicationSigns-draft - add initial files

justin-atwell:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


gregscullard (changes requested):
line 65, maybe add a comment to indicate the client's operator and key don't need to be set since the transaction is fully signed, the client only acts as "transport" here.

gregscullard (changes requested):
Line 60, maybe add comment to indicate the transaction should be "shipped" to the other user for signing using a REST api or other protocol.

gregscullard:
We probably should include how a transaction can be signed using a signer which exemplifies how an external "process" can sign a transaction on behalf of the application.

sergmetelin (approved):
LGTM

gregscullard (approved):
LGTM

PR #217, opened 2021-11-12
Sam at luther/hip 10 uri refinement

sam-at-luther:
**Description**:

This PR modifies HIP-10 in order to support 
* NFT token type metadata
* CAIDs via HIP-30
* versioning


sergmetelin (approved):
LGTM

PR #232, opened 2021-11-20
Consider Expanding the Allowed Characters for HTS Ticker Symbols

scalemaildev:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-0000-template.md
+++ b/HIP/hip-0000-template.md
@@ -1,0 +1,45 @@
+---
+hip: <HIP number (this is determined by the HIP editor)>
+title: Expand Allowed Characters for HTS Ticker Symbols
+author: John Conway (@scalemaildev)
+type: Standards Track
+category: Core
+needs-council-approval: Yes
+status: Draft
+created: 2021-11-19
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/231
+---
+
+## Abstract
+
+Allow some symbols and non-capitalized alpha characters to be used in HTS ticker symbols.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


HIP/hip-0000-template.md
```diff
@@ -0,0 +1,45 @@
...
+category: Core
+needs-council-approval: Yes
+status: Draft
+created: 2021-11-19
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/231
+---
+
+## Abstract
+
+Allow some symbols and non-capitalized alpha characters to be used in HTS ticker symbols.
```
sergmetelin, line 15: Have you tried creating symbols with symbols? 
both `name` and `symbol` can be any Unicode sequence at all
the only restriction is that the UTF-8 encoding of this Unicode cannot contain the 0 bytes (`NUL`)

scalemaildev: I have not but when reading through the SDK documentation [here](https://docs.hedera.com/guides/docs/sdks/tokens/define-a-token) it states:

```
The publicly visible token symbol. It is UTF-8 capitalized alphabetical string identifying the token. The token symbol is not unique. Maximum of 100 characters.
```

Do you know if this is still true? I can do some tests later myself to find out.

sergmetelin: https://docs.hedera.com/guides/docs/sdks/tokens/define-a-token - we've updated documentation now to clarify!

sergmetelin: If this addresses the functionality in question in this HIP - feel free to close the PR!

PR #249, opened 2021-11-26
Make HIP-222 more explicit

tinker-michaelj:
**Description**:
- Clarify that the proposed signing algorithm is `keccak256`-with-ECDSA.
- Clarify that the proposed encoding for signatures is the "raw" concatenation of the `(r, s)` values.


sergmetelin (approved):
LGTM

PR #329, opened 2022-01-14
Support `CREATE2` opcode

tinker-michaelj:
**Description**:
- Proposes to support the `CREATE2` EVM opcode on the Hedera smart contract service.


sergmetelin (approved):
LGTM

PR #335, opened 2022-01-19
Update hip-227 response JSON

Nana-EC:
Signed-off-by: Nana-EC <nana.essilfie-conduah@hedera.com>

**Description**:
Update Contract Results log response `/api/v1/contracts/{id}/results/logs` with missing values

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-227.md
+++ b/HIP/hip-227.md
@@ -71,13 +71,16 @@ The following JSON represents a typical response result from either of these cal
 
 Optional filters
 
-- `limit`
-- `order`
-- `timestamp`
-- `topic0`
-- `topic1`
-- `topic2`
-- `topic3`
+- `index`: Contract log index
+- `limit`: The maximimum number of items to return in response.
+- `order`: The order in which items are listed - asc (default) or desc.
+- `timestamp`: The consensus timestamp in seconds.nanoseconds format. Required when a topic is specified and should not span a time range of more than seven days.  The following optional comparison operators can be added - :gt(e), :lt(e), :eq.
+- `topic0`: The first topic associated with a contract log. Topic should be hex encoded with `0x` characters optional.
+- `topic1`: The second topic associated with a contract log. Topic should be hex encoded with `0x` characters optional.
+- `topic2`: The third topic associated with a contract log. Topic should be hex encoded with `0x` characters optional.
+- `topic3`: The fourth topic associated with a contract log. Topic should be hex encoded with `0x` characters optional.
+
+> **_NOTE:_** Paging will not be supported if number of logs exceed the page size (default 25), optional parameters will need to be adjusted by callers in this case.
 
 ## Backwards Compatibility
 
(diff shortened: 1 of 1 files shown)


HIP/hip-227.md
```diff
@@ -71,7 +74,6 @@ Optional filters
 - `limit`
 - `order`
 - `timestamp`
```
steven-sheehy, line 76: We should document what values are supported like the comparison operators or if it's hex-encoded and `0x` is optional, etc.

HIP/hip-227.md
```diff
@@ -71,7 +74,6 @@ Optional filters
...
-- `address`
 - `topic0`
```
steven-sheehy, line 77: We should mention able the idiosyncrasies around this API including that timestamp is required if a topic is supplied. Also that timestamp must have a to and from within 7 days of each other. And that there's no paging so users will need to adjust their parameters on their own if more than page size results.

steven-sheehy (changes requested):

HIP/hip-227.md
Nana-EC: Added

HIP/hip-227.md
Nana-EC: Noted, quite a few for this endpoint

HIP/hip-227.md
```diff
@@ -52,30 +55,32 @@ The following JSON represents a typical response result from either of these cal
...
 
-- `limit`
-- `order`
-- `timestamp`
-- `address`
-- `topic0`
-- `topic1`
-- `topic2`
-- `topic3`
+- `index`: Contract log index
```
steven-sheehy, line 74:
```suggestion
- `index`: The unique log index within a particular contract result. Starts at zero and increases.
```

steven-sheehy (changes requested):

HIP/hip-227.md
```diff
@@ -52,30 +55,32 @@ The following JSON represents a typical response result from either of these cal
...
+- `limit`: The maximimum number of items to return in response.
+- `order`: The order in which items are listed - asc (default) or desc.
```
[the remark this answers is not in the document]
Nana-EC: Updated

HIP/hip-227.md
```diff
@@ -52,30 +55,32 @@ The following JSON represents a typical response result from either of these cal
...
+- `timestamp`: The consensus timestamp in seconds.nanoseconds format. Required when a topic is specified and should not span a time range of more than seven days.  The following optional comparison operators can be added - :gt(e), :lt(e), :eq.
```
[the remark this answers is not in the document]
Nana-EC: Updated

HIP/hip-227.md
Nana-EC: Updated

steven-sheehy (approved):
LGTM

PR #338, opened 2022-01-20
feat: add hip for signature and wallet providers

danielakhterov:


HIP/hip-100.md
```diff
@@ -0,0 +1,479 @@
...
+<summary>abstract class Signer</summary>
+  
+### Methods
+
+##### `async signMessage` ( `messages`: `List` < `bytes` > ): `List` < `List` < `SignerSignature` > >
+
+Sign a list of messages
+
+**NOTE**: Each element in the outer list of the result is all the signatures for the
+at the same index.
```
shemnon, line 36:
```suggestion
**NOTE**: Each element in the outer list of the result is all the signatures for the
message at the same index.
```

HIP/hip-100.md
```diff
@@ -0,0 +1,479 @@
...
+
+---
+
+##### `async signTransaction` ( `transaction`: `Transaction` )
+
+Signs the transaction
+
+**NOTE**: Use `Transaction.getSignatures()` to see the actual signatures.
+at the same index.
```
shemnon, line 45:
```suggestion
```

HIP/hip-100.md
```diff
@@ -0,0 +1,479 @@
...
+
+---
+
+##### `getPublicKey` (): `PublicKey`
+
+Return the public key associated with this wallet.
+
+---
+
+##### `async createRandomED25519` (): `Wallet`
```
shemnon, line 179: Should this be static?

HIP/hip-100.md
```diff
@@ -0,0 +1,479 @@
...
+
+Creates a wallet with a new ED25519 key
+
+**NOTE**: This would create an alias key account on Hedera
+
+---
+
+##### `async createRandomECDSA` (): `Wallet`
```
shemnon, line 187: Should this be static?

HIP/hip-100.md
```diff
@@ -0,0 +1,479 @@
...
+<details>
+<summary>class LocalWallet implements Wallet</summary>
+
+### Static Methods
+
+### Constructor
+
+##### `constructor`()
+
+Creates an `LocalWallet` from the environment variables `OPERATOR_KEY`, `OPERATOR_ID`, and `HEDERA_NEWTORK`
```
shemnon, line 220: Can we get a variant with explicit args as well?

HIP/hip-100.md
```diff
@@ -0,0 +1,479 @@
...
+needs-council-approval: Yes
+status: Draft
+created: 2022-01-20
+discussions-to: TBD
+updated: 2022-01-20
+---
+
+## **Abstract**
+
+Ethereum like signature and wallet providers with the Hedera SDKs
```
shemnon, line 16: All Hedera SDKs or just the JS sdk?  Go and Java don't really map to some JS concepts well, would those be in another HIP?

HIP/hip-100.md
```diff
@@ -0,0 +1,479 @@
...
+  
+### Methods
+
+##### `async signMessage` ( `messages`: `List` < `bytes` > ): `List` < `List` < `SignerSignature` > >
+
+Sign a list of messages
+
+**NOTE**: Each element in the outer list of the result is all the signatures for the
+at the same index.
+
```
gregscullard, line 37: Presume these are arbitrary data payloads, not necessarily transactions or even "messages" ?
Suggest renaming to `signData` or just `sign` since the data could be anything and we have an explicit `signTransaction` method.

HIP/hip-100.md
```diff
@@ -0,0 +1,479 @@
...
+---
```
gregscullard, line 38: Return should include signing account id if applicable (this enables the application using this call to verify the signature against a known public key)

HIP/hip-100.md
```diff
@@ -0,0 +1,479 @@
...
+
+**NOTE**: Unlike `Provider.sendTransaction()` this method will automatically wait for
+the receipt of the transaction.
+
+---
+
+##### `async checkTransaction` ( `transaction`: `Transaction` )
+
+Determines if all the properties requried are set and sets the transaction ID. If the transaction ID was already set it checks if the account ID of it is the same as the users.
+
```
gregscullard, line 61: Who sets the transaction Id ? The signer ?

HIP/hip-100.md
```diff
@@ -0,0 +1,479 @@
...
+---
+
+##### `async populateTransaction` ( `transaction`: `Transaction` )
+
+Sets the transaction ID of the transaction to the current account ID of the signer.
+
```
gregscullard, line 67: Suggest rename to `setTransactionId` for clarify. Although, the signer may also need to set the `node_account_id` to completely populate a transaction.
The SDK may not know the network the transaction is being sent to and therefore won't know the list of nodes

HIP/hip-100.md
```diff
@@ -0,0 +1,479 @@
...
+---
+</details>
+
+#### Example RPCs
+
+<details>
+
+<summary>class GrpcSignatureProvider extends Provider</summary>
+
+### Protobuf Defintions
```
gregscullard, line 277: Not sure we need these here. The SDK should only concern itself with the interfaces/abstract classes, not messaging.

HIP/hip-100.md
gregscullard: Not sure we should, if someone wants to do this, they should create their own implementation of `Wallet` and use that as a Provider/Signer. This would ensure the provider/signer pattern is adhered to.

gregscullard:
Signer/provider to optionally return a list of nodes (or node), particularly in the case of sign where the SDK submits to the network (not the external signer/provider). 
The node_account_id(s) may be set by the external signer/provider meaning the SDK needs to know the respective IP addresses for these.

Probably need to figure out a workflow for the above scenario(s) for documentation purposes, is it even desirable to have the SDK request a signature and send separately ?

gregscullard:
Need to add support for queries other than balance/info/records/receipt (contract local call, token info, etc...).
This should be a generic signQuery and sendQuery to match transaction patterns.

HIP/hip-100.md
gregscullard: Primary requirement is for the JS SDK to implement this, although the same principles should be extended to all SDKs eventually.

HIP/hip-100.md
```diff
@@ -0,0 +1,479 @@
...
+##### `getProvider` (): `Provider`
+
+Return the provider
+
+---
+
+##### `getPublicKey` (): `PublicKey`
+
+Return the public key associated with this wallet.
+
```
gregscullard, line 176: add `getAccountId` (): `AccountId`

HIP/hip-100.md
```diff
@@ -0,0 +1,479 @@
...
+---
+
+##### `async getAccountInfo` ( `accountId`: `AccountId` ): `AccountInfo`
+
+Get the info for an account
+
+---
+
+##### `async getTransactionReceipt` ( `transactionId`: `TransactionId` ): `TransacitonReceipt`
```
gregscullard, line 132: Typo `TransacitonReceipt`

HIP/hip-100.md
```diff
@@ -0,0 +1,479 @@
...
+
+### Fields
+
+##### `publicKey`: `PublicKey`
+
+The public key that signed this request
+
+##### `signature`: `bytes`
+
+The siganture for the message
```
gregscullard, line 207: Typo: `siganture`

HIP/hip-100.md
gregscullard, line 220: Typo "Creates an" -> "Creates a"

HIP/hip-100.md
```diff
@@ -0,0 +1,479 @@
...
+
+##### `getPublicKey` (): `PublicKey`
+
+Will return `null`
+
+---
+
+##### `async signTransaction` ( `transaction`: `Transaction` )
+
+Will do nothing
```
gregscullard, line 260: Should it not  throw an error as signing a transaction using a `VoidSigner` is not supported.

gregscullard:
This HIP probably should also explain the differences/relationships between signer and provider, maybe some pseudo-code examples

HIP/hip-100.md
danielakhterov: I was a bit conflicted here tbh. One one hand it does make sense for it to throw an error, but on the other handle I feel it also makes sense for it to not do anything. This would allow the SDK to simply use all signers to sign without checking to see if it should attempt to sign with a particular signer.

HIP/hip-100.md
danielakhterov: Hmm, signer setting the transaction ID doesn't make too much sense; perhaps this should be moved into `Wallet`?

HIP/hip-100.md
danielakhterov: Added `accountId: AccountId` to `SignerSignature`

HIP/hip-100.md
danielakhterov: Added a simple message about the primary concern being the JS SDK.

HIP/hip-100.md
danielakhterov: I agree with @gregscullard here. I think the base `LocalWallet` that is included in the SDK should use hard coded environment variables. Perhaps we can change these required ones to be `HEDERA_OPERATOR_KEY` and `HEDERA_OPERATOR_ID` to make it less likely these envs collide with other envs users might be using?

danielakhterov:
> Need to add support for queries other than balance/info/records/receipt (contract local call, token info, etc...). This should be a generic `signQuery` and `sendQuery` to match transaction patterns.

@gregscullard The idea behind this sounds fine, but unfortunately when it comes to `signQuery` and `sendQuery` it can't be supported at the moment for the following reasons:
 - the SDKs do not have stable `Query.[from|to]Bytes()` methods. (Technically JS has these methods, but that is by mistake).
 - Queries are not sign-able right now. https://github.com/hashgraph/hedera-sdk-js/issues/477
 - Unlike `sendTransaction` `sendQuery` could not exit because the output of the query is different per query.

HIP/hip-338.md
```diff
@@ -0,0 +1,460 @@
...
+
+### Methods
+
+##### `getProvider` (): `Provider`
+
+Return the provider
+
+---
+
+##### `getPublicKey` (): `PublicKey`
```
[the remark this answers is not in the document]
danielakhterov: Went with `getAccountKey(): Key` instead.

HIP/hip-338.md
```diff
@@ -0,0 +1,460 @@
...
+The difference between `Signer` and `Provider` is the `Signer` signs requests while `Provider` provides
+the network to which this request should be submitted to.
+
+<details>
+<summary>MyHbarWallet Example</summary>
+
+```typescript
+// Only accessible from within the browser environment after logging into MyHbarWallet
+class MyHbarWalletTestnetWallet {
+    private privateKey: PrivateKey;
```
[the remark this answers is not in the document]
danielakhterov: Not sure if resolved, but I put in some stubs on how secure storage could be used instead.

HIP/hip-338.md
```diff
@@ -0,0 +1,758 @@
...
+
+    constructor(accountKey: Key, accountId: AccountId, ledgerId: LedgerId) {
+        super();
+
+        this.accountKey = accountKey;
+        this.accountId = accountId;
+        this.provider = new MyHbarWalletProvider(ledgerId);
+    }
+
+    static async login(accountId: AccountId): Promise<MyHbarWalletSigner> {
```
danielakhterov, line 529: Maybe this should be optional `accountId`?

danielakhterov:
```Executable.executeWithSigner(Signer signer);

var signer = MyHbarWalletSigner.login();

new TransferTransaction()
	.addHbarTransfer("0.0.3", new Hbar(1))
	.addHbarTransfer("0.0.3", new Hbar(1).negated())
	.signWithSigner(signer);
[populateTransaction|signTransaction|sendTransaction]
[freeze|sign|execute]WithSigner(signer);```

gregscullard (approved):
LGTM - let's go

sergmetelin (approved):
LGTM

SimiHunjan (approved):

sergmetelin (approved):
LGTM

PR #358, opened 2022-02-09
HIP-0000 Token Create through HTS Precompiled Contract

stoqnkpL:
Signed-off-by: Stoyan Panayotov <stoyan.panayotov@limechain.tech>

**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/assets/hip-206/solidity/IHederaTokenService.sol
+++ b/assets/hip-206/solidity/IHederaTokenService.sol
@@ -49,6 +49,168 @@ interface IHederaTokenService {
         NftTransfer[] nftTransfers;
     }
 
+    /// Basic properties of a Hedera Token - name, symbol, memo, tokenSupplyType, maxSupply, 
+    /// treasury, freezeDefault. These properties are related both to Fungible and NFT token types
+    struct HederaToken {
+        // The publicly visible name of the token. The token name is specified as a Unicode string. 
+        // Its UTF-8 encoding cannot exceed 100 bytes, and cannot contain the 0 byte (NUL).
+        string name;
+
+        // The publicly visible token symbol. The token symbol is specified as a Unicode string. 
+        // Its UTF-8 encoding cannot exceed 100 bytes, and cannot contain the 0 byte (NUL).
+        string symbol;
+
+        // The memo associated with the token (UTF-8 encoding max 100 bytes)
+        string memo;
+
+        // IWA compatibility. Specified the token supply type. Defaults to INFINITE
+        bool tokenSupplyType;
+
+        // IWA Compatibility. Depends on TokenSupplyType. For tokens of type FUNGIBLE_COMMON - the
+        // maximum number of tokens that can be in circulation. For tokens of type NON_FUNGIBLE_UNIQUE -
+        // the maximum number of NFTs (serial numbers) that can be minted. This field can never be changed!
+        uint32 maxSupply;
+
+        // The ID of the account which will act as a treasury for the token as a solidity address.
+        // This account will receive the specified initial supply or the newly minted NFTs in
+        // the case for NON_FUNGIBLE_UNIQUE Type
+        address treasury;
+
+        // The default Freeze status (frozen or unfrozen) of Hedera accounts relative to this token. If
+        // true, an account must be unfrozen before it can receive the token
+        bool freezeDefault;
+    }
+
+    /// Expiry properties of a Hedera token - second, autoRenewAccount, autoRenewPeriod
+    struct Expiry {
+        // The epoch second at which the token should expire; if an auto-renew account and period are
+        // specified, this is coerced to the current epoch second plus the autoRenewPeriod
+        uint32 second;
+
+        // ID of an account which will be automatically charged to renew the token's expiration, at
+        // autoRenewPeriod interval, expressed as a solidity address
+        address autoRenewAccount;
+
+        // The interval at which the auto-renew account will be charged to extend the token's expiry
+        uint32 autoRenewPeriod;
+    }
+
+    /// A fixed number of units (hbar or token) to assess as a fee during a transfer of
+    /// units of the token to which this fixed fee is attached.
+    struct FixedFee {
+        // The number of units to assess as a fee
+        uint32 amount;
+
+        // The denomination of the fee; taken as hbar if left unset and taken as the id
+        // of the newly created token if set to the sentinel value of 0x0
+        address tokenId;
+
+        // The ID of the account to receive the custom fee, expressed as a solidity address
+        address feeCollector;
+    }
+
+    /// A fraction of the transferred units of a token to assess as a fee. The amount assessed will never
+    /// be less than the given minimumAmount, and never greater than the given maximumAmount.  The
+    /// denomination is always units of the token to which this fractional fee is attached.
+    struct FractionalFee {
+        // A rational number's numerator, used to set the amount of a value transfer to collect as a custom fee
+        uint32 numerator;
+
+        // A rational number's denominator, used to set the amount of a value transfer to collect as a custom fee
+        uint32 denominator;
+
+        // The minimum amount to assess
+        uint32 minimumAmount;
+        
+        // The maximum amount to assess (zero implies no maximum)
+        uint32 maximumAmount;
+        bool netOfTransfers;
+
+        // The ID of the account to receive the custom fee, expressed as a solidity address
+        address feeCollector;
+    }
+
+    /// A fee to assess during a transfer that changes ownership of an NFT. Defines the fraction of
+    /// the fungible value exchanged for an NFT that the ledger should collect as a royalty. ("Fungible
+    /// value" includes both ℏ and units of fungible HTS tokens.) When the NFT sender does not receive
+    /// any fungible value, the ledger will assess the fallback fee, if present, to the new NFT owner.
+    /// Royalty fees can only be added to tokens of type type NON_FUNGIBLE_UNIQUE.
+    struct RoyaltyFee {
+        // A fraction's numerator of fungible value exchanged for an NFT to collect as royalty
+        uint32 numerator;
+
+        // A fraction's denominator of fungible value exchanged for an NFT to collect as royalty
+        uint32 denominator;
+
+        // If present, the fixed fee to assess to the NFT receiver when no fungible value is exchanged
+        // with the sender
+        FixedFee fixedFee;
+
+        // The ID of the account to receive the custom fee, expressed as a solidity address
+        address feeCollector;
+    }
+
+    /// A Key can be a public key from either the Ed25519 or ECDSA(secp256k1) signature schemes, where
+    /// in the ECDSA(secp256k1) case we require the 33-byte compressed form of the public key. We call
+    /// these public keys <b>primitive keys</b>.
+    /// A Key can also be the ID of a smart contract instance, which is then authorized to perform any 
+    /// precompiled contract action that requires this key to sign. 
+    /// Note that when a Key is a smart contract ID, it <i>doesn't</i> mean the contract with that ID 
+    /// will actually create a cryptographic signature. It only means that when the contract calls a 
+    /// precompiled contract, the resulting "child transaction" will be authorized to perform any action
+    /// controlled by the Key.
+    /// Exactly one of the possible values should be populated in order for the Key to be valid.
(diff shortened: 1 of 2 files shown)


assets/hip-206/solidity/IHederaTokenService.sol
```diff
@@ -49,6 +49,168 @@ interface IHederaTokenService {
...
+        // will belong to the given contract, since it could be running another contract's code via 
+        // <tt>delegatecall</tt>. So setting this key is a more permissive version of setting the
+        // contractID key, which also requires the code in the active message frame belong to the
+        // the contract with the given id.)
+        address delegatableContractId;
+    }
+
+    /// A convenience struct used to group different types of Keys we want to set to a Token during
+    /// token creation.
+    struct TokenKeys{
```
shemnon, line 184: Rather than a struct with all of the possible keys what about a struct with a key type (one of admin, kyc, ...) and then the Key.  The API would accept an array of this key struct.

And then, what if we made the type actually a bit field, so you could set multiple keys with one entry.  Odds are most contracts will set all to one, or one to each.  For the all to one scenario we save a lot of gas.

shemnon (commented):
Before we "lock in" this API I think we should see what the solidity code that exercises these APIs would look like for common scenarios.  Some constructs may be very unweildy to build.

Two simple scenarios:
* Token with no custom fees, with a user account as admin, contract as supply and pause key.
* NFT token with a royalty fee, contract has the mint and admin key.

Also, do we need APIs to update fee schedule? And Pause a token, as well as freeze/kyc/wipe accounts?  Maybe APIs to do admin operations (like update keys I guess?)

assets/hip-206/solidity/IHederaTokenService.sol
```diff
@@ -49,6 +49,148 @@ interface IHederaTokenService {
...
+        // will be created for the token.
+        // 0th bit: adminKey
+        // 1st bit: kycKey
+        // 2nd bit: freezeKey
+        // 3rd bit: wipeKey
+        // 4th bit: supplyKey
+        // 5th bit: feeScheduleKey
+        // 6th bit: pauseKey
+        // 7th bit: ignored
+        uint8 keyType;
```
shemnon, line 174: Unless this is a "packed" struct we should kale it at least uint16.  since it's not intended to "go over the wire" there is no need to pack it. perhaps just `uint`?

assets/hip-206/solidity/IHederaTokenService.sol
```diff
@@ -133,6 +275,44 @@ interface IHederaTokenService {
...
+    /// @return bool whether the create was successful
+    /// @return address the address of the created token
+    function createFungibleToken(
+        HederaToken memory token, 
+        uint initialTotalSupply, 
+        uint decimals, 
+        Expiry memory expiry, 
+        FixedFee[] memory fixedFees, 
+        FractionalFee[] memory fractionalFees, 
+        Key[] memory keys) external returns (bool success, bytes tokenAddress);
```
shemnon, line 296: obsessive nit: I would put `keys` earlier in the list because they are more important.  Before the custom fees.  Possibly just before expiry.  Custom fees may be empty and I would want mandatory fields before optional.

assets/hip-206/solidity/TokenCreateContract.sol
```diff
@@ -0,0 +1,159 @@
...
+        IHederaTokenService.FractionalFee[] memory fractionalFees;
+
+        IHederaTokenService.HederaToken memory myToken;
+        myToken.name = "MyToken";
+        myToken.symbol = "MTK";
+        myToken.treasury = address(this);
+        myToken.tokenKeys = keys;
+
+        (bool success, address token, bytes memory errorMessage) = 
+            HederaTokenService.createFungibleToken(myToken, 200, 8, fixedFees, fractionalFees);
```
shemnon, line 49: I think we should have a varient without custom fees, or fold it into the token description.

assets/hip-206/solidity/TokenCreateContract.sol
```diff
@@ -0,0 +1,159 @@
...
+        // turn on bit corresponding to admin key type
+        adminKeyType = adminKeyType.setBit(0);
+        // set the value of the key to be inherited from the calling account
+        adminKeyValue.inheritAccountKey = true;
+        
+        // create TokenKey of types supplyKey and pauseKey with value a contract address passed as function arg
+        uint supplyPauseKeyType;
+        IHederaTokenService.KeyValue memory supplyPauseKeyValue;
+        // turn on bits corresponding to supply and pause key types
+        supplyPauseKeyType = supplyPauseKeyType.setBit(4);
```
shemnon, line 28: We should have constants for these in IHederaTokenService instead of relying on magic numbers.

assets/hip-206/solidity/TokenCreateContract.sol
```diff
@@ -0,0 +1,159 @@
...
+
+        IHederaTokenService.HederaToken memory myToken;
+        myToken.name = "MyNFT";
+        myToken.symbol = "MNFT";
+        myToken.treasury = address(this);
+        myToken.tokenKeys = keys;
+
+        // create the token through HTS with default expiry and royalty fees;
+        (bool success, address token, bytes memory errorMessage) = 
+            HederaTokenService.createNonFungibleToken(myToken, fixedFees, fractionalFees, royaltyFees);
```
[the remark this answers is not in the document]
stoqnkpL: I thought about folding the custom fees into the `HederaToken` object, but I see two problems with that:

- `RoyaltyFee` contains a `FixedFee` within its structure and I'm not sure Solidity will handle having a struct containing both FixedFees and RoyaltyFees with FixedFees within them well
- As the `HederaToken` struct gets more and more complicated it becomes harder for decoding and testing all edge-cases

But I do agree that we need to simplify the interface and adding two more create functions with no custom fee arguments is necessary.

stoqnkpL:
One question that came up internally today is whether we really want to allow updates of tokens for the threshold keys with this HIP or just limit the Token create HIP to the create functions with 1 of 1 keys and introduce the Update functionality as part of a separate HIP

SimiHunjan:
Limit the Token create HIP to the create functions with 1 of 1 keys for this release

HIP/hip-358.md
```diff
@@ -0,0 +1,119 @@
...
+| ---------- | -------------------------------------------------------------------------------------------------|--------------------------|
+|  | `createFungibleToken(HederaToken, uint, uint)`                                                             | `(bool, addess, bytes)`  |
+|  | `createFungibleTokenWithCustomFees(HederaToken, uint, uint, FixedFee[], FractionalFee[])`                  | `(bool, addess, bytes)`  |
+|  | `createNonFungibleToken(HederaToken, uint, uint)`                                                          | `(bool, addess, bytes)`  |
+|  | `createNonFungibleTokenWithCustomFees(HederaToken, uint, uint, FixedFee[], FractionalFee[], RoyaltyFee[])` | `(bool, addess, bytes)`  |
+
+### Precompile Gas Costs
+
+Gas will be priced so that it will be cheaper to do calls directly to HTS rather
+than through HSCS where such calls are possible.
```
shemnon, line 75: We will need actual numbers before this can go final, but it won't block initial submission.

Can you add a table with the functiongroups and gas cost column with `TBD` as the price?

shemnon (dismissed):

sergmetelin (changes requested):

shemnon (approved):

assets/hip-206/solidity/IHederaTokenService.sol
```diff
@@ -160,8 +160,12 @@ interface IHederaTokenService {
         uint32 amount;
 
         // The denomination of the fee; taken as hbar if left unset and taken as the id
-        // of the newly created token if set to the sentinel value of 0x0
+        // of the newly created token if set to the sentinel value of 0x0 (since there
+        // is no notion of unset or null value in Solidity and the default value of an
+        // address variable is 0x0, the `isTokenIdSet` variable is used to distinguish
+        // between the two cases)
         address tokenId;
+        bool isTokenIdSet;
```
shemnon, line 168: Why would tokenId ever be un-set?  If passed in as an argument I think it's reasonable to expect that the required fields are set.  Is there any meaning to a fixed fee with an amount, a fee collector, and an unset tokenI?

stoqnkpL: Yes, there are some special meanings in the protobufs values 0 and unset for this field:

- if tokenId is 0, then use the token that's being created in this tx's id
- if tokenId is unset, then use hbars
- if tokenId is set to non-zero number, then use the id specified

When decoding Solidity input args however it's impossible to distinguish between cases 1 and 2 since missing value and zero value are both decoded to 0x0.

I do think that we can make it easier to read and understand in the Solidity struct though, maybe something like:

address tokenId;
bool useHbarsForPayment;
bool useCurrentTokenForPayment;

and specify that exactly one of the fields must be populated. If more than one field is populated we would fail the decoding, set error message in the message frame and return null from the precompile invocation.

sergmetelin (approved):
LGTM

HIP/hip-358.md
```diff
@@ -0,0 +1,123 @@
...
+| hash       | effective signature                                                                       | return           |
+| ---------- | ------------------------------------------------------------------------------------------|------------------|
+| `7812a04b` | `createFungibleToken(HederaToken, uint, uint)`                                            | `(int, addess)`  |
+| `4c381ae7` | `createFungibleTokenWithCustomFees(HederaToken, uint, uint, FixedFee[], FractionalFee[])` | `(int, addess)`  |
+| `9dc711e0` | `createNonFungibleToken(HederaToken)`                                                     | `(int, addess)`  |
+| `5bc7c0e6` | `createNonFungibleTokenWithCustomFees(HederaToken, FixedFee[], RoyaltyFee[])`             | `(int, addess)`  |
+
+### Precompile Gas Costs
+
+In order to allow creation of token through the precompiled contract, a new mechanism for payment is introduced - precompile calls are partially payed by Hbar value transfers.
```
shemnon, line 67: We will have to update `com.hedera.services.store.contracts.HederaWorldState#TOKEN_CALL_REDIRECT_CONTRACT_BINARY` - make it payable.  The current evm code will fail when any value is passed in.  We will need to move the value check to another part of the HTS precompile.

HIP/hip-358.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Specification
+
+### EVM Precompile extension
+
+The [Solidity file for development](../assets/hip-206/solidity/IHederaTokenService.sol) is updated with new types and function signatures that the precompile will respond to. It is included in this HIP by reference.
+
+### Solidity Function Signatures 
+
+The following structs are defined to simplify the interface between solidity and the HTS precompile:
```
shemnon, line 44: For the data structures we should describe what each parameter does, not just the type.  Probably in a markdown prose section not in the table.

HIP/hip-358.md
```diff
@@ -0,0 +1,123 @@
...
+| `9dc711e0` | `createNonFungibleToken(HederaToken)`                                                     | `(int, addess)`  |
+| `5bc7c0e6` | `createNonFungibleTokenWithCustomFees(HederaToken, FixedFee[], RoyaltyFee[])`             | `(int, addess)`  |
+
+### Precompile Gas Costs
+
+In order to allow creation of token through the precompiled contract, a new mechanism for payment is introduced - precompile calls are partially payed by Hbar value transfers.
+There is a base cost of 100,000 Gas for each call to the create functions. Additional Hbar cost is required to be transferred as `value` to the precompile call. The Hbar price will be calculated using the following formula:
+
+The total cost of HTS Token create through the precompile is equal to the cost of a HAPI Create Token Transaction with similar arguments at that instance plus a 20% premium. 
+The Hbars required to be transferred as value in the precompile call will be equal to the total cost required minus the Hbar cost for the base 100,000 gas cost at that instance.
```
shemnon, line 71: > 100,000 gas cost

"100,000 gas cost converted into hbar according to the current gas price"

shemnon (approved):
This has the core of the idea, but before it comes out of review we need to firm up the meaning of the field in the HIP and not via the supplementary solidity fiels.

SimiHunjan (approved):

PR #360, opened 2022-02-11
Update HIP-329 with final implementation details

tinker-michaelj:
**Description**:
- Update `CREATE2` spec with several details finalized during the implementation.


sergmetelin (approved):
LGTM

PR #367, opened 2022-02-18
HIP-367 : Remove Token Association Limit

anighanta:
**Description**:

**Related issue(s)**:

HIP for removing the token association limit.

**Notes for reviewer**:

**Checklist**


HIP/hip-1000.md
```diff
@@ -0,0 +1,111 @@
...
+requires: <HIP number(s)>
+replaces: <HIP number(s)>
+superseded-by: <HIP number(s)>
+---
+
+## Abstract
+
+Describes the removal of the limit [1000] on the number of tokens an account can be associated to and the refactor of 
+how we store this **unlimited** list of associated tokens on the Account state. 
+
```
rbair23, line 22: Maybe we can reword this slightly:

> Permits every account to hold an unlimited number of token associations. Retains the current pricing for each token association. Modifies the number of token associations returned in the `getAccountBalance` query.

HIP/hip-1000.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Motivation
+
+The current Services API allows the user to only have upto 1000 tokens associated at a time.
+For users that need more associations :
+- one can create multiple accounts to hold 1000 token associations each.
+- user can create a smartContract and manage the mapping of multiple accounts with 1000 token associations each.
```
rbair23, line 29: Nit: typo "upto". Perhaps reword:
> The current Services API permits the user to have at most 1,000 token associations per account.
> A user can work around this limitation by creating multiple accounts, each of which is limited to
> 1,000 token associations. For some use cases, this 1,000 token association limit is frustrating.
> Due to the current implementation, it makes state proofs large and expensive. In addition, the
> current implementation imposes excessive costs as the number of token associations per account
> grows.

HIP/hip-1000.md
```diff
@@ -0,0 +1,111 @@
...
+
+and a few others which are mentioned [here](https://www.notion.so/hedera/Increase-1000-associations-account-limit-95e882c6954443c6bc6502ad2818d15b#e8d1f192f620414c88a5ae18dfcca69b)
```
rbair23, line 31: We must not reference our notion pages in a HIP. The world does not have access to our Notion pages so including a link doesn't make any sense. If the other workarounds are worth listing, then we should restate them in the HIP.

HIP/hip-1000.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Rationale
+
+Rather, we should remove the limit we put in place for the token associations per account.
+
+And as the array of tokenIds that are associated to an account on the Account state grow, hashing/re-hashing every time
+we modify this array becomes expensive. A refactor of how we handle these associated tokenIds is needed.
```
rbair23, line 38: I think each section should stand alone. By starting the sentence with "Rather", we are continuing from the above section. Lets rework the wording. I'm also not sure how Rationale differs from Motivation?

HIP/hip-1000.md
```diff
@@ -0,0 +1,111 @@
...
+
+## User stories
+- As a primary creator of an NFT collection, I should be able to mint a large number of NFT serials in a single collection
+- As a secondary NFT marketplace dApp, I should be able to create many thousands of collections (created by multiple primary creators of the collections), such that each collection can hold a large number of NFTs
+- As a NFT creator, I should be able to set custom fees for my NFTs
+    - Is it a must-have requirement to be able to do this on per-serial number? or is it sufficient to specify this fee at the token-type level?
```
rbair23, line 44: What does this have to do with the association limit?

HIP/hip-1000.md
```diff
@@ -0,0 +1,111 @@
...
+and a few others which are mentioned [here](https://www.notion.so/hedera/Increase-1000-associations-account-limit-95e882c6954443c6bc6502ad2818d15b#e8d1f192f620414c88a5ae18dfcca69b)
+
+## Rationale
+
+Rather, we should remove the limit we put in place for the token associations per account.
+
+And as the array of tokenIds that are associated to an account on the Account state grow, hashing/re-hashing every time
+we modify this array becomes expensive. A refactor of how we handle these associated tokenIds is needed.
+
+## User stories
```
rbair23, line 40: We need to flesh this section out. The User Stories should directly go into test scenarios. So for example, we should have user stories like:

> As a user, I can associate my account with more than 1,000 different fungible tokens
> As a user, I can associate my account with a combination of more than 1,000 different fungible and non-fungible tokens

And so forth. I like how you have a stronger sense of user than just "a user", such as "a primary creator", "a secondary NFT marketplace dApp", etc. Those are nice. Lets really drill into the use cases. This is the most important part of the HIP in that it is where we set expectations for what we're trying to do.

HIP/hip-1000.md
```diff
@@ -0,0 +1,111 @@
...
+![proposed.png](../assets/hip-1000/proposed.png)
+
+## Backwards Compatibility
+
+There are no known backwards compatibility issues.
+
+## Security Implications
+
+This shouldn’t necessarily change security implications not already known by Hedera,
+but could introduce denial of service ? Which can be controlled by throttling and increasing fee as the associations grow.
```
rbair23, line 92: This needs to be clarified. There is no new denial of service because an attacker today can still flood the system with accounts and associations. Creating a new account every 1,000 associations is not hard. So there is no new security implications here.

HIP/hip-1000.md
```diff
@@ -0,0 +1,111 @@
...
+![current.png](../assets/hip-1000/current.png)
+
+We propose a new way of efficiently doing this using a `MapValueLinkedList` which will store the Account-Token relationship as a Map of <EntityNum pair of accountID and tokenID> to <MerkleTokenRelStatus>
+where the keys will be linked by a linkedList. And only the latest account-token association key is saved on the MerkleAccount state.
+
+![proposed.png](../assets/hip-1000/proposed.png)
+
+## Backwards Compatibility
+
+There are no known backwards compatibility issues.
```
rbair23, line 87: When the account has more than 1,000 associations, a `getAccountBalance` call will not return more than those. We also want to deprecate this setting so we either return no associations in that query, or we limit it to a small number (like 5) in the future. While we would not change the number returned in this HIP, we should document the eventual goal to limit this number significantly in the future and encourage callers to go to the mirror nodes for this state.

HIP/hip-1000.md
```diff
@@ -0,0 +1,111 @@
...
+
+We currently store the list of tokens associated to an account on the state using a MerkleLeaf node which hold the array of Ids of the token and each MerkleAccount will have this as a child.
+So as the array grows it will be very costly to operate on this array. 
+
+![current.png](../assets/hip-1000/current.png)
+
+We propose a new way of efficiently doing this using a `MapValueLinkedList` which will store the Account-Token relationship as a Map of <EntityNum pair of accountID and tokenID> to <MerkleTokenRelStatus>
+where the keys will be linked by a linkedList. And only the latest account-token association key is saved on the MerkleAccount state.
+
+![proposed.png](../assets/hip-1000/proposed.png)
```
rbair23, line 83: We need a much clearer description of the change.

rbair23 (changes requested):

HIP/hip-1000.md
```diff
@@ -0,0 +1,161 @@
...
+represented by Account-Token relationship 
+Map of <EntityNumPair, MerkleTokenRelStatus> in which `MerkleTokenRelStatus` serves as 
+a node of a doubly linkedList. The nextKey and prevkey in this `MerkleTokenRelStatus` 
+will point to the next and previous token relationships of the account. 
+And only the latest account-token association key is saved in MerkleAccountState.
+
+![img.png](../assets/hip-1000/updates.png)
+
+#### Migration
+When loading older states, use the deprecated `MerkleAccountTokens` to deseiralize the associated tokens of an account
```
Neeharika-Sompalli, line 75: typo `deserialize`

HIP/hip-1000.md
```diff
@@ -0,0 +1,161 @@
...
+### Token Dissociation
+When dissociating a token, get the next and prev relationship and udpate the keys such that they point to each other rather than the dissociating token relationship.
+
+```
+Algorithm:
+1. if the dissociating token relationship is the  account's `lastAssociatedToken`, check if there are any other associations
+    a. if not then update the `lastAssociatedToken` field to the default [0]
+    b. else get the next `MerkleTokenRelStatus` and update the prevKey to the default [0] and set the `lastAssociatedToken` to the next realtion's key 
+2. else get the next and prev `MerkleTokenRelStatus` association objects and update the next and prevKeys to point to each other respectively.
+3. if the nextKey in step 2 is deafult [0] then we are dissocaiting the last tokenAssocation in the list and we dont have to process the next realtion and just set the prev relation's next key as default [0]
```
Neeharika-Sompalli, line 109: typo `default`

HIP/hip-1000.md
```diff
@@ -0,0 +1,161 @@
...
+
+#### Migration
+When loading older states, use the deprecated `MerkleAccountTokens` to deseiralize the associated tokens of an account 
+form the state during migration. Then use the Map<EntityNum, MerkleAccount> accounts 
+and Map<EntityNumPair, MerkleTokenRelStatus> tokenAssociations maps which will not have any of the new fields populated to fill out the newly added fields.
+All of this has to be done during the `migrate()` call in `ServicesState`
+
+```
+Algorithm:
+1. for each account in the accounts Map fetch the list of tokenIds from `MerkleAccountTokens`.
```
Neeharika-Sompalli, line 82: `map`

HIP/hip-1000.md
```diff
@@ -0,0 +1,161 @@
...
+2. for each token on this list, use the account id and token id to build `EntityNumPair` and fetch the MerkleTokenRelStatus from the tokenAssociations map.
+3. update the nextKey and prevKey for each of these associations and persist the changes.
+4. set the account's `lastAssociatedToken` to the last key that is built in the above loop.
+```
+
+### Token Association
+As the account's `lastAssociatedToken` will always has the latest association key which suggest we always insert at the head of the linkedList.
```
Neeharika-Sompalli, line 89: `will always have`

HIP/hip-1000.md
```diff
@@ -0,0 +1,161 @@
+---
+hip:1000
```
rbair23, line 2: We can update this to 367 now.

HIP/hip-1000.md
```diff
@@ -0,0 +1,161 @@
...
+title: HIP Unlimited token associations per Account
+author: Anirudh Ghanta <anirudh.ghanta@hedera.com>
+working-group: a list of the technical and business stakeholders' name(s) and/or username(s), or name(s) and email(s).
```
rbair23, line 5: We should probably populate this too.

HIP/hip-1000.md
```diff
@@ -0,0 +1,161 @@
...
+type: Standards Track
+category: Service
+needs-council-approval: Yes
+status: Draft <Draft | Review | Last Call | Active | Inactive | Deferred | Rejected | Withdrawn | Accepted | Final | Replaced>
```
rbair23, line 9: I think we remove everything inside the diamonds

HIP/hip-1000.md
```diff
@@ -0,0 +1,161 @@
...
+created: 2022-02-17
+discussions-to: <a URL pointing to the official discussion thread>
```
rbair23, line 11: We need to create a discussion thread and link it here.

HIP/hip-1000.md
```diff
@@ -0,0 +1,161 @@
...
+updated: <comma separated list of dates>
+requires: <HIP number(s)>
+replaces: <HIP number(s)>
+superseded-by: <HIP number(s)>
```
rbair23, line 15: I think all 4 of these are removed? Unless there was a previous HIP that discussed token associations, in which case we should add a link here to say this one replaces it.

HIP/hip-1000.md
```diff
@@ -0,0 +1,161 @@
...
+## Abstract
+Permits every account to hold an unlimited number of token associations. Retains the current pricing for each token association. Modifies the number of token associations returned in the getAccountInfo and getAccountBalance query.
+
+## Motivation
+The current Services API permits the user to have at most 1,000 token associations per account.
+A user can work around this limitation by creating multiple accounts, each of which is limited to
+1,000 token associations. For some use cases, this 1,000 token association limit is frustrating.
+Due to the current implementation, it makes state proofs large and expensive. In addition, the
+current implementation imposes excessive costs as the number of token associations per account
+grows.
```
rbair23, line 27: > Finally, returning large numbers of token associations in the free queries can lead to excessively long query response times.

rbair23 (changes requested):

HIP/hip-367.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+### Token Dissociation
+When dissociating a token, get the next and prev relationship and update the keys such that they point to each other rather than the dissociating token relationship.
+
+```
+Algorithm:
+1. if the dissociating token relationship is the  account's `lastAssociatedToken`, check if there are any other associations
+    a. if not then update the `lastAssociatedToken` field to the default [0]
+    b. else get the next `MerkleTokenRelStatus` and update the prevKey to the default [0] and set the `lastAssociatedToken` to the next realtion's key
```
Neeharika-Sompalli, line 113: typo `relation`

HIP/hip-367.md
```diff
@@ -0,0 +1,238 @@
...
+## User stories
+
+- As a user I should be able to associate to any number of tokens without any limitations under one account instead of dispersing them across multiple accounts.
+- As a user I should be able to pay for as many token associations upfront as I want to, using the automatic association slots 
+- As a primary creator of an NFT collection, I should be able to mint a large number of NFT serials in a single collection
+- As a secondary NFT marketplace dApp, I should be able to create many thousands of collections (created by multiple primary creators of the collections), such that each collection can hold a large number of NFTs
+- As an avid collector of NFT, I want to manage my entire collection within the same account.
+
+## Specification
+Currently, we have a limit of 1000 token associations per Account put in place which is dictated by the bootstrap property `tokens.maxPerAccount`.
```
Neeharika-Sompalli, line 37:
```suggestion
Currently, we have a limit of 1000 token associations per account put in place which is dictated by the bootstrap property `tokens.maxPerAccount`.
```

HIP/hip-367.md
```diff
@@ -0,0 +1,238 @@
...
+    }
+```
+
+When using the Account model:
+```java
+    public void associateWith(List<Token> tokens, int maxAllowed, boolean automaticAssociation){
+        final var alreadyAssociated=associatedTokens.size();
+        final var proposedNewAssociations=tokens.size()+alreadyAssociated;
+        validateTrue(proposedNewAssociations<=maxAllowed,TOKENS_PER_ACCOUNT_LIMIT_EXCEEDED);
+        ...
```
Neeharika-Sompalli, line 57: nit : spaces before and after `=`
```suggestion
    public void associateWith(List<Token> tokens, int maxAllowed, boolean automaticAssociation){
        final var alreadyAssociated = associatedTokens.size();
        final var proposedNewAssociations = tokens.size()+alreadyAssociated;
        validateTrue(proposedNewAssociations <= maxAllowed,TOKENS_PER_ACCOUNT_LIMIT_EXCEEDED);
        ...
```

HIP/hip-367.md
```diff
@@ -0,0 +1,238 @@
...
+That enforcement also has to be removed.
+
+The list of tokens associated to an account are saved on the state using a MerkleLeaf node `MerkleAccountTokens` which holds the array of Ids of the token as `CopyOnWriteIds` and each MerkleAccount will have this as a child. 
+The new way of efficiently doing this is by using a `MapValueLinkedList` data structure which will be represented by Account-Token relationship Map of <EntityNumPair, MerkleTokenRelStatus> in which `MerkleTokenRelStatus` serves as a node of a doubly linkedList. 
+The `nextKey` and `prevkey` in this `MerkleTokenRelStatus`will point to the next and previous token relationships of the account. And `tokenAssociaitonMetadata` is saved on MerkleAccountState.
+The `TokenAssociationMetadata` encapsulates three properties of the MerkleAccount
+1. The total number of tokens that this account is associated with `numAssociations`
+2. The total number among these associations which have zero balance `numZeroBalances`
+3. And finally the latest association represented as an `EntityNumPair` of MerkleAccount's id and token's id `lastAssociation`
```
Neeharika-Sompalli, line 70: 3. And finally the latest association represented as an `EntityNumPair` of account's id and token's id as `lastAssociation` 
```

**Review by Neeharika-Sompalli — DISMISSED** [review]
Other than minor comments. LGTM

File: HIP/hip-367.md [hunk 24]
```diff
@@ -0,0 +1,238 @@
...
+---
+
+## Abstract
+Permits every account to hold an unlimited number of token associations. Retains the current pricing for each token association. Modifies the number of token associations returned in the `getAccountInfo` and `getAccountBalance` query. Changes the criteria for `cryptoDelete` transaction to successfully mark the account as `deleted`
+
+## Motivation
+The current Services API permits the user to have at most 1,000 token associations per account.
+A user can work around this limitation by creating multiple accounts, each of which is limited to
+1,000 token associations. For some use cases, this 1,000 token association limit is frustrating.
+Due to the current implementation, it makes state proofs large and expensive. In addition, the
```
tinker-michaelj, line 22: I'm not sure any state proofs will get smaller after this HIP; am I missing something? 🤔

HIP/hip-367.md
```diff
@@ -0,0 +1,238 @@
...
+- As a primary creator of an NFT collection, I should be able to mint a large number of NFT serials in a single collection
+- As a secondary NFT marketplace dApp, I should be able to create many thousands of collections (created by multiple primary creators of the collections), such that each collection can hold a large number of NFTs
+- As an avid collector of NFT, I want to manage my entire collection within the same account.
+
+## Specification
+Currently, we have a limit of 1000 token associations per account put in place which is dictated by the bootstrap property `tokens.maxPerAccount`.
+This is enforced everytime we associate token/tokens to an account. 
+
+When using the HederaTokenStore :
+```java
```
tinker-michaelj, line 41: Imo HIP's don't require code snippets; in part because the codebase is changing so often---for example, this excerpt no longer exists (as you know 😁).

HIP/hip-367.md
```diff
@@ -0,0 +1,238 @@
...
+The new way of efficiently doing this is by using a `MapValueLinkedList` data structure which will be represented by Account-Token relationship Map of <EntityNumPair, MerkleTokenRelStatus> in which `MerkleTokenRelStatus` serves as a node of a doubly linkedList. 
+The `nextKey` and `prevkey` in this `MerkleTokenRelStatus`will point to the next and previous token relationships of the account. And `tokenAssociaitonMetadata` is saved on MerkleAccountState.
+The `TokenAssociationMetadata` encapsulates three properties of the MerkleAccount
+1. `numAssociations` : The total number of tokens that this account is associated
+2. `numZeroBalances` : The total number among these associations which have zero balance
+3. `lastAssociation`: The latest association represented as an `EntityNumPair` of account's id and token's id 
+
+![img.png](../assets/hip-1000/updates.png)
+
+These fields are tracked / updated at :
```
tinker-michaelj, line 74: This kind of algorithmic summary is great imo. 👍

tinker-michaelj (commented):
This is a very readable HIP even with the fine-grained detail, which is hard to accomplish! 🥂 

To preserve its readability for future audiences, I would suggest removing detail that is specific to the current codebase (e.g. code snippets).

Non-developers do not benefit from this level of detail; and anyone who did need implementation details for some reason would be already be doing extensive archeology in the GitHub repo. 😁

tinker-michaelj (dismissed):
LGTM

SimiHunjan (dismissed):

SimiHunjan (approved):

rbair23 (approved):

PR #376, opened 2022-03-02
HIP-376 Support Approve/Allowance/transferFrom standard calls from ERC20 and ERC721

stoqnkpL:
**Description**:
Extend the support for ERC20 and ERC721 functions to be called on HTS tokens with support for Approve/allowance and transferFrom

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-376.md
+++ b/HIP/hip-376.md
@@ -1,5 +1,5 @@
 ---
-hip: 
+hip: 376
 title: Support for Approve/Allowance and ERC20/ERC721 transferFrom in HTS precompile
 author: Stoyan Panayotov <stoyan.panayotov@limechain.tech>
 type: Standards Track
--- a/HIP/hip-376.md
+++ b/HIP/hip-376.md
@@ -24,11 +24,11 @@ The goal is to expose [HIP-336](https://hips.hedera.com/hip/hip-336): Approval a
 
 ### Fungible token related stories
 
-As a hedera smart contract services user, I want to enable another user the ability to spend `x` fungible tokens from my account so that they can trade on my behalf without requiring that I expose my private key to them.
+As a hedera smart contract services user, I want to enable another user or contract the ability to spend `x` fungible tokens from my account so that they can trade on my behalf without requiring that I expose my private key to them.
 
 As a hedera smart contract services user, I want to query tokens for the current token allowances that have been granted to a user (spender) for a specific owner account.
 
-As a hedera smart contract services user, I want to modify or remove an allowance I have granted to another user so that I can manage that user's ability to spend my tokens over time.
+As a hedera smart contract services user, I want to modify or remove an allowance I have granted to another user or contract so that I can manage that user/contract's ability to spend my tokens over time.
 
 As a hedera smart contract services user that has been granted an allowance, I want to be able to  transfer tokens owned by the owner account as if they were my tokens so that I can issue transfers on their behalf without them having to expose their private key.
 
@@ -40,8 +40,7 @@ As a hedera smart contract services user, I want to enable or disable another us
 
 As a hedera smart contract services user, I want to modify or remove an allowance I have granted to another user so that I can manage that user's ability to spend my tokens over time.
 
-// TODO: can we allow multiple accounts for the same NFT in Hedera? What should happen then when we call getApproved?
-As a hedera smart contract services user, I want t
+As a hedera smart contract services user, I want to query a token for the information whether an account has approve for all nfts permission.
 
 As a hedera smart contract services user that has been granted an allowance, I want to be able to  transfer tokens owned by the owner account as if they were my tokens so that I can issue transfers on their behalf without them having to expose their private key.
 
@@ -53,7 +52,7 @@ Extend the functionality from [HIP-218](https://hips.hedera.com/hip/hip-218): Sm
 
 #### Supported ERC-20 operations
 
-Tokens of type `FUNGIBLE_COMMON` will support standard `ERC-20` calls.
+Tokens of type `FUNGIBLE_COMMON` support standard `ERC-20` calls.
 
 The following `ERC-20` operations will be supported. Standard `ERC-20 Events` will be emitted as appropriate.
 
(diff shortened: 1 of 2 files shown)


HIP/hip-376.md
```diff
@@ -0,0 +1,155 @@
...
+
+As a hedera smart contract services user, I want to be able to access my HTS
+non-fungible tokens like an Ethereum ERC-721 token.
+
+As a hedera smart contract services user, I want to be able to access metadata
+provided by the getTokenInfo in a smart contract.
+
+## Specification
+
+### EVM Precompile
```
shemnon, line 49: I would collapse this and the next two section into a reference to HIP-218. We are not creating new redirect contracts but are adding functions to the existing system contract.

HIP/hip-376.md
```diff
@@ -0,0 +1,155 @@
...
+
+For function calls that cause HTS token transfers the appropriate gas charges
+will be applied. Calls that emit events will be charged the event fees too.
+
+For other functions that return data the cost will be 10 gas per 32 bytes
+returned or fraction fo 32 bytes returned. For functions that return integers or
+an address this will be a flat 10 gas. For string values (name, symbol, etc.)
+this will be a variable fee.
+
+#### Supported ERC-20 operations
```
shemnon, line 93: Is the mapping from the solidity functions to the hedera functions simple and obvious?  If there is a difference between the APIs we should describe what the "thunking" between the calls looks like.

HIP/hip-376.md
```diff
@@ -0,0 +1,155 @@
...
+>
+>// Put CALLER in memory
+>// Put the remaining CALLDATA into memory
+>// DELEGATECALL 0x127 with all remaining gas
+>// if the DELEGATECALL failed REVERT()
+>// RETURN the exact same results as the DELEGATECALL
+>
+>```
+
+#### Gas Schedule
```
shemnon, line 83: Now that we know what we did for HIP-206 and HIP-218 gas-wise I think we should add a table for each "function group" and the gas price we expect, including the assumptions used to calculate them.

We take the canonical max calls per second, and divide that into the 15MM gas per second, and add a 20% premium.  The trick is figuring out which max TPX measure to use.

We also want to think of the event emissions cost. My thought is we include those in the precompile gas cost.

HIP/hip-376.md
```diff
@@ -0,0 +1,155 @@
...
+
+
+## Open Issues
+
+None at this time.
+
+## References
+
+- [ERC-20: Token Standard](https://eips.ethereum.org/EIPS/eip-20)
+- [ERC-721: Non-Fungible Token Standard](https://eips.ethereum.org/EIPS/eip-721)
```
shemnon, line 150: Add HIP-218

HIP/hip-376.md
```diff
@@ -0,0 +1,155 @@
...
+---
+
+## Abstract
+
+Extend the functionality from HIP-218 for ERC20 and ERC721 funtions working on
+HTS tokens with support for approve/allowance and transferFrom functions.
+
+## Motivation
+
+The goal of this HIP is to enable users of the HSCS to interact with Hedera
```
shemnon, line 22: re-do motivation.  The goal is to expose HIP-??? functionality in industry standard ERC-20 and ERC-721 calls.

HIP/hip-376.md
```diff
@@ -0,0 +1,156 @@
...
+Calling ERC20 `approve` with amount `200` will be implemented in the following way in the precompile:
+ - query the current configured allowance for the `owner`, `spender` and `tokenId`
+ - calculate the adjustment that needs to be made to have the final allowance equal to the amount received as argument to the ERC20 `approve`
+ - create a `CryptoAdjustAllowance` tx with the calculated adjustment
+
+##### ERC721 approve and multiple approved accounts for NFT in Hedera
+
+ERC721 is designed in such a way that an NFT can have either 0 or 1 approvals e.g. if Alice owns NFT 'X' and Alice calls the approve function two times, once with Bob's account and once with Carol's account, the second call to approve will override Bob's allowance and only Carol will have approved allowance for that NFT. In Hedera the behaviour is different and both Bob and Carol would have approved allowances after the second call to approve.
+To make sure there is no unexpected result for smart contract developers working with HTS tokens, ERC721 `approve` and `getApproved` will behave in the following way with HTS tokens:
+When `approve` is called:
```
Daniel-K-Ivanov, line 92: I have 3 concerns related to the proposed API behaviour.
1. We have potentially unbounded computation or most probably limited to the number of maximum `granted nft allowances`. This should be accounted in the `gas` costs of the precompile
2. It is common for users to change the approvals for specific tokens sequentially, meaning:
```
0. User A is the owner of tokenId=x
1. User A sets approval to User B
2. User A sets approval to User C
```

as per the specification, we will revert on step 2.

3. In the ERC721 standard, allowances are cleared on every transfer. It might be a surprise for developers to see that approvals are not cleared on transfers.

HIP/hip-376.md
```diff
@@ -0,0 +1,156 @@
...
+ - first we'll check if the address from the function arguments is the zero address
+ - if yes, then we'll iterate through the account's `GrantedNFTAllowances` and revoke (adjust with -) any existing approvals for the NFT serial number
```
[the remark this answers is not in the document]
stoqnkpL: From [HIP-336](https://hips.hedera.com/hip/hip-336): Each account is limited to 100 allowances. This limit spans hbar, fungible and non-fungible token allowances. Each NFT serial number will count as a single allowance when computing the total allowances an account currently has.

I agree with the concern you expressed. As I see it we have two options:

1.  Calculate gas price based on number of `GrantedNFTAllowances` we need to revoke
2. Only revoke allowance if there's exactly 1 granted, otherwise revert and require HAPI manipulation of approvals

Personally I like option 1 better since option 2 seems to be a bit limiting, if the user **knows** he'd like to revoke all grants.

HIP/hip-376.md
stoqnkpL: 1. Answered in the other comment reply
2. I think we can update this step from the specification:
`if no, then we'll check if the owner account has exactly 0 existing approvals for that NFT serial number`
to:
`if no, then we'll check if the owner account has exactly 0 or 1 existing approvals for that NFT serial number`
and for the case there is 1 existing approval granted, to replace it (revoke old + grant new)
3. Just checked behaviour in Hedera and allowances are not cleared on transfer, even though the account with the granted allowance is no longer the owner of the token. We could add logic for revoking grants on `transferFrom`. Main concerns are this would result in higher gas price and different behaviour than HAPI `cryptoTransfer` flow.

sergmetelin (dismissed):
LGTM

sergmetelin (approved):
LGTM

PR #400, opened 2022-03-16
Address custom fee has no effect on allowance adjustments

anighanta:
address that the customFee will not be included in the allowance adjustments during allowance transfers


superboo (dismissed):
LGTM

tinker-michaelj (dismissed):
Makes sense to me!

SimiHunjan (dismissed):

PR #415, opened 2022-04-03
HIP: Introduction of Blocks

Daniel-K-Ivanov:
Signed-off-by: Daniel Ivanov <daniel.k.ivanov95@gmail.com>

**Description**:
Adds a HIP for the introduction of Blocks into Hashgraph on the protocol level

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

- [X] Documented (Code comments, README, etc.)


HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+
+Based on the described design goals above, the outlined specification defines that block properties are to be computed and populated at different points in time and by different components of the network, based on their responsibility.
+
+## Specification
+
+**Definitions**
+
+- `block` → `Record file` containing all `Record Stream Objects` grouped in a 2-second time frame.
+- `block number` → consecutive number of the `Record file` that is being incremented by `1` for every new `Record file`. This value can be initially bootstrapped through `Mirror Nodes` and after that maintained by services nodes.
+- `block hash` → The `32 byte keccak256 running hash` of the `Record file`. That is the running hash of the last `Record Stream Object` from the previous `Record file`.
```
tinker-michaelj, line 51: Platform currently supports only a SHA-384 (SHA-2) running hash; switching to keccak256 would have significant ripple effects for mirror nodes I think.

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+| timestamp | Services | Stored in services state (only the last block timestamp) and computed by services. It is the consensusTimestamp of the first transaction in the record file. It is exposed to the 1) EVM during Transaction Execution through the TIMESTAMP opcode; 2) implicitly exported in the record file through the TS of the first transaction in the record file. |
+| hash | Services | Stored in services (last 256 blocks). It is the keccak256(runningHash) of the previous record file. That is the running hash of the last record stream object from the previous record file. It is exposed to the 1) EVM during Transaction Execution through the BLOCKHASH opcode; 2) In the record file as the End Object Running Hash |
+| baseFeePerGas (??) | Mirror Node | Computed by Mirror Node(s). It is a hex encoded value of the gas pricing based in the FeeSchedule file. |
+  | difficulty | Mirror Node | Hardcoded by Mirror Node(s) to hex encoded 0 |
+  | extraData | Mirror Node | Hardcoded by Mirror Nodes(s) to 0x |
+  | gasLimit | Mirror Node |  Computed by Mirror Node(s). The sum of the gasLimit values for all ContractLocalCall, ContractCall and ContractCreate transactions within the block. |
```
steven-sheehy, line 140: `ContractLocalCall` won't appear in the record stream as it does not contribute to consensus. Same comment for other references to it.

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+  | miner | Mirror Node | Hardcoded by the Mirror Nodes to the 0.0.98 address in the long zero prefix format.  |
+  | mixHash | Mirror Node | Hardcoded by Mirror Node(s) to 0x |
+  | nonce | Mirror Node | Hardcoded by Mirror Node(s) to hex-encoded 0 |
+  | parentHash | Mirror Node | Computed by Mirror Node(s). The hash of the parent block. |
+  | receiptsRoot | Mirror Node | (???) |
+  | sha3Uncles | Mirror Node | Hardcoded by Mirror Node(s) to the SHA3 computation of empty array (0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347) |
+  | size | Mirror Node | Computed by Mirror Node(s). The sum of the bytes of all transaction(s) of type ContractLocalCall, ContractCall and ContractCreate within the block. transaction is the transaction specified in the RecordStreamObject and not the Transaction Record |
+  | stateRoot | Mirror Node | (???) |
+  | totalDifficulty | Mirror Node | Hardcoded by Mirror Nodes(s) to hex-encoded 0 |
+  | transactions | Mirror Node | Computed by Mirror Node(s) by ingesting the record stream and aggregating RecordStreamObjects of type ContractLocalCall, ContractCall and ContractCreate within the block. |
```
steven-sheehy, line 152: Believe precompiled contracts would also be included.

steven-sheehy (changes requested):

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+- `block timestamp` → Instant of consensus timestamp of the first `transaction`/`Record Stream Object` in the `Record file`.
+
+### Platform
+
+Adapt `TimestampStreamFileWriter` to include `blockNumber` in the `Record File`. Introduce a new field `firstConsensusTimeInCurrentFile` to be used as a marker when to start a new Record file. Use the field `lastConsensusTimestamp` to keep track of the last-seen consensus timestamp that was processed. In this way, we can ensure that we have at least `1000ns`  difference between the last processed transaction before a new file is created. In this way if we have a `parent transaction` with at least one `child precompile transaction` they all will be included in the same `block`/`Record file`. Otherwise, we might have a corner case where a `parent transaction` is included in one block, and its `child precompile transaction` falls into the next block since it will increase the `consensus timestamp` with `1ns`. Therefore, the algorithm for checking whether to start a new file would be the following:
```
shemnon, line 56: What is the unit of `firstConsensusTime` and `lastConsensusTimestamp`?  seconds, milli?  micro?  nano? The math implies nano but we should be explicit about it.

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+2. `T - firstConsensusTimeInCurrentFile > 2s`
+
+or
+
+if `lastConsensusTime` or `firstConsensusTimeInCurrentFile` is `null`
+
+### Services
+
+Services is to update the processing logic of transactions so that it supports logic for determining new 2-second periods, increment `block` number and keep `block` relevant data.
+The proposed solution specifies a `long` field to be used for the block number counter, incremented every 2 seconds.
```
shemnon, line 70: If we are going to not spell out 2 second blocks I think we need some rationale whu 2^63 is enough and 2^31 isn't for a block counter.

> Using a signed 32-bit int would result in the block number rolling over in 140 years for the initial 2 second length. Sub-second block lengths would exhaust that number well within the operational lifetime of typical networks.  A signed 64 bit integer provides a much longer timeframe.

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+```markdown
+Properties in State:
+- blockHashes `map(number -> keccak256(RunningHash))` - stores the hashes of the last 256 blocks
+- blockNumber `long` - stores the current block number
+- blockTimestamp - `Instant` - stores the timestamp of the block
+
+handleTransaction() {
+	...
+	bool `newBlock` = shouldCreateNewBlock() {
+			if (`currentTS` - `lastConsensusTime` > `1000ns` && `currentTS` - `blockTimestamp` > `2s`) return `true`; else `false`
```
shemnon, line 83: `>=` - greater than or equal? Either this would cause 3 second blocks or a slow drift away from two seconds.  Perhaps express it as a modulus of the preferred block time? Or we consider the floors of the current and block timestamp to be >= 2?

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+- [https://github.com/hashgraph/hedera-state-proof-verifier-go](https://github.com/hashgraph/hedera-state-proof-verifier-go)
+
+### Mirror Nodes
+
+Based on the updates specified above, the record stream objects will pass all the necessary information for Mirror Nodes to build `Record files`/`blocks` and store enough data about them to be able to answer all block related queries outlined above. To do this they will need to collect the `number`, `hash` and `timestamp` specified in the `Record file`. For old record files, they will be derived through a migration process as specified below.
+
+**Historical State**
+
+`Record files` prior to the introduction of the block properties will not expose the required information. In order to be able to answer historical queries, `mirror nodes` will need to iterate over the existing/already imported record files since genesis, assign the corresponding `block number`, `hash` and migrate their DB collections such that they are able to answer queries of type:
```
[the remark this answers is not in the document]
shemnon: Is a file containing the block hash to block hashings out of the question?  It's a one time item after we switch over to storing the hashes in-stream.

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+  | mixHash | Mirror Node | Hardcoded by Mirror Node(s) to 0x |
+  | nonce | Mirror Node | Hardcoded by Mirror Node(s) to hex-encoded 0 |
+  | parentHash | Mirror Node | Computed by Mirror Node(s). The hash of the parent block. |
+  | receiptsRoot | Mirror Node | (???) |
+  | sha3Uncles | Mirror Node | Hardcoded by Mirror Node(s) to the SHA3 computation of empty array (0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347) |
+  | size | Mirror Node | Computed by Mirror Node(s). The sum of the bytes of all transaction(s) of type ContractLocalCall, ContractCall and ContractCreate within the block. transaction is the transaction specified in the RecordStreamObject and not the Transaction Record |
+  | stateRoot | Mirror Node | (???) |
+  | totalDifficulty | Mirror Node | Hardcoded by Mirror Nodes(s) to hex-encoded 0 |
+  | transactions | Mirror Node | Computed by Mirror Node(s) by ingesting the record stream and aggregating RecordStreamObjects of type ContractLocalCall, ContractCall and ContractCreate within the block. |
+  | transactionsRoot | Mirror Node | (???) |
```
shemnon, line 153: This is related to the ethereum consensus calculations and I think we could drop it or echo the block hash, since it is the running transaction hash.

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+## Rejected Ideas
+
+Two iterations were conducted prior to settling on this approach.
+
+1. Rounds as blocks - the first iteration was proposing that each block is the `consensus round`. It became clear that this approach is not suitable as it implied that multiple blocks are to be issued per second. That would add additional load to infrastructure providers when clients are iterating and going through blocks to query for data.
+- Events - the second iteration suggested that `consensus events` are to be defined as `blocks`. The proposal had a much higher cognitive load (1 block = 1 record file is easier to grasp) and it required more changes to the `platform` in order to be implemented. To add on top of that, the same drawback as the `high frequency` blocks was present as well.
+
+## Open Issues
+
+- What will be the considered `baseFeePerGas` property for blocks? There are different pricing for `Create`, `Call` and `LocalCall` transactions.
```
shemnon, line 186: I think we lock it in at zero until we need to use it.  This is a congestion related aspect of Ethereum's block production process (Specificallt EIP-1559), which we don't do.

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+
+`number`, `timestamp` and `hash` (map of the 256 most recent blocks) must be available during transaction execution since the following opcodes are to be supported as per the EVM specification:
+
+- `BLOCKHASH` → Accepts the block `NUMBER` for which to return the `hash`. Valid range is the last `256` blocks (not including the current one)
+- `NUMBER` → Returns the current block number
+- `TIMESTAMP` → Returns the unix timestamp of the current block
+
+**Record File**
+
+Record files are to be updated with a new property appended at the end of the `version 5` record files. The property will be encoded after the `End Object Running Hash` and will be `long` `blockNumber`. It is required for `services` to propagate this property to `mirror nodes` since there are `partial mirror nodes`, that don’t keep the full history from the `first record file`. Due to that, they are unable to calculate the `block number`, thus all other `block properties`.
```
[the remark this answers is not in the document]
shemnon: Yea, I'm not sure if restating prior record files is a good plan.  Could we ship a table that could be used against old data?  Mapping record file hash to "canonical" block numbers?

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+| timestamp | Services | Stored in services state (only the last block timestamp) and computed by services. It is the consensusTimestamp of the first transaction in the record file. It is exposed to the 1) EVM during Transaction Execution through the TIMESTAMP opcode; 2) implicitly exported in the record file through the TS of the first transaction in the record file. |
+| hash | Services | Stored in services (last 256 blocks). It is the keccak256(runningHash) of the previous record file. That is the running hash of the last record stream object from the previous record file. It is exposed to the 1) EVM during Transaction Execution through the BLOCKHASH opcode; 2) In the record file as the End Object Running Hash |
+| baseFeePerGas (??) | Mirror Node | Computed by Mirror Node(s). It is a hex encoded value of the gas pricing based in the FeeSchedule file. |
```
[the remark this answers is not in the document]
shemnon: The meaning of this particular field however deals with EIP-1559 fee burn, which we don't do.
```suggestion
| baseFeePerGas | Mirror Node | Always zero, since there is no EIP-1559 style floating block capacity fees in Hedera. |
```

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+  | difficulty | Mirror Node | Hardcoded by Mirror Node(s) to hex encoded 0 |
```
shemnon, line 138: This should be some function of the block size and gas throttle, as it represents the theoretical max not the actual.
```
| gasLimit | Mirror Node |  Computed by Mirror Node(s). The gas throttle limit per second multiplied by the target block time. |
```

HIP/hip-0000-introduction-of-blocks.md
shemnon: This should be some function of the block size and gas throttle, as it represents the theoretical max not the actual.
```suggestion
  | gasLimit | Mirror Node |  Computed by Mirror Node(s). The gas throttle limit per second multiplied by the target block time. |
```

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+  | extraData | Mirror Node | Hardcoded by Mirror Nodes(s) to 0x |
+  | gasLimit | Mirror Node |  Computed by Mirror Node(s). The sum of the gasLimit values for all ContractLocalCall, ContractCall and ContractCreate transactions within the block. |
+  | gasUsed | Mirror Node | Computed by Mirror Node(s). The sum of the gasUsed value for all ContractLocalCall, ContractCall and ContractCreate transactions within the block. |
+  | logsBloom | Mirror Node | (???) |
```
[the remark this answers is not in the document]
shemnon: My thought is we don't implement it and see if people notice.  We can calculate it on the mirror node if we have all the logs for the transaction.
Here's java code that calculates it for Besu - https://github.com/hyperledger/besu/blob/fd961d5296212568f9c82e7e2d5604a965e69a56/evm/src/main/java/org/hyperledger/besu/evm/log/LogsBloomFilter.java#L110

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+  | miner | Mirror Node | Hardcoded by the Mirror Nodes to the 0.0.98 address in the long zero prefix format.  |
+  | mixHash | Mirror Node | Hardcoded by Mirror Node(s) to 0x |
+  | nonce | Mirror Node | Hardcoded by Mirror Node(s) to hex-encoded 0 |
+  | parentHash | Mirror Node | Computed by Mirror Node(s). The hash of the parent block. |
+  | receiptsRoot | Mirror Node | (???) |
```
shemnon, line 147: This is where we could put our sidecar hash.

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+  | sha3Uncles | Mirror Node | Hardcoded by Mirror Node(s) to the SHA3 computation of empty array (0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347) |
+  | size | Mirror Node | Computed by Mirror Node(s). The sum of the bytes of all transaction(s) of type ContractLocalCall, ContractCall and ContractCreate within the block. transaction is the transaction specified in the RecordStreamObject and not the Transaction Record |
```
shemnon, line 149: I think we don't populate this or return it.  It is supposed to be the size of the Ethereum block as a whole.

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+  | stateRoot | Mirror Node | (???) |
```
shemnon, line 150: I'd leave it out for now.  There may be some value from state proofs we want to put in there in the future.

HIP/hip-0000-introduction-of-blocks.md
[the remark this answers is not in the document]
steven-sheehy: True, I like the file idea better. Forgot you had mentioned that. We can even store that in the mirror node classpath for each public network.

HIP/hip-0000-introduction-of-blocks.md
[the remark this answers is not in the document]
steven-sheehy: My comment isn't about the migration path of older files, but how to structure new record files so as to contain a block number. I do agree for migration of old files that a mapping file would be useful.

HIP/hip-0000-introduction-of-blocks.md
steven-sheehy: I think it would actually be the file size of the record file, which would be trivial to calculate and store.

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+| baseFeePerGas (??) | Mirror Node | Computed by Mirror Node(s). It is a hex encoded value of the gas pricing based in the FeeSchedule file. |
+  | difficulty | Mirror Node | Hardcoded by Mirror Node(s) to hex encoded 0 |
+  | extraData | Mirror Node | Hardcoded by Mirror Nodes(s) to 0x |
+  | gasLimit | Mirror Node |  Computed by Mirror Node(s). The sum of the gasLimit values for all ContractLocalCall, ContractCall and ContractCreate transactions within the block. |
+  | gasUsed | Mirror Node | Computed by Mirror Node(s). The sum of the gasUsed value for all ContractLocalCall, ContractCall and ContractCreate transactions within the block. |
+  | logsBloom | Mirror Node | (???) |
+  | miner | Mirror Node | Hardcoded by the Mirror Nodes to the 0.0.98 address in the long zero prefix format.  |
+  | mixHash | Mirror Node | Hardcoded by Mirror Node(s) to 0x |
+  | nonce | Mirror Node | Hardcoded by Mirror Node(s) to hex-encoded 0 |
+  | parentHash | Mirror Node | Computed by Mirror Node(s). The hash of the parent block. |
```
steven-sheehy, line 146: It's not computed, it's just the previous block's hash.
```suggestion
  | parentHash | Mirror Node | The hash of the previous block. |
```

HIP/hip-0000-introduction-of-blocks.md
Daniel-K-Ivanov: The HIP has been updated to use a 32 byte prefix of the running hash

HIP/hip-0000-introduction-of-blocks.md
Daniel-K-Ivanov: Do mirror nodes have the necessary information to compute that? @tinker-michaelj do we export throttle configurations in any way currently and if we do, @steven-sheehy do you know whether they are ingested by mirror nodes?

HIP/hip-0000-introduction-of-blocks.md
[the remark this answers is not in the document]
Daniel-K-Ivanov: I've updated the section to include a version change to the record file stream

HIP/hip-0000-introduction-of-blocks.md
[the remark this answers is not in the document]
Daniel-K-Ivanov: @shemnon can you elaborate more on the `block hash` to `block hashings` file idea? (We can discuss it on a call if it requires a lot of build-up / context)

HIP/hip-0000-introduction-of-blocks.md
Daniel-K-Ivanov: I think that we should consider precompile contract calls as `internal` transactions to the parent one as @shemnon mentions.

Daniel-K-Ivanov:
@shemnon I am wondering whether for `logsBloom`, `receiptsRoot` and `stateRoot` we should:
1. populate them with `SHA3 computation of empty array`?
2. leave them empty strings?
3. not include them in block responses at all?

HIP/hip-0000-introduction-of-blocks.md
Daniel-K-Ivanov: updated it with @steven-sheehy 's suggestion

HIP/hip-0000-introduction-of-blocks.md
Daniel-K-Ivanov: updated the HIP to set the block hash

HIP/hip-0000-introduction-of-blocks.md
Daniel-K-Ivanov: We are considering that the minimum period of time for a block to be created is `hedera.recordStream.logPeriod` seconds, which currently is set to `2 seconds`. If there is no transaction coming in, record files are not getting exported, thus blocks are not created.

I am not sure that I understand the first part of your comment.

HIP/hip-0000-introduction-of-blocks.md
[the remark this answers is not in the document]
steven-sheehy: @Daniel-K-Ivanov Mirror node will be implementing such a mapping to correct partial mirror node block numbers even before this HIP goes live. I've detailed the algorithm in our [ticket](https://github.com/hashgraph/hedera-mirror-node/issues/3543#issuecomment-1092940604).

HIP/hip-0000-introduction-of-blocks.md
steven-sheehy: The gas throttles are available in network file 0.0.121 app.properties, but only if overridden. So in practice, no, they're not available to mirror node since they're not usually overridden.

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,223 @@
...
+(`0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347`) |
+  | miner | Mirror Node | Hardcoded by the Mirror Nodes to the 0.0.98 address in the long zero prefix format.  |
+  | mixHash | Mirror Node | Hardcoded by Mirror Node(s) to 0x |
+  | nonce | Mirror Node | Hardcoded by Mirror Node(s) to hex-encoded 0 |
+  | parentHash | Mirror Node | The hash of the previous block. |
+  | receiptsRoot | N/A | Hardcoded by Mirror Node(s) to the SHA3 computation of empty array 
+(`0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347`). It is expected to be used and populated in the future once traceability work is finalised.  |
+  | sha3Uncles | Mirror Node | Hardcoded by Mirror Node(s) to the SHA3 computation of empty array (`0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347`) |
+  | size | Mirror Node | Computed by Mirror Node(s). The size of the record file. |
+  | stateRoot | Mirror Node | Hardcoded by Mirror Node(s) to the SHA3 computation of empty array (`0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347`) |
```
steven-sheehy, line 174: Why SHA3 empty array for this and `receiptsRoot` instead of `0x` like `extraData`? It makes sense for `sha3Uncles` since it's a SHA3, but shouldn't these others just be empty array `0x`?

steven-sheehy (changes requested):

HIP/hip-0000-introduction-of-blocks.md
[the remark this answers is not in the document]
Daniel-K-Ivanov: I've updated the section to reflect the description as per the ticket

HIP/hip-0000-introduction-of-blocks.md
Daniel-K-Ivanov: I've updated the HIP with an Open Question related to the initial "feed" of the values.

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,223 @@
...
+- `NUMBER` → Returns the current block number
+- `TIMESTAMP` → Returns the unix timestamp of the current block
+
+**Record File**
+
+Record file version is to be updated to `6`. New `long` property is to be encoded after the `End Object Running Hash` 
+and will be `long` `blockNumber`. It is required for `services` to propagate this property to `mirror nodes` since there
+are `partial mirror nodes`, that don’t keep the full history from the `first record file`. Due to that, they are unable
+to calculate the `block number`, thus all other `block properties`. Block `hash` and `timestamp` will not be included in
+the record files, since `block hash` is the `keccak256(End Object Running Hash)` and `timestamp` is
```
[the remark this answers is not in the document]
Daniel-K-Ivanov: updated

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,223 @@
...
+| hash | Services | Stored in services (last 256 blocks). It is the 32 byte prefix of the runningHash of the previous record file. That is the running hash of the last record stream object from the previous record file. It is exposed to the 1) EVM during Transaction Execution through the BLOCKHASH opcode; 2) In the record file as the End Object Running Hash |
+| baseFeePerGas | Mirror Node | Always zero, since there is no EIP-1559 style floating block capacity fees in Hedera. |
+  | difficulty | Mirror Node | Hardcoded by Mirror Node(s) to hex encoded 0 |
+  | extraData | Mirror Node | Hardcoded by Mirror Nodes(s) to 0x |
+  | gasLimit | Mirror Node |  Computed by Mirror Node(s). The gas throttle limit per second multiplied by the target block time. |
+  | gasUsed | Mirror Node | Computed by Mirror Node(s). The sum of the gasUsed value for all ContractCall and ContractCreate transactions within the block. |
+  | logsBloom | Mirror Node | Hardcoded by Mirror Node(s) to the SHA3 computation of empty array
```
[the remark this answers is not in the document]
Daniel-K-Ivanov: updated

HIP/hip-0000-introduction-of-blocks.md
Daniel-K-Ivanov: updated

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,199 @@
...
+
+The Block concept must not add a lot of complexity and performance overhead to the processing of transactions. It must have minimal impact on the TPS of the network.
+
+Based on the described design goals above, the outlined specification defines that block properties are to be computed and populated at different points in time and by different components of the network, based on their responsibility.
+
+## Specification
+
+**Definitions**
+
+- `block` → `Record file` containing all `Record Stream Objects` grouped in a 2-second time frame.
```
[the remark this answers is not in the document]
steven-sheehy: I think an extra 4 or 8 byte every 2 seconds is not a lot of information. The suggestion by either you or Danno during the initial json-rpc investigation was `Have the services node expose the chainId and for the mirror node to provide it through its API`. If the mirror node should expose it via its blocks API it would be best if it was directly associated with each block instead of in a separate file.

Also, I think it might be useful for security reasons to have the chain/leader ID in the block. It can provide an extra level of verification that mirror nodes configured to talk to mainnet will reject non-mainnet blocks.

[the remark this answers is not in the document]
rbair23: The chainId is the same for the duration of the network. We definitely would not want to resend that in every block. In the (hopefully not distant) future of BLS the ledger ID will literally be the key that signed the block, so it will be included by necessity.

HIP/hip-0000-introduction-of-blocks.md
```diff
@@ -0,0 +1,223 @@
...
+
+`number`, `timestamp` and `hash` (map of the 256 most recent blocks) must be available during transaction execution since the following opcodes are to be supported as per the EVM specification:
+
+- `BLOCKHASH` → Accepts the block `NUMBER` for which to return the `hash`. Valid range is the last `256` blocks (not including the current one)
+- `NUMBER` → Returns the current block number
+- `TIMESTAMP` → Returns the unix timestamp of the current block
+
+**Record File**
+
+Record file version is to be updated to `6`. New `long` property is to be encoded after the `End Object Running Hash`
```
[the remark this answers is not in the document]
Daniel-K-Ivanov: Defining the record file migration to protobufs seems like a standalone HIP. We have started working on HIP-435 (Record Stream V6) which would include the protobuf definition of record files and the side car files that will be exported.

HIP/hip-415.md
```diff
@@ -0,0 +1,255 @@
...
+### Mirror Nodes
+
+Based on the updates specified above, the record stream objects will pass all the necessary information for Mirror Nodes to build `Record files`/`blocks` and store enough data about them to be able to answer all block related queries outlined above. 
+To do this they will need to read and store the `number` specified in the `Record file`. For 
+old record files, they will be derived through a migration process as specified below.
+
+**Historical State / Migration**
+
+`Record files` prior to the introduction of the block properties will not expose the required information. `mirror 
+nodes` that have full sync on `testnet` / `mainnet` networks will export a file containing mapping of record file
```
steven-sheehy, line 151: nit: We're not exporting it but just maintaining a copy in the mirror node's classpath.

HIP/hip-415.md
```diff
@@ -0,0 +1,255 @@
...
+`Record files` prior to the introduction of the block properties will not expose the required information. `mirror 
+nodes` that have full sync on `testnet` / `mainnet` networks will export a file containing mapping of record file 
+consensus end to block number in compressed CSV format:
+```
+...
+1615410294009988000,5548088
+1615410300360028000,5548089
+```
+The file would be generated by executing the SQL `select concat(consensus_end, ',', index) from record_file order by consensus_end asc`
+via psql from `mainnet` and `testnet` databases. It will be taken from mirror nodes that have a full history from stream
```
steven-sheehy, line 159: Would prefer database details be left out as the mirror node database is not a public API that we want to have to maintain in docs.

HIP/hip-415.md
```diff
@@ -0,0 +1,255 @@
...
+
+**Definitions**
+
+- `block` → `Record file` containing all `Record Stream Objects` for a given time frame. Block times are to be 
+  at least `hedera.recordStream.logPeriod` seconds. The genesis block and therefore number (`blockNumber=0`) is 
+  considered the stream start date with the first RCD file exported from `services` nodes.
+- `block number` → consecutive number of the `Record file` that is being incremented by `1` for every new `Record 
+  file`. For already existing networks, this value will be initially bootstrapped through `Mirror Nodes` and after that 
+  maintained by services nodes.
+- `block hash` → `32 byte` prefix (out of `48 bytes`) of the running hash of the last `Record Stream Object` from
```
steven-sheehy, line 59: I confirmed in the mirror node database that the first 32 bytes of the hash are still unique across all rows.

HIP/hip-415.md
```diff
@@ -0,0 +1,255 @@
...
+| hash | Services | Stored in services (last 256 blocks). It is the 32 byte prefix of the runningHash of the previous record file. That is the running hash of the last record stream object from the previous record file. It is exposed to the 1) EVM during Transaction Execution through the BLOCKHASH opcode; 2) In the record file as the End Object Running Hash |
+| baseFeePerGas | Mirror Node | Always zero, since there is no EIP-1559 style floating block capacity fees in Hedera. |
+  | difficulty | Mirror Node | Hardcoded by Mirror Node(s) to hex encoded 0 |
+  | extraData | Mirror Node | Hardcoded by Mirror Nodes(s) to 0x |
+  | gasLimit | Mirror Node |  Computed by Mirror Node(s). The gas throttle limit per second multiplied by the target block time. |
+  | gasUsed | Mirror Node | Computed by Mirror Node(s). The sum of the gasUsed value for all ContractCall and ContractCreate transactions within the block. |
+  | logsBloom | Mirror Node | Computed by Mirror Node(s). It is the bloom filter for the logs within the block. |
+  | miner | Mirror Node | Hardcoded by the Mirror Nodes to the 0.0.98 address in the long zero prefix format.  |
```
steven-sheehy, line 190: Not sure we should populate this as in the future staking accounts 800/801 will also receive block rewards.

HIP/hip-415.md
```diff
@@ -0,0 +1,255 @@
...
+---
+
+## Abstract
+
+Specifies how to introduce and formalize the concepts of Blocks in Hedera Hashgraph so that it can be used as a foundation on which further interoperability with existing DLT networks and infrastructure can be built.
+
+## Motivation
+
+The concept of blocks is a vital part of the existing Ethereum infrastructure and, as such, the introduction of blocks 
+in Hedera can be considered a foundational step towards greater interoperability with EVM based tooling, explorers, exchanges and wallet providers.
```
shemnon, line 26:
```suggestion
The concept of blocks is a vital part of the existing Ethereum infrastructure and, as such, the introduction of  a
standard mapping between existing transaction ordering in Hedera and the notion of blocks can be considered a
foundational step towards greater interoperability with EVM based tooling, explorers, exchanges and wallet
providers.
```

HIP/hip-415.md
```diff
@@ -0,0 +1,255 @@
...
+
+## Rationale
+
+Hedera must have a single consistent answer to what a block number is, the hash and which transactions are included in that block number. This is required for two reasons:
```
shemnon, line 30:
```suggestion
Hedera must have a single consistent answer to what transactions belong to a block and the identifying hash and number for that block.  This is required for two reasons:
```

HIP/hip-415.md
shemnon: Miner implies two things, (a) who gets paid and (b) who decided what transactions went into that block.  Since hedera doesn't really have a single party picking transactions and ordering them for any timeframe an argument can be made there is no miner, so an all-zero address may be more correct and dodges the 0.0.98, 0.0.600, 0.0.601 question.

HIP/hip-415.md
```diff
@@ -0,0 +1,255 @@
...
+  | extraData | Mirror Node | Hardcoded by Mirror Nodes(s) to 0x |
+  | gasLimit | Mirror Node |  Computed by Mirror Node(s). The gas throttle limit per second multiplied by the target block time. |
+  | gasUsed | Mirror Node | Computed by Mirror Node(s). The sum of the gasUsed value for all ContractCall and ContractCreate transactions within the block. |
+  | logsBloom | Mirror Node | Computed by Mirror Node(s). It is the bloom filter for the logs within the block. |
+  | miner | Mirror Node | Hardcoded by the Mirror Nodes to the 0.0.98 address in the long zero prefix format.  |
+  | mixHash | Mirror Node | Hardcoded by Mirror Node(s) to 0x |
+  | nonce | Mirror Node | Hardcoded by Mirror Node(s) to hex-encoded 0 |
+  | parentHash | Mirror Node | The hash of the previous block. |
+  | receiptsRoot | N/A | Hardcoded by Mirror Node(s) to hex-encoded 0. It is expected that in the future it will be 
+populated once traceability is implemented. |
```
shemnon, line 195:
```suggestion
  | receiptsRoot | N/A | Hardcoded by Mirror Node(s) to hex-encoded 0. |
```
I don't expect we will ever populate it. This is core to ethereum's validation loop and since we don't validate in the same way we shouldn't imply it's any reasonable value.  This differes from uncles root, where there are never uncles so an empty list is correct.

HIP/hip-415.md
```diff
@@ -0,0 +1,255 @@
...
+  | sha3Uncles | Mirror Node | Hardcoded by Mirror Node(s) to the SHA3 computation of empty array (`0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347`) |
+  | size | Mirror Node | Computed by Mirror Node(s). The size of the record file. |
+  | stateRoot | Mirror Node | Hardcoded by Mirror Node(s) to hex-encoded 0. May contain state proof data in the future |
```
shemnon, line 198:
```suggestion
| stateRoot | Mirror Node | Hardcoded by Mirror Node(s) to hex-encoded 0. |
```
Lets not imply any future use. It may box us in if our state proof resolves to a single root too.

HIP/hip-415.md
```diff
@@ -0,0 +1,255 @@
...
+  | transactions | Mirror Node | Computed by Mirror Node(s) by ingesting the record stream and aggregating RecordStreamObjects of type ContractCall and ContractCreate within the block. |
+  | transactionsRoot | Mirror Node | The same value as block `hash` |
+  | uncles | Mirror Node | Hardcoded by Mirror Node(s) to empty array [] |
+
+## Backwards Compatibility
+
+The following breaking changes will be introduced with the implementation of the HIP:
+
+- `BLOCKHASH` will no longer return an empty `hash` but the actual `hash` of the block as per the EVM specification.
+- `NUMBER` will no longer return the `timestamp` of the `block`, but rather the proper `block` number as specified in the HIP.
```
shemnon, line 209: So this makes a commitment for us that we will never use the inter-block times for number, and that we will know when the block counter increments prior to the execution of the EVM.

HIP/hip-415.md
Daniel-K-Ivanov: updated

HIP/hip-415.md
Daniel-K-Ivanov: Your argument is valid. I was focused on the (a) part and haven't thought about the (b) part. I am okay with setting it to `0.0.0`

steven-sheehy (approved):
LGTM

HIP/hip-415.md
Daniel-K-Ivanov: Yes, that is correct.
The block updates are occurring at the and of the `handleTransaction` thread post Transaction Execution, at the point where we export the TX Record for streaming. At that point, the check for the new block/record file is executed. If new block/file is to be created, the transaction is handled and exported will be included in the new block and will "close" the previous one. 

We must verify that the correct `timestamp`, `blockhash` and `number` are set and returned during the execution of the first transaction within the block. It might be the case that the execution is performed against the "old" values  (aka parent block data).

HIP/hip-415.md
```diff
@@ -0,0 +1,265 @@
...
+---
+hip: 415 
+title: Introduction of Blocks
+author: Daniel Ivanov<daniel-k-ivanov95@gmail.tech>, Ivan Kavaldzhiev<ivan.kavaldzhiev@limechain.tech>
+working-group: Danno Ferrin<@shemnon>, Richard Bair<@rbair23>, Steven Sheehy<steven.sheehy@hedera.com>, Mitchell 
+Martin<mitch@swirlds.com>
+type: Standards Track 
+category: Core 
+needs-council-approval: Yes 
+status: Draft
```
mgarbs, line 12:
```suggestion
status: Last Call
```

HIP/hip-415.md
```diff
@@ -0,0 +1,262 @@
...
+author: Daniel Ivanov<daniel-k-ivanov95@gmail.tech>, Ivan Kavaldzhiev<ivan.kavaldzhiev@limechain.tech>
+working-group: Danno Ferrin<@shemnon>, Richard Bair<@rbair23>, Steven Sheehy<steven.sheehy@hedera.com>, Mitchell Martin<mitch@swirlds.com>
+type: Standards Track
+category: Core
+needs-council-approval: Yes
+status: Last Call
+created: 2022-03-28
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/434
+last-call-date-time: 2022-05-03
+requires: HIP-435
```
mgarbs, line 13: When 435 gets into master, link to it here [HIP-435](https://hips.hedera.com/hip/hip-435)

HIP/hip-415.md
```diff
@@ -0,0 +1,262 @@
...
+title: Introduction Of Blocks
+author: Daniel Ivanov<daniel-k-ivanov95@gmail.tech>, Ivan Kavaldzhiev<ivan.kavaldzhiev@limechain.tech>
+working-group: Danno Ferrin<@shemnon>, Richard Bair<@rbair23>, Steven Sheehy<steven.sheehy@hedera.com>, Mitchell Martin<mitch@swirlds.com>
+type: Standards Track
+category: Core
+needs-council-approval: Yes
+status: Last Call
+created: 2022-03-28
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/434
+last-call-date-time: 2022-05-03
```
mgarbs, line 12:
```suggestion
last-call-date-time: 2022-05-03 07:00:00 +0000
```

HIP/hip-415.md
```diff
@@ -0,0 +1,262 @@
...
+1. `T - lastConsensusTime > 1000ns`
+2. `T - firstConsensusTimeInCurrentFile > 2s`
+
+or
+
+if `lastConsensusTime` or `firstConsensusTimeInCurrentFile` is `null`
+
+### Services
+
+Services is to update the processing logic of transactions so that it supports logic for determining new record file
```
mgarbs, line 91:
```suggestion
Services are to update the processing logic of transactions so that it supports logic for determining new record file 
```

HIP/hip-415.md
```diff
@@ -0,0 +1,262 @@
...
+2. `T - firstConsensusTimeInCurrentFile > 2s`
+
+or
+
+if `lastConsensusTime` or `firstConsensusTimeInCurrentFile` is `null`
+
+### Services
+
+Services is to update the processing logic of transactions so that it supports logic for determining new record file 
+periods, increment `block` number and keep `block` relevant data. The proposed solution specifies a `long` field to be
```
mgarbs, line 92:
```suggestion
periods, incrementing `block` number and keeping `block` relevant data. The proposed solution specifies a `long` field to be
```

HIP/hip-415.md
```diff
@@ -0,0 +1,262 @@
...
+used for the block number counter, incremented every `hedera.recordStream.logPeriod` seconds. Using a signed 32-bit int
+would result in the block number rolling over in 140 years (if current 2-second length is kept). Sub-second block 
+lengths would exhaust that number well within the operational lifetime of typical networks. A signed 64 bit integer
```
mgarbs, line 95:
```suggestion
lengths would exhaust that number well within the operational lifetime of typical networks. A signed 64-bit integer
```

HIP/hip-415.md
```diff
@@ -0,0 +1,262 @@
...
+`number`, `timestamp` and `hash` (map of the 256 most recent blocks) must be available during transaction execution
+since the following opcodes are to be supported as per the EVM specification:
+
+- `BLOCKHASH` → Accepts the block `NUMBER` for which to return the `hash`. Valid range is the last `256` blocks (not including the current one)
+- `NUMBER` → Returns the current block number
+- `TIMESTAMP` → Returns the unix timestamp of the current block
+
+**Record File**
+
+Once HIP-435 is implemented, record files are to be migrated to protobufs. New `long blockNumber` property is to be
```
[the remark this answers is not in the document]
Daniel-K-Ivanov: I think that HIP-435 will be merged after this PR. Maybe we can update the link after we merge HIP-415 (post-factum)

HIP/hip-415.md
```diff
@@ -0,0 +1,263 @@
+---
+hip: 415
+title: Introduction Of Blocks
+author: Daniel Ivanov<daniel-k-ivanov95@gmail.tech>, Ivan Kavaldzhiev<ivan.kavaldzhiev@limechain.tech>
```
mgarbs, line 4:
```suggestion
author: Daniel Ivanov <daniel.k.ivanov95@gmail.tech>, Ivan Kavaldzhiev <ivan.kavaldzhiev@limechain.tech>
```

HIP/hip-415.md
```diff
@@ -0,0 +1,263 @@
...
+working-group: Danno Ferrin<@shemnon>, Richard Bair<@rbair23>, Steven Sheehy<steven.sheehy@hedera.com>, Mitchell Martin<mitch@swirlds.com>
+type: Standards Track
+category: Core
+needs-council-approval: Yes
+status: Last Call
+created: 2022-03-28
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/434
+last-call-date-time: 2022-05-03 07:00:00 +0000
+requires: HIP-435
```
mgarbs, line 13:
```suggestion
requires: 435
```

HIP/hip-415.md
```diff
@@ -0,0 +1,263 @@
+---
+hip: 415
+title: Introduction Of Blocks
+author: Daniel Ivanov<daniel-k-ivanov95@gmail.tech>, Ivan Kavaldzhiev<ivan.kavaldzhiev@limechain.tech>
+working-group: Danno Ferrin<@shemnon>, Richard Bair<@rbair23>, Steven Sheehy<steven.sheehy@hedera.com>, Mitchell Martin<mitch@swirlds.com>
```
mgarbs, line 5:
```suggestion
working-group: Danno Ferrin <@shemnon>, Richard Bair <@rbair23>, Steven Sheehy <steven.sheehy@hedera.com>, Mitchell Martin <mitch@swirlds.com>
```

HIP/hip-415.md
```diff
@@ -0,0 +1,263 @@
...
+type: Standards Track
+category: Core
+needs-council-approval: Yes
+status: Last Call
+created: 2022-03-28
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/434
+last-call-date-time: 2022-05-03 07:00:00 +0000
```
mgarbs, line 12:
```suggestion
last-call-date-time: 2022-05-03T07:00:00Z
```

mgarbs (changes requested):
Please see comments

shemnon:
LGTM, but addressing @mgarbs changes will clear my approval, so ping me when it's ready.

shemnon (approved):

mgarbs (approved):
LGTM

PR #419, opened 2022-04-06
Add ethereumNonce

shemnon:
**Description**:
Add the current EthereumNonce to the CryptoGetInfoQuery.

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-410.md
+++ b/HIP/hip-410.md
@@ -109,11 +109,14 @@ section of the transaction (also referred to as callData or inputData) to be
 stored as a Hedera file. In these instances the transaction body contains an
 edited version of the Transaction, replacing a zero length string for the `data`
 field in the RLP set in the `ethereumData` message field and storing the call
-data in a Hedera File ?? as a hex string ?? as a binary blob ??. When the
-transaction is executed and before the signatures are extracted
+data in a Hedera File as a hex string. When the transaction is executed and
+before the signatures are extracted the EthereumTransaction bytes will be
+reconstituted to include the bytes from the reference file.
 
 To support this transaction there will also need to be some changes in other
-messages that would be customary for a new transaction body.
+messages that would be customary for a new transaction body. The `ethereumNonce`
+will also be reported in the `AccountInfo` message in the response to 
+`CrytpoGetInfoQuery`.
 
 ```protobuf
 // basic_types.proto 
@@ -137,6 +140,16 @@ message TransactionBody {
         //... 
     } 
     //... 
+}
+
+// crypto_get_info.proto
+message CryptoGetInfoResponse {
+    //...
+    message AccountInfo {
+        //...
+        int64 ethereumNonce = 24;
+    }
+    //...
 }
 ```
 


HIP/hip-410.md
```diff
@@ -109,11 +109,14 @@ section of the transaction (also referred to as callData or inputData) to be
 stored as a Hedera file. In these instances the transaction body contains an
 edited version of the Transaction, replacing a zero length string for the `data`
 field in the RLP set in the `ethereumData` message field and storing the call
```
steven-sheehy, line 111: This HIP should also denote the addition of the 32 byte ethereum transaction hash and its externalization in the record.

HIP/hip-410.md
```diff
@@ -109,11 +109,14 @@ section of the transaction (also referred to as callData or inputData) to be
...
-data in a Hedera File ?? as a hex string ?? as a binary blob ??. When the
-transaction is executed and before the signatures are extracted
+data in a Hedera File as a hex string. When the transaction is executed and
+before the signatures are extracted the EthereumTransaction bytes will be
+reconstituted to include the bytes from the reference file.
 
 To support this transaction there will also need to be some changes in other
-messages that would be customary for a new transaction body.
+messages that would be customary for a new transaction body. The `ethereumNonce`
```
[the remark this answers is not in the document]
shemnon: I'm thinking no?  It's a value in the transaction RLP data and all we would be doing is copying it straight over.  I'm thinking we will want to parse the RLP as part of the ingest anyway to get gasLimit, maxPriorityGas, callData, chainId, value transferred, and target contract.  Or are all of these other fields already part of the transaction record except for nonce?

HIP/hip-410.md
shemnon: This is derivable data.  Just take the ethTxBytes from the transaction and keccak256 it.  Are we adding that to the TransactionRecord too?

HIP/hip-410.md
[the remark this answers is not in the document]
steven-sheehy: Some of those are part of the `ContractFunctionResult`, but not all. But if we can get the nonce out of the ethereum bytes then that should suffice.

HIP/hip-410.md
steven-sheehy: So we only need to hash `ethereumData` and not the contents of the callData file bytes? If so, then that should suffice. Perhaps calling this logic out in a mirror node (or record/block) section might be useful. Basically `ethereumNonce` can be extracted from X and `ethereumHash` can be extracted from Y.

shemnon: So the transaction hash will not be easily extractable unless you can marshal the file for the case of large transactions.   If that is too heavy a lift for the mirror node ingest then we can add it to the record, and then the nonce while we are at it.

HIP/hip-410.md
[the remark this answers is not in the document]
steven-sheehy: Note to future self, can use this code as an example on how to extract nonce: https://github.com/hashgraph/hedera-services/blob/f9af8ad338534275c84564345e77a05a3c9a7594/hedera-node/src/main/java/com/hedera/services/txns/ethereum/EthTxData.java#L60

rbair23 (approved):

HIP/hip-410.md
steven-sheehy: As discussed, I do think we should add `ethereumHash` to the record.

Daniel-K-Ivanov (approved):

PR #420, opened 2022-04-07
Update HIP-260 with side-car records approach

stoqnkpL:
**Description**:

Updates HIP-260 for smart contract traceability with side-car records approach. 
The list of contract state changes is no longer part of contract function result on a ContractCall or ContractCreate transaction, instead it's moved to a new record file capable of tracking optional, related to existing records information.

**Related issue(s)**: https://github.com/hashgraph/hedera-services/issues/3098

Fixes #

**Notes for reviewer**:

**Checklist**


HIP/hip-260.md
```diff
@@ -122,6 +126,63 @@ final updated value of the storage slot after the completion of the smart
...
+
+The sidecar + record file + signature file will all be zipped together into a
+single zip and uploaded to the AWS/Google buckets.
+
+The side car file will have the following structure:
+
+| Name                              | Type (Bytes) |  Description                                                     |
+| --------------------------------- | -----------: | ---------------------------------------------------------------: |
+| Record Stream File Format Version |    int (4)   | "Record stream file format version, Value: 5"                    |
+| Hapi Proto Major Version          |    int (4)   | Matches NetworkGetVersionInfo.hapiProtoVersion Major.Minor.Patch |
```
IvanKavaldzhiev, line 152: I think the Description for the first row has overflowed to the second row and has shifted all other descriptions with one row and now there is a misalignment.

HIP/hip-260.md
```diff
@@ -122,6 +126,63 @@ final updated value of the storage slot after the completion of the smart
...
+
+Record File         (.rcd): 2020-10-19T21_35_39.000000000Z.rcd
+Side car File  (.rcd_side): 2020-10-19T21_35_39.000000000Z.rcd_side
+
+#### Record file structure update
+
+Record files are to be updated with a new property appended at the end of the 
+version 5 record files. The property will be encoded after the long blockNumber
+and will be a byte SideCarHash. This property will hold the value of the
+keccak256(Side car file). It is required in order for the mirror nodes to be
```
IvanKavaldzhiev, line 173: Currently SHA-384 algorithm is used to calculate the running hashes of the Record Stream Objects/Record Files. Wouldn't it be a better idea to also use SHA-384 for the SideCarHash so that the mechanism for validating and calculating hashes is unified.

HIP/hip-260.md
```diff
@@ -122,6 +126,63 @@ final updated value of the storage slot after the completion of the smart
 contract call. Transient states between the start and finish of the contract are
 not stored in the record stream.
 
+### Record Stream File Changes
+
+Currently Hedera nodes produce record stream files that contain a series of
+transactions within a two second interval. The record stream file (.rcd)
```
IvanKavaldzhiev, line 132: Probably we can describe that "an interval with given amount of X seconds" is used to determine the contents of a given record file/block. In this way we would be more compliant if in the future we change the two second interval with four second interval for example.

HIP/hip-260.md
```diff
@@ -122,6 +126,63 @@ final updated value of the storage slot after the completion of the smart
...
+| Hapi Proto Patch Version          |    int (4)   | Minor version: 9;                                                |
+| Object stream version             |    int (4)   | Patch version: 0                                                 |
+| 1st RecordStreamObject            |    byte[]    | Serialized RecordStreamObject bytes.                             |
+| 2nd RecordStreamObject            |    byte[]    | ...                                                              |
+| ...                               |    byte[]    | ...                                                              |
+
+#### Record file names
+
+Side car records will be externalised in a new file with the same name as 
+the main record stream file and the new extension ".rcd_side", e.g.
```
steven-sheehy, line 163: Mirror node would like sidecar files to be put into a separate folder. Doing so will reduce the number of rows we have to search in S3, reduce S3 list response size, and avoid having to filter them out locally. Probably `recordstreams/record0.0.3/sidecar`.

HIP/hip-260.md
steven-sheehy, line 173: To allow multiple sidecars in the future, we should have a list of hashes. Also, we may want some mechanism to indicate what's in the sidecar. Perhaps first sidecar contains contract state changes and second contains flat traces. That way mirror nodes can choose which of the data to download by category.

steven-sheehy (changes requested):

HIP/hip-260.md
steven-sheehy: Possible format:

```protobuf
enum SidecarType {
  ContractStateChange = 1;
  ContractAction = 2;
  // Other future categories
}

message SidecarReference {
  bytes hash = 1;
  long id = 2;
  SidecarType type = 3;
}

message RecordFileObject {
  TransactionRecord record = 1;
  Transaction transaction = 2;
}

message RecordStreamFile {
  ...
  repeated RecordStreamObject stream_objects = ?; 
  repeated SidecarReference sidecars = ?;
}
```

HIP/hip-260.md
```diff
@@ -90,6 +83,17 @@ message StorageChange {
 }
 ```
 
+Transaction ID will be the same as the transaction the side car is related to.
```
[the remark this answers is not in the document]
stoqnkpL: Agreed, changed to consensus timestamp

HIP/hip-260.md
stoqnkpL: Changed to "preconfigured interval"

HIP/hip-260.md
```diff
@@ -122,6 +126,63 @@ final updated value of the storage slot after the completion of the smart
...
+
+#### Side car records file
+
+A new file of type (.rcd_side) will be introduced to track all side car records
+produced during the same 2 second period as the main record file (.rcd). All
+side car records in the .rcd_side file should have a corresponding main record
+with the same id in the main record file (.rcd).
+
+The sidecar + record file + signature file will all be zipped together into a
+single zip and uploaded to the AWS/Google buckets.
```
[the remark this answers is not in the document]
stoqnkpL: The idea is to have a lambda running on AWS/Google that will unzip and store the files separately. Added more info about this in the HIP

HIP/hip-260.md
```diff
@@ -122,6 +126,63 @@ final updated value of the storage slot after the completion of the smart
...
+Side car records will be externalised in a new file with the same name as 
+the main record stream file and the new extension ".rcd_side", e.g.
+
+Record File         (.rcd): 2020-10-19T21_35_39.000000000Z.rcd
+Side car File  (.rcd_side): 2020-10-19T21_35_39.000000000Z.rcd_side
+
+#### Record file structure update
+
+Record files are to be updated with a new property appended at the end of the 
+version 5 record files. The property will be encoded after the long blockNumber
```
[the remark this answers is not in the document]
stoqnkpL: > I don't believe adding a field at the end of a v5 file is backwards compatible. So we should create a v6 record file.

Richard wanted to bundle the updates for the block HIP and sidecar HIP together and have a single update to the record file, that's why I proposed using v5.
> Richard had said he wanted to use protobuf for the new file.

Not sure what that implies.
> We should also keep in mind the block concept HIP that will be creating a new version and coordinate the two.

AFAIK the block concept HIP adds the `blockNumber`, that's why I specified here that the `SideCarHash` will be encoded after the `blocknumber` property.

HIP/hip-260.md
```diff
@@ -122,6 +126,68 @@ final updated value of the storage slot after the completion of the smart
...
+#### Side car records file
+
+A new file of type (.rcd_side) will be introduced to track all side car records
+produced during the same preconfigured interval as the main record file (.rcd). All
+side car records in the .rcd_side file should have a corresponding main record
+with the same id in the main record file (.rcd).
+
+The sidecar + record file + signature file will all be zipped together into a 
+single zip and uploaded to the AWS/Google buckets. A lambda running on AWS/Google 
+will unzip and store the files separately. Mirror nodes will have the option as
```
steven-sheehy, line 146: This process doesn't sound very decentralized. Even though the bucket is owned by Hedera, only individual nodes should have permission to write to their individual paths within the bucket. This will become important once we have community and permissionless nodes. Having a lambda that can write to paths that only nodes should write to sounds centralized.

Instead, I would suggest that the record file and sidecar be gzipped on the node and uploaded separately to their final position. The signature is not worth gzipping separate as it is actually smaller if you don't zip it. This will ensure a similar amount of overall bytes is uploaded while still being decentralized.

HIP/hip-260.md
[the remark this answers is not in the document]
steven-sheehy: I'm saying that adding fields to the end of the v5 format would necessitate a v6 format since it's not backwards compatible to clients. And if we're doing a new version anyway we should upgrade the format from its current complex, proprietary binary format to our usual standard of using protobuf.

HIP/hip-260.md
```diff
@@ -53,31 +53,24 @@ transactions.
...
 
-### Add ContractStateChange to ContractFunctionResult
+### Side car record
 
-Add and populate a new repeated field `stateChanges` of
-type `ContractStateChange` to the `ContractFunctionResult` message. This will
-be populated whenever a transaction causes contract storage states to change.
-This field will include the changes across all contract storage and not just the
-initial contract in the transaction.
+A side car record will be created for each ContractCall and ContractCreate
```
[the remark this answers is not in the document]
stoqnkpL: I agree, we need some way to distinguish between a `record` and a `record file`. Maybe specifying that each `ContractCreate` and `ContractCall` transaction will create an entry in a list of side car records in a record file?

HIP/hip-260.md
```diff
@@ -122,6 +126,63 @@ final updated value of the storage slot after the completion of the smart
...
+| 2nd RecordStreamObject            |    byte[]    | ...                                                              |
+| ...                               |    byte[]    | ...                                                              |
+
+#### Record file names
+
+Side car records will be externalised in a new file with the same name as 
+the main record stream file and the new extension ".rcd_side", e.g.
+
+Record File         (.rcd): 2020-10-19T21_35_39.000000000Z.rcd
+Side car File  (.rcd_side): 2020-10-19T21_35_39.000000000Z.rcd_side
```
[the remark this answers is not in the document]
stoqnkpL: My idea was to support different types of side car records but they would all go in the same side car **record file**. That's why I have defined:
```
message SideCar {
    Timestamp consensusTimestamp = 1;
    oneof data {
        ContractTraceabilitySideCar = 2;
    }
}
```

the record file can be defined like this:

```
message RecordFile {
    repeated SideCar = 1;
}
```
Future growth can be achieved by defining new types of SideCar messages.

But if the mirror node wants to have the option to download only certain types of side car records, it makes sense to externalise them in separate files and have some mapping between zero-padded sequence number <> side car types.

HIP/hip-260.md
stoqnkpL: This should not be a problem when looking at the rendered markdown

PR #435, opened 2022-04-13
HIP-435 Record Stream V6

stoqnkpL:
**Description**:

Introduce Sidecar Transaction Records and sidecar record files.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


HIP/hip-435.md
```diff
@@ -0,0 +1,131 @@
...
+
+`TransactionSideCarRecord` is a protobuf which will be used to create sidecar records complementing `TransactionRecord` and storing additional information about a transaction execution.
+The `TransactionSideCarRecord`s created by Hedera nodes will be externalised in a new file, part of the new RecordStream structure described bellow.
+
+```
+message TransactionSideCarRecord {
+    // Consensus timestamp will be the same as the consensus timestamp of the 
+    // transaction the side car is related to. This offers a convenient 
+    // way to match record to sidecar. 
+    Timestamp consensusTimestamp = 1;
```
shemnon, line 64: So this is not a one-to-one index, correct?  We can have a sidecar message for storage change and then another sidecar message for contract actions?

HIP/hip-435.md
```diff
@@ -0,0 +1,131 @@
...
+
+message RecordStreamFile {
+  bytes startObjectRunningHash = 1;
+  repeated RecordFileObject fileObjects = 2;
+  bytes endObjectRunningHash = 3;
+  uint64 blockNumber = 4;
+  repeated bytes SideCarFileHash = 5; //allows multiple sidecar files to be linked to this RecordStreamFile
+}
+
+message SideCarFile {
```
shemnon, line 91: Do we want only one record per sidecar file?  Or is the sidecar file a stream of side car records?

HIP/hip-435.md
```diff
@@ -0,0 +1,131 @@
...
+  TransactionRecord record = 1;
+  Transaction transaction = 2;
+}
+
+message RecordStreamFile {
+  bytes startObjectRunningHash = 1;
+  repeated RecordFileObject fileObjects = 2;
+  bytes endObjectRunningHash = 3;
+  uint64 blockNumber = 4;
+  repeated bytes SideCarFileHash = 5; //allows multiple sidecar files to be linked to this RecordStreamFile
```
shemnon, line 88: Do we want to identify which kind of sidecar file goes with which hash?  Perhaps a repeated struct with two fields, one is an enum (repeated?) of the kinds of data stored in that sidecar and then the hash?  

I don't think we want to bake the name in but I do think we want to bake in the kind of content it has. That way if we decide that contract actions are too much data we can only emit the state changes and put that in as part of the record stream?

HIP/hip-435.md
```diff
@@ -0,0 +1,131 @@
...
+```
+message RecordFileObject {
+  TransactionRecord record = 1;
+  Transaction transaction = 2;
+}
+
+message RecordStreamFile {
+  bytes startObjectRunningHash = 1;
+  repeated RecordFileObject fileObjects = 2;
+  bytes endObjectRunningHash = 3;
```
shemnon, line 86: The running hash only covers the fileObjects and there is no cryptographic integrity check for the side car files?

HIP/hip-435.md
stoqnkpL: Yes, and both sidecar messages will have the same timestamp since they're related to the same transaction record.

HIP/hip-435.md
stoqnkpL: The SideCarFile will contain all sidecar records of a given type (e.g. ContractAction, StateChanges) that were produced during the same ~2s period that the record file was produced. In other words all the sidecar records relating to transaction records in the record file for that period. This also means that multiple SideCarFiles will be produced per RecordStreamFile - 1 for ContractAction sidecars, 1 for StateChanges, 1 for each new type of sidecar records we introduce in the future.

HIP/hip-435.md
```diff
@@ -0,0 +1,131 @@
+---
+hip: 435
+title: Record Stream V6
+author: Stoyan Panayotov <stoyan.panayotov@limechain.tech>
+type: Standards Track 
+category: Service
+needs-council-approval: Yes
+status:
```
mgarbs, line 8:
```suggestion
status: Draft
```

mgarbs (changes requested):
Please see comments

HIP/hip-435.md
stoqnkpL: The record signature file that is created for each record stream file signs the hash of the entire corresponding stream file. The list of sidecar file hashes is included in the record stream file. This way mirror nodes or any interested party can download the record stream file and all sidecar files and verify that:
1. repeated bytes SideCarFileHash is correct; 
2. the signature file signed the correct hash of the entire record stream file;

Added this to the HIP as well.

stoqnkpL: I didn't want to add the list of sidecar hashes to the RecordFileObject itself in order to avoid requiring the mirror nodes to download that information as well.

HIP/hip-435.md
stoqnkpL: The way I figured this would work is:
The transaction and sidecar records themselves (e.g. `RecordFileObject` and `TransactionSideCarRecord`) will be created on the services side and sent to the platform in a `List<RecordFileObject>` and a `Map<int, List<TransactionSideCarRecord>>`. The `int` in the map will specify the type of the sidecar record and the list will be used to group all sidecars of the same type in the same file in a generic way. The platform will then create the files using the new protobufs `RecordStreamFile` and `SideCarFile` and put the int in each sidecar file name. This way in the future when we add new types of sidecars we shouldn't have to make a code change on the platform side. 
So if we wanted to stop externalising `ContractActions` for example we would just need to stop sending that type of sidecar records to platform through the map. We can even introduce feature flags for each type of sidecar and turn them off and on.

HIP/hip-435.md
```diff
@@ -0,0 +1,178 @@
...
+      repeated ContractAction = 3; // Example sidecar type
+      // Other future categories
+    }
+}
+```
+
+`TransactionSideCarRecord` is a protobuf message which will be used to create sidecar records complementing `TransactionRecord` and storing additional information about a transaction execution. 
+
+```
+message SideCarFile {
```
steven-sheehy, line 111: The `c` in Sidecar should not be capitalized. Need to adjust naming here and everywhere else.

```suggestion
message SidecarFile {
```

HIP/hip-435.md
```diff
@@ -0,0 +1,178 @@
...
+
+  // Running Hash of all RecordFileObjects before closing this file.
+  bytes endObjectRunningHash = 3; 
+
+  // The block number associated with this period.
+  uint64 blockNumber = 4; 
+
+  // List of the hashes of all the Sidecar record files created for the same period.
+  // Allows multiple sidecar files to be linked to this RecordStreamFile
+  repeated bytes SideCarFileHash = 5;
```
steven-sheehy, line 85: Instead of just a hash we should add the sidecar index and type information here so that clients can choose which files to download. This is preferable to having type information encoded in the filename. Potentially the index can be omitted and be inferred from the list position.

```suggestion
  repeated SidecarMetadata sidecars = 5; 
```

```proto
enum SidecarType {
  ContractStateChange = 1;
  ContractAction = 2;
  // Other future categories
}

message SidecarMetadata {
  bytes hash = 1;
  long id = 2;
  repeated SidecarType types = 3;
}
```

steven-sheehy (changes requested):

HIP/hip-435.md
```diff
@@ -0,0 +1,196 @@
...
+    }
+}
+```
+
+`TransactionSidecarRecord` is a protobuf message which will be used to create sidecar records complementing `TransactionRecord` and storing additional information about a transaction execution. 
+
+```
+message SidecarFile {
+  // Version of HAPI that was used to serialise the file
+  SemanticVersion hapiProtoVersion = 1;
```
steven-sheehy, line 128: Don't think this is needed since the parent RecordStreamFile has this info already

HIP/hip-435.md
```diff
@@ -0,0 +1,178 @@
...
+The nano-of-second always outputs nine digits with padding when necessary, to ensure same length filenames and proper sorting.
+
+The sidecar record files will have the same format but will also include the number representing the type of sidecar records that make up that file. This provides an easy way for mirror nodes to download only those files that contain sidecar records they're interested in.
+
+Examples of a record stream file name with corresponding signature file and sidecar record files:
+
+Example 1:
+Record File (.rcd): 2020-10-19T21_35_39.000000000Z_432.rcd
+Sidecar Record File 1: 2020-10-19T21_35_39.000000000Z_432_1.rcd
+Sidecar Record File 2: 2020-10-19T21_35_39.000000000Z_432_2.rcd
```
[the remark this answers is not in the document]
stoqnkpL: Actually, the `432` in this example isn't the sidecar type, it's the block number. Should we also drop the blockNo? The corresponding rec file is named: 
Record File (.rcd): 2020-10-19T21_35_39.000000000Z_432.rcd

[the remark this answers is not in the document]
steven-sheehy: Right, we don't need the block number in the filename since it's within the contents.

HIP/hip-435.md
```diff
@@ -0,0 +1,178 @@
...
+A record file name is created using a string representation of the Instant of consensus timestamp of the first transaction in the file using ISO-8601 representation, with colons converted to underscores for Windows compatibility. The block number will be appended to the record file name.
+
+The nano-of-second always outputs nine digits with padding when necessary, to ensure same length filenames and proper sorting.
+
+The sidecar record files will have the same format but will also include the number representing the type of sidecar records that make up that file. This provides an easy way for mirror nodes to download only those files that contain sidecar records they're interested in.
+
+Examples of a record stream file name with corresponding signature file and sidecar record files:
+
+Example 1:
+Record File (.rcd): 2020-10-19T21_35_39.000000000Z_432.rcd
```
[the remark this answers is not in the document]
steven-sheehy: This was resolved but I see no mention of compression in the updated HIP.

HIP/hip-435.md
```diff
@@ -0,0 +1,247 @@
...
+
+The `TransactionSidecarRecord`s created by Hedera nodes will be externalised in `SidecarFile` files grouped by the type of the records. Files will be created using the SidecarFile protobuf holding all the `TransactionSidecarRecords` of a same type (e.g. `ContractStateChange`, `ContractAction`) related to `RecordFileObject`s created for the same period.
+
+The `SidecarFile`s will be uploaded into a separate folder in the cloud bucket. Doing so will reduce the number of rows Mirror nodes have to search in S3, reduce S3 list response size, and avoid having to filter them out locally. Probably `recordstreams/record0.0.3/sidecar`.
+
+```
+message SignatureFile {
+  
+  // version of HAPI that was used to serialise this file.
+  int fileVersion = 1;
```
steven-sheehy, line 156: This wouldn't be related to HAPI version since that tracks services tags. Record file versions change at an infrequent rate. This would be set to 6. 

```suggestion
  // The signature file version that was used to serialise this file.
  int version = 1;
```

steven-sheehy (changes requested):

HIP/hip-435.md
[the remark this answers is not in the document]
stoqnkpL: There is a mention of this topic after the example file names:
"All record, sidecar, and signature files will be compressed."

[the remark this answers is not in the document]
steven-sheehy: Ah, there was a typo so my search didn't find it originally.

HIP/hip-435.md
```diff
@@ -0,0 +1,243 @@
...
+}
+```
+
+`RecordStreamFile` will be used to serialize all `RecordFileObject`s that are part of the same period into record stream files. Periods are intervals of `hedera.recordStream.logPeriod` seconds used to group `RecordFileObject`s into record stream files.
+The block structure is similar to the current  structure of record files, block number and sidecars are new fields.
+
+```
+enum SidecarType {
+  ContractStateChange = 0;
+  ContractAction = 1;
```
steven-sheehy, line 102: Another category we probably need right away is `ContractBytecode` to contain both the initcode and runtime bytecode.

steven-sheehy (changes requested):

steven-sheehy (approved):
LGTM

mgarbs (dismissed):
LGTM

stoqnkpL:
Force pushed to fix DCO

steven-sheehy (approved):

mgarbs (approved):
LGTM

abonander:
@stoqnkpL @mgarbs  I wish I got to this before it was merged but the format version being inside the protobuf definition is problematic:

Existing consumers parsing the v2 and v5 record formats are going to be interpreting the first 4 bytes of the file as the version integer, but the Protobuf encoding of v6 will add the tag and wire type first before the version number. 

In addition, the encoding order of protobuf fields is not defined. The version could theoretically appear anywhere in the file. Just because the Java implementation encodes fields in declaration order in one version doesn't mean that will stay the same: https://developers.google.com/protocol-buffers/docs/encoding#order

This means that a consumer of the record stream, especially mainnet, will have to either hardcode the timestamp at which records switch to v6 (which is likely not possible to determine ahead of time), assume an invalid version integer is just a v6 protobuf and switch to parsing that, or attempt to parse the entire record file as the v6 format first then fall back to v2/v5 parsing, which will slow down ingestion considerably.

Additionally, since determining the format version post-v6 will require parsing the Protobuf, it means backwards-incompatible changes to the format will be much more difficult to handle in the future. It would possibly require parsing the whole file twice, once just to find the version integer.

I recommend keeping the format version as a 4-byte integer at the beginning of the file and then adding the encoded protobuf bytes after it.

stoqnkpL:
> @stoqnkpL @mgarbs I wish I got to this before it was merged but the format version being inside the protobuf definition is problematic:
> 
> Existing consumers parsing the v2 and v5 record formats are going to be interpreting the first 4 bytes of the file as the version integer, but the Protobuf encoding of v6 will add the tag and wire type first before the version number.
> 
> In addition, the encoding order of protobuf fields is not defined. The version could theoretically appear anywhere in the file. Just because the Java implementation encodes fields in declaration order in one version doesn't mean that will stay the same: https://developers.google.com/protocol-buffers/docs/encoding#order
> 
> This means that a consumer of the record stream, especially mainnet, will have to either hardcode the timestamp at which records switch to v6 (which is likely not possible to determine ahead of time), assume an invalid version integer is just a v6 protobuf and switch to parsing that, or attempt to parse the entire record file as the v6 format first then fall back to v2/v5 parsing, which will slow down ingestion considerably.
> 
> Additionally, since determining the format version post-v6 will require parsing the Protobuf, it means backwards-incompatible changes to the format will be much more difficult to handle in the future. It would possibly require parsing the whole file twice, once just to find the version integer.
> 
> I recommend keeping the format version as a 4-byte integer at the beginning of the file and then adding the encoded protobuf bytes after it.

Hey @abonander, thank you for your comment. We did tackle the 4 bytes for the version of the record file problem and here is what we found:
- protobufs have a fixed32 type which always encodes in 4 bytes
- I wrote a small java application that serializes a message starting with a fixed32 scalar type and writes it in a file

When reading the first 4 bytes of the file:
- the first byte always has the value 13, I think that's the field descriptor
- the second byte has the value that was set to the `fixed32` field (for smaller values used)
- bytes 3 and 4 have the value 0 (at least for lower values used to initialize the field)

So we did think it should be possible to use the `fixed32` type for the version property. 
What you mentioned about the encoding order is worrisome though and makes me think we may need to keep the format version as a 4-byte integer at the beginning of the file and then add the encoded protobuf bytes after it, as you recommend.
@steven-sheehy does it sound good to you? I can update the HIP tomorrow if this is the agreed approach.

abonander:
That is the encoding of the field, yes. The first byte is `(1 << 3) | 5` or `0b00001101` where the tag is left-shifted and then the wire type `5` ( `fixed32bit`) is OR'd in. The next four bytes are the 32-bit value in little-endian (thus the value appearing in the first byte followed by three zero-bytes).

The encoding order is an issue since it is _explicitly_ not defined. It may be the same as declaration order for the current version of the Java implementation the node is using, but that can theoretically change at any time.

I do want to reiterate the other issue, the inability to make breaking changes to the format later, since the file would need to be fully parsed successfully to determine the format version.

If for some reason v6 turns out to be not the final record file format version, it's best to make choices that minimize the future implementation complexity of consumers.

Note that removing a field is not a breaking change as long as the default value for its type in the Protobuf spec is acceptable, but reusing its tag for a different field is a breaking change as is changing the tags for existing fields.

steven-sheehy:
I wasn't aware the encoding order was not defined. So I agree we definitely should add a 4-byte integer to the beginning before the protobuf bytes.

xin-hedera:
The protobuf nondeterministic serialization issue also affects the hash of the record file itself. In case enough nodes come up with different serialized bytes for the content-wise same record file, the record file hashes stored in the signature files will differ too. As a result, for mirror nodes, the processing of record stream files will come to a stop.

stoqnkpL:
Hey @abonander, @steven-sheehy, @xin-hedera,
I opened [this PR](https://github.com/hashgraph/hedera-improvement-proposal/pull/476) with some updates to the HIP. Mainly removing the version property from the protobuf definitions.
I still need to think about the problem with hashing nondeterministically serialized files... this one is not resolved 😞 
I'm not allowed to add reviewers to the PR, that's why I'm sharing it here.

PR #478, opened 2022-05-17
Interoperability Between Smart Contracts and the HCS

scalemaildev:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


hip-0000-smart-contract-hcs-interoperable.md
```diff
@@ -0,0 +1,74 @@
...
+type: Standards Track
+category: Core
+needs-council-approval: <Yes | No>
+status: Draft
+created: 05-17-2022
+discussions-to: <a URL pointing to the official discussion thread>
+updated: <comma separated list of dates>
+requires: <HIP number(s)>
+replaces: <HIP number(s)>
+superseded-by: <HIP number(s)>
```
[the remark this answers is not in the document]
mgarbs: Revisit this default text

hip-0000-smart-contract-hcs-interoperable.md
```diff
@@ -0,0 +1,82 @@
...
+As a developer, I want to be able to create HCS topics from my smart contracts.
+  
+## Specification
+
+Enabling smart contracts to interact with HCS topics would follow the same approach as the [https://hips.hedera.com/hip/hip-206](HTS pre-compiled contracts HIP).
+
+## Backwards Compatibility
+
+All HIPs that introduce backward incompatibilities must include a section describing these incompatibilities and their severity. The HIP must explain how the author proposes to deal with these incompatibilities. HIP submissions without a sufficient backward compatibility treatise may be rejected outright.
+
```
mgarbs, line 48: Default content

hip-0000-smart-contract-hcs-interoperable.md
```diff
@@ -0,0 +1,82 @@
...
+## Rejected Ideas
+
+Throughout the discussion of a HIP, various ideas will be proposed which are not accepted. Those rejected ideas should be recorded along with the reasoning as to why they were rejected. This both helps record the thought process behind the final version of the HIP as well as preventing people from bringing up the same rejected idea again in subsequent discussions.
+
+In a way, this section can be thought of as a breakout section of the Rationale section that focuses specifically on why certain ideas were not ultimately pursued.
+
+## Open Issues
+
+While a HIP is in draft, ideas can come up which warrant further discussion. Those ideas should be recorded so people know that they are being thought about but do not have a concrete resolution. This helps make sure all issues required for the HIP to be ready for consideration are complete and reduces people duplicating prior discussions.
```
mgarbs, line 72: Default content

mgarbs (changes requested):
Let's do something with this default content. Either keep it if we need it or remove it

mgarbs:
We need to fix the DCO check!

scalemaildev:
@sergmetelin Please sign commit.

scalemaildev:
@mgarbs he seems to have disappeared is there some way we can roll back his commit and I'll just copy his changes in?

scalemaildev:
I am not seeing any comments or change requests other than removing the default content.

mgarbs:
> I am not seeing any comments or change requests other than removing the default content.

That was my fault. Sorry about that. You should be able to see my comments now 😃

hip-0000-smart-contract-hcs-interoperable.md
```diff
@@ -0,0 +1,58 @@
...
+## Specification
+
+Enabling smart contracts to interact with HCS topics would follow the same approach as the [https://hips.hedera.com/hip/hip-206](HTS pre-compiled contracts HIP).
+
+## Security Implications
+
+With regards to signing transactions from within a smart contract, the main concern is storing private keys in-state. So, some possible solutions are:
+
+1. Give a different kind of permission to deployed smart contracts for interacting with certain topics. This could function similarly to how AWS differentiates between key-based permissions and role-based permissions.
+2. Follow the same approach from the [https://hips.hedera.com/hip/hip-206](HTS pre-compiled contracts HIP) to signing transactions and interacting with topics.
```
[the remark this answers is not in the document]
walter-hernandez: @mgarbs A topic needs an [admin key for access control to update/delete a topic and a submit key for access control of submitMessage](https://docs.hedera.com/guides/docs/sdks/consensus/create-a-topic).  Slightly similar, to mint a token in Hedera and to manage the ownership and access control of other functionality of a token, it is possible to pass an [admin, supply, pause and freeze keys to a token being minted in Hedera using HTS pre-compiled contracts](https://github.com/hashgraph/hedera-smart-contracts/blob/main/hts-precompile/TokenCreateContract.sol):

image

image

Then, the suggestion is to take advantage of the implementation of HTS pre-compiled smart contracts for handling and managing of keys in the implementation of this HIP for interoperability of Smart Contracts and HCS. So, in the case of an smart contract deployed in Hedera and which needs to read from a topic, which uses HCS in the background for ordering and validity of its messages, the smart contract can be update or read directly from the topic using the same approach for keys usage that HTS pre-compiled smart contracts do for minting of tokens

Of course, the issue would be the storing of private keys in the state, which is highlighted in the HIP. However, I think in HTS pre-compiled smart contracts, private keys storing in the state may have been something that may have already been considered and addressed

All of the above is in the case of using keys. Another option to allow interaction of smart contracts with topics is to implement role-based access:
1. Smart contract A could have a role for only reading from a topic
2. Smart contract B could have a role for only updating a topic

Both smart contracts could have functions to interact with the topic. These functions could be only accessible by the owner of the contract or approved addresses. These functions could also be triggered externally with scheduled transactions, which could open the door to other possible use cases (schedule a transaction to trigger a function in a contract that updates a topic from which other contracts are listening to)

Let me know what you think and what you see feasible and/or if there are better approaches

hip-0000-smart-contract-hcs-interoperable.md
```diff
@@ -0,0 +1,58 @@
...
+
+1. By enabling smart contracts to interact with the HCS, this would allow the creation of HCS-based Oracles. For example, a smart contract could subscribe to a private or public topic in order to read its external data, such as the prices of crypto-assets.
+
+2. This would also enable topic-based NFT minting rights. A smart contract could ingest an HCS topic to calculate the user's available NFTs. Then, after minting the NFT and transferring it to the user, the smart contract could update that same HCS topic. Currently, this process is only achievable via a centralized, server-based chain of events (such as on [https://hashandslash.io](hashandslash.io)).
+
+3. Any other Oracle use case could be repurposed by the HCS.
+
+## User stories
+
+As a developer, I want to be able to read HCS topics from my smart contracts.
```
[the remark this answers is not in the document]
walter-hernandez: @mgarbs Agree, all the data related part of oracles is a process that would be outside Hedera Smart Contracts Service because of the computing needed for data collection, cleaning, and calculation of most common values and/or median (in the case of numerical values) from the data collected. 

Now, the idea of reading HCS topics within Smart Contracts is to bring the data from outside Hedera's network into it in order that Hedera Smart Contract Service can expand the functionality that its contracts can have. For example, a function within a contract can get the latest cryptoasset prices from an Oracle Smart Contract and pays to this Oracle Smart Contract for it. The Oracle smart contract gets the latest cryptoasset prices by reading a private topic, which is updated from outside Hedera network

Let me know what you think and if what you see it feasible and/or if there are better approaches

hip-0000-smart-contract-hcs-interoperable.md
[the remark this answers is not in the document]
mgarbs: >Of course, the issue would be the storing of private keys in the state, which is highlighted in the HIP. I think in HTS pre-compiled smart contracts, private keys storing in the state may have been something that may have already been considered and addressed

I am not sure I understand what you mean here.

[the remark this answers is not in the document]
mgarbs: These extra details look great! Can you add them to the hip too?

hip-0000-smart-contract-hcs-interoperable.md
```diff
@@ -0,0 +1,58 @@
...
+
+As a developer, I want to be able to create HCS topics from my smart contracts.
+  
+## Specification
+
+Enabling smart contracts to interact with HCS topics would follow the same approach as the [https://hips.hedera.com/hip/hip-206](HTS pre-compiled contracts HIP).
+
+## Security Implications
+
+With regards to signing transactions from within a smart contract, the main concern is storing private keys in-state. So, some possible solutions are:
```
Nana-EC, line 45: The Hedera network does not store private keys.
I see you noted the keys passed in during HTS create flows in a comment below. Note those are public keys and allow the network to verify that any submitted transaction was signed by the appropriate private key without storing the private key.

Can you clarify why you fee private keys need to be stored?

hip-0000-smart-contract-hcs-interoperable.md
```diff
@@ -0,0 +1,58 @@
...
+
+As a developer, I want to be able to read HCS topics from my smart contracts.
+
+As a developer, I want to be able to send HCS topic messages from my smart contracts.
+
+As a developer, I want to be able to create HCS topics from my smart contracts.
+  
+## Specification
+
+Enabling smart contracts to interact with HCS topics would follow the same approach as the [https://hips.hedera.com/hip/hip-206](HTS pre-compiled contracts HIP).
```
Nana-EC, line 41: Like the noted HIP 206 can you call out the desired  pre-compile functions you're looking to be added, noting the input and output params in a similar fashion.
You can leave the hash empty to start with.

It would also be good to note any additional structs if you have any in mind.

hip-0000-smart-contract-hcs-interoperable.md
```diff
@@ -0,0 +1,58 @@
...
+
+Enable smart contract interaction with the Hedera Consensus Service, by allowing contracts to create, read, and write to HCS topics.
+
+## Motivation
+
+Interoperability between the HCS and smart contracts would enable developers to leverage HCS-stored data in their dApps.
+
+## Rationale
+
+1. By enabling smart contracts to interact with the HCS, this would allow the creation of HCS-based Oracles. For example, a smart contract could subscribe to a private or public topic in order to read its external data, such as the prices of crypto-assets.
```
Nana-EC, line 25: Can you elaborate on the topic subscription.
Currently that is offered through a gRPC service on the mirror node only.
How would you envision it be supported at the smart contract layer?

hip-0000-smart-contract-hcs-interoperable.md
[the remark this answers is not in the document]
Nana-EC: @walter-hernandez mgarbs is correct.
The smart contract at most would be able to read the topic entity information, submit a topic message but will not have access to individual topic messages already emitted.

Are you suggesting network changes to have access to past messages?

Nana-EC (changes requested):
Thanks for the HIP, nice idea. Interesting rationale examples.
However, given the gap that exists I think the HIP needs more information and suggestions on how items around subscription and even topic message retrieval in general would be achieved

hip-0000-smart-contract-hcs-interoperable.md
walter-hernandez: @Nana-EC My thinking is that it could be:

- A topic subscription could be a combination of scheduled transactions with Topics. For example, a transaction could be scheduled to read the latest message in a topic and send the message to a smart contract. The scheduled transactions are the middleman to carry the message from the topic to the function in the smart contract

Of course, the limitations above is:

- I think Hedera's Scheduled Transactions may not have interoperability with Hedera Smart Contract Service
- In the scenario of Hedera's Scheduled Transactions being possible to integrate with Hedera Smart Contract Service, there could be some difficulties for signing of messages sent from Hedera's Scheduled Transactions to  Hedera Smart Contract Service

The approach of using gRPC at the mirror node works quite well for topic subscription and integration with a solution outside of Hedera Smart Contract Service. 

Let me know what you and if you know if there is a better way to do it or if my thinking/assumption may be flawed

hip-0000-smart-contract-hcs-interoperable.md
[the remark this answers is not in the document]
walter-hernandez: @mgarbs thank you for your comments and suggestions about help to get something working! :-)

@Nana-EC In the case of past messages, an Oracle Smart Contract could be used for storing some of the topic messages that it has received from the topic and it could have logic to delete/update messages as needed

hip-0000-smart-contract-hcs-interoperable.md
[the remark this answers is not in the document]
walter-hernandez: @mgarbs will add them after hearing from @Nana-EC and anyone else about the feasibility of topics interaction with smart contracts

hip-0000-smart-contract-hcs-interoperable.md
Nana-EC: @walter-hernandez thanks for the details. 
Similar to a comment before I believe you'll want to expand this HIP with the details (descriptions and diagrams) of the precompile contracts and the network to highlight capability.
Especially as it comes to the scheduled component. 
One thing to note is, the network does not store topic messages in state, only the topic information highlighted in [Consensus Topic Info](https://github.com/hashgraph/hedera-protobufs/blob/main/services/consensus_topic_info.proto).

The message queueing service probably needs its own separate discussion.

hip-0000-smart-contract-hcs-interoperable.md
Nana-EC: Just the large caveat that Hedera won't store private keys.
Wallets would be the component for this.

hip-0000-smart-contract-hcs-interoperable.md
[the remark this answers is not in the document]
Nana-EC: @walter-hernandez given the complexity of what you're suggesting I recommend adding a sequence diagram or another appropriate architectural diagram showcasing your suggested flow.

I have a concept of what you're suggesting for the Oracle Smart Contract but don't want to assume as there seems to be a mixture of responsibilities between the a devs smart contract, the precompile smart contract on the EVM and the Hedera network.
Note that anything that impacts state on the EVM or network comes with performance concerns and higher gas costs (which is something you'll want to explore).

The message queue suggestion is an interesting one that comes up. However, I think that would be outside of this HIP and would likely require a HIP on its own if you're suggesting it be implemented at the network services level.

hip-0000-smart-contract-hcs-interoperable.md
gregscullard: In the case of the Token Service, Token compliance keys (supply, etc...) can be replaced with the ID of a smart contract, essentially making the smart contract "the key". If a smart contract is the supply key, it can mint and burn tokens, no private key is stored on Hedera.

walter-hernandez: @gregscullard wonderful! That is the right approach them to handle the keys part to engage with HCS

hip-0000-smart-contract-hcs-interoperable.md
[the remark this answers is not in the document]
walter-hernandez: @gregscullard HCS as a data feed and the consumption of its feed would be as a message queue system. Now, you raise a good consideration in the design about the frequency of messages and their consumption by HSCS.

In the case of a feed for an exchange rate, I guess the smart contract service could consume only the latest message in the queue and the oldest messages in the queue (HCS) should be removed. 

@gregscullard In general, how feasible do you think it is to implement some interoperability between HCS and smart contracts service? What do you think is pending in this PR for it to be closed and development to start?

hip-0000-smart-contract-hcs-interoperable.md
```diff
@@ -0,0 +1,58 @@
+---
+hip: 478
+title: Interoperability Between Smart Contracts and the HCS
+author: John Conway (@scalemaildev), Walter Hernandez (@walter-hernandez)
+type: Standards Track
+category: Core
```
mgarbs, line 6:
```suggestion
category: Application
```

hip-0000-smart-contract-hcs-interoperable.md
```diff
@@ -0,0 +1,58 @@
...
+needs-council-approval: Yes
```
mgarbs, line 7:
```suggestion
needs-council-approval: No
```

hip-0000-smart-contract-hcs-interoperable.md
gregscullard: HCS is not strictly like a queue management system such as Kafka, it doesn't inherently know who is subscribing to it and doesn't know if any subscriber did get the message, it's up to the subscriber to keep track of the last message timestamp they received and query a mirror node from that timestamp onwards.

Providing an automated data feed into HSCS directly from HCS bring up issues of the TPS differential between HCS and Smart Contracts (10k vs 350k).

You could consider that a specific topic could be throttled to match the HSCS speed, but that would entail Hedera knowing which topics are subscribed to by smart contracts, knowing that a message on a topic will result in x gas consumption, etc... which would not be trivial (I suspect impossible).

Added to which, executing a smart contract requires gas which isn't specified in the topic message submission.

The best model as far as I can tell is for an external oracle to subscribe to HCS messages and deal with making the contract calls.

mgarbs (approved):

PR #499, opened 2022-05-31
Remove duplicate hip-475 in root

mgarbs:
Signed-off-by: Michael Garber <michael.garber@hedera.com>

We had hip-475 in the root and the /HIP folders, so it showed up twice at hips.hedera.com. This PR removes the one in the root folder


sergmetelin (approved):
LGTM

PR #511, opened 2022-06-06
Update staking HIP with more implementation details

tinker-michaelj:
**Description**:
Updates pseudocode with more implementation details, including:
 - Rounding of staked amounts to whole hbars.
 - Use of `rewardedSinceLastStakingChange` and `stakeAtStartOfLastRewardedPeriod` fields to ensure `pendingRewards` is always exact.
 - Use of node-specific reward rates (including down-scaled rates for nodes with stake greater than `maxStake`).


lbaird (approved):
LGTM

PR #520, opened 2022-07-07
Update hip-1: outline building out features after TechCom approval

mgarbs:
This PR informs HIP authors that they can add details to the HIP after it has been either Approved or Conditionally Approved by the techcom, without going through re-approvals if the only changes to the HIP are really about adding more details or polishing off the APIs, etc.


sergmetelin (approved):
LGTM

PR #521, opened 2022-07-07
Update hip-1 to add new status called `Stagnant`

mgarbs:
This PR introduces a new status called `Stagnant` and header property `implementer` and outlines what to do with hips that are gathering dust. If a hip doesn't get enough community support and an implementer who is willing to do the work then after four weeks the hip gets a `Stagnant` status. After 6 months in `Stagnant` status, hips are rejected


sergmetelin (approved):
LGTM

PR #540, opened 2022-08-05
Create hip-0000-token-update-remove-keys.md

justynspooner:
This HIP proposes a modification to the `TokenUpdateTransaction` feature to allow the admin key to sign an update to remove itself and/or any other key (Wipe, KYC, Freeze, Pause, Supply, Fee Schedule) from a Token.


hip-0000-token-update-remove-keys.md
```diff
@@ -0,0 +1,127 @@
...
+discussions-to: <a URL pointing to the official discussion thread>
+updated: <comma separated list of dates>
+requires: <HIP number(s)>
+replaces: <HIP number(s)>
+superseded-by: <HIP number(s)>
+---
+
+## Abstract
+
+This proposal will result in a modification to the `TokenUpdateTransaction` feature to allow the admin key to sign an update to remove itself and/or any other key (Wipe, KYC, Freeze, Pause, Supply, Fee Schedule) from a Token.
```
mgarbs, line 20: >allow the admin key to sign an update to remove itself and/or any other key from a Token.

To me, this sounds like only the admin key can do this. It would be good if any key that is supplied may remove itself as well.

mgarbs (commented):
This file needs to be in the `/HIP` path of this repo; otherwise, the validation check fails. So, when you get a chance, update this path and let the validation script run against this draft.

mattsmithies:
For cases that an NFT with no admin keys but access to other privileged keys nullable options should still be possible, perhaps based on the owner of the collection?

hip-0000-token-update-remove-keys.md
justynspooner: That's a good point. I'll get that updated. Thanks for the feedback. 👍

justynspooner:
Thanks for the comments. I have created a discussion doc here to make it easier to process: https://docs.google.com/document/d/1P-qeYaXOcPIKCkTvfqFSbuVWfgrwvf0pwNMbGfurNjI/edit?usp=sharing

bugbytesinc:
> For cases that an NFT with no admin keys but access to other privileged keys nullable options should still be possible, perhaps based on the owner of the collection?

I disagree.  If the plan is to remove mint keys in the future, then the admin owner can remove the mint key then the admin key (or both at the same time) and accomplish the same thing.  IIRC, once a key is removed, it can not be added, so an admin key can only facilitate rotating existing keys (or deleting them) and/or change memo symbol etc.  So even with an admin key, consumers still typically base their "trust" on which keys already exist.  (because you can't "add" any).

HIP/hip-0000-token-update-remove-keys.md
```diff
@@ -0,0 +1,176 @@
...
+type: Standards Track
+category: Service
+needs-council-approval: Yes
+status: Draft
+created: 2022-08-05
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/545
+updated: 2022-10-05, 2022-10-12
+requires:
+replaces:
+superseded-by:
```
mgarbs, line 15: This hip shouldn't need these as it doesn't rely on another hip or replace any existing one.

mgarbs:
> Thanks for the comments. I have created a discussion doc here to make it easier to process: https://docs.google.com/document/d/1P-qeYaXOcPIKCkTvfqFSbuVWfgrwvf0pwNMbGfurNjI/edit?usp=sharing

Could you please migrate this discussion to GitHub when you get a chance? https://github.com/hashgraph/hedera-improvement-proposal/discussions/new

HIP/hip-0000-token-update-remove-keys.md
```diff
@@ -0,0 +1,173 @@
+---
+hip: 0000
+title: Remove Existing Keys From A Token
+author: Justyn Spooner <justyn@dovu.io>
+working-group: Justyn Spooner <justyn@dovu.io>
+type: Standards Track
+category: Application
+needs-council-approval: Yes
```
mgarbs, line 8:
```suggestion
needs-council-approval: No
```

HIP/hip-0000-token-update-remove-keys.md
```diff
@@ -0,0 +1,183 @@
+---
+hip: 0000
+title: Remove Existing Keys From A Token
+author: Justyn Spooner <justyn@dovu.io>
+working-group: Justyn Spooner (@justynjj), Cooper Kunz (@cooper_kunz), Jason Fabritz (@bugbytesinc)
+type: Standards Track
+category: Application
```
mgarbs, line 7:
```suggestion
type: Informational
```

HIP/hip-0000-token-update-remove-keys.md
```diff
@@ -0,0 +1,197 @@
+---
+hip: 0000
+title: Remove Existing Keys From A Token Or Contract
+author: Justyn Spooner <justyn@dovu.io>
+working-group: Justyn Spooner (@justynjj), Cooper Kunz (@cooper_kunz), Jason Fabritz (@bugbytesinc)
+type: Informational
+category: Application
```
mgarbs, line 7:
```suggestion
```

Cooper-Kunz:
Any reason this is taking precedent over the HIP discussed in https://github.com/hashgraph/hedera-improvement-proposal/discussions/522?

HIP/hip-0000-token-update-remove-keys.md
Cooper-Kunz: Shouldn't this be a Standard - Service change?

HIP/hip-0000-token-update-remove-keys.md
Cooper-Kunz: Why doesn't this need council approval? 

Otherwise who is deciding to add the feature, which is a change to the way the network works?

HIP/hip-0000-token-update-remove-keys.md
mgarbs: Adding from the discussion you created for this HIP:
The intent here with my Informational suggestion was to have a standard with a "trash key" that SDKs could adopt. I was trying to steer it in this direction because the actual key removal may have unintended consequences. Take, for instance, a KYC key scenario. A contract could be using KYC and a disgruntled employee who used to have access to that key could remove it and mess up the original contract. There may be legal issues with this case, but I'm not a lawyer.

There was a further conversation that updated keys may need to sign the tx, so you don't ever really fix the problem with a trash key. In any case, I would like to note this in this discussion.

mgarbs:
> Any reason this is taking precedent over the HIP discussed in #522?

No reason other than I saw it as a PR, and I'm a maintainer of the PRs. Looking at your discussion about this HIP, I'd say add yourself as an author in the HIP, and we can also update the discussion to include yours too. Does this sound reasonable @Cooper-Kunz @justynspooner ?

HIP/hip-0000-token-update-remove-keys.md
```diff
@@ -0,0 +1,196 @@
...
+**Key.None**
+We can’t use `null` as the value for these fields because behind the scenes protobuf removes them for optimisation. An empty KeyList could be used however, since this action can't be undone, we propose having a dedicated constant `Key.None`. The constant provides clearer intent which is crucial when dealing with a transaction that can't be reversed.
+
+**Requirements**
+
+- If an empty KeyList is passed in, the SDK should throw an error to avoid accidental removal of keys.
+- The `TokenUpdateTransaction` should be able to remove any/all keys from a token if the key exists on the token.
+- If the Admin key is removed as part of the update transaction, then all other updates should happen first and the admin key removed last to avoid any `TOKEN_IS_IMMUTABLE` errors.
+- If a key doesn't exist on the token and a call is made to remove it then return a `TOKEN_HAS_NO_SUPPLY_KEY`, `TOKEN_HAS_NO_PAUSE_KEY` etc. This error response is the same as when trying to update a key that doesn't exist.
+- All transactions to remove a key must be signed with the admin key so if a token has no admin key, then any other keys present can't be removed.
```
bugbytesinc, line 126: All transactions should be signed with all keys involved in the change, including those being removed and those being added _in addition to_ the ADMIN key.  This would be consistent with pre-existing HAPI behavior implemented elsewhere in the system.

bugbytesinc (changes requested):

HIP/hip-0000-token-update-remove-keys.md
```diff
@@ -0,0 +1,196 @@
+---
+hip: 0000
+title: Remove Existing Keys From A Token Or Contract
+author: Justyn Spooner <justyn@dovu.io>, Cooper Kunz (@cooper_kunz), Jason Fabritz (@bugbytesinc)
+working-group: Justyn Spooner (@justynjj), Cooper Kunz (@cooper_kunz), Jason Fabritz (@bugbytesinc)
+type: Informational
```
sergmetelin, line 6: I don't think it's an Informational HIP, but rather Services HIP.

HIP/hip-0000-token-update-remove-keys.md
```diff
@@ -0,0 +1,196 @@
...
+
+// Sign the transaction with the admin key
+const signPauseTx = await transaction.sign(adminKey);
+```
+
+## Open Issues
+
+Greg Scullard raises an important point on why an early decision was made to prevent the removal of keys. This should be discussed as part of this HIP review so everyone is clear how removing certain keys could affect the token.
+
+> Generally speaking, whenever a key is set on an entity, it cannot be removed (so the same applies to supply, freeze, etc... keys). This is an early design decision which makes sense for some keys, if you had a freeze key and frozen accounts, unsetting the key would mean these accounts would be frozen for ever, unless a check is made which could be costly in terms of performance… - Greg Scullard [Discord](https://discord.com/channels/373889138199494658/616725732650909710/935199340555800616)
```
rbair23, line 188: Thank you for including this open issue in the HIP. I think we should add to the HIP a list of each type of key, and what the expected behavior is if the key is removed from each permutation of state that affects the key. The edge cases are the thing that concerns me most. 

I have a token. The token was frozen with the freeze key. The token was going to be frozen for only a short time until burned, but before it was burned the admin removed the burn and freeze keys. Now, the token is stuck in my account and I can **never** get rid of it. I will be stuck with this token forever.

We need to work through all these edge cases.

HIP/hip-0000-token-update-remove-keys.md
```diff
@@ -0,0 +1,196 @@
...
+  feeScheduleKey: Key.None,
+  adminKey: Key.None,
+}).freezeWithClient(client);
+
+// Sign the transaction with the admin key
+const signTx = await transaction.sign(adminKey);
+```
+
+**Key.None**
+We can’t use `null` as the value for these fields because behind the scenes protobuf removes them for optimisation. An empty KeyList could be used however, since this action can't be undone, we propose having a dedicated constant `Key.None`. The constant provides clearer intent which is crucial when dealing with a transaction that can't be reversed.
```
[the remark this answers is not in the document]
rbair23: In discussion with some members of my team, we had another thought. The problem is this is a "forever" change. As such, having something really explicit (even more explicit than `Key.None`) would be a good thing. On the other hand, adding a new Key type is ugly because what does it mean to have a Threshold key with a keylist where one of three keys are Key.None? Probably an error?

What if, instead, we just used empty keylist as proposed, but we also add a new boolean to update transactions that is like "-f" and can be used to mean "yes, I really really really want to set a key to null, so do allow this". That would give us the safety of something like Key.None but in a (mostly) compatible way and without having to change the Key type.

rbair23 (changes requested):

hip-0000-token-update-remove-keys.md
nkavian: Adding another perspective: it would be preferable that only the admin can change or unset a sub key. Otherwise the admin key is not truly an admin but rather a manager-of-miscellaneous-fields-on-the-token-key. The admin should not have to play tug of war; it was granted the role of admin.

There's perhaps a corner case which could be of interest to discus: if the admin key is null and the kyc key is set, the kyc key can remove itself but not change itself to a new value.

Both of these support the least-privilege model.

kantorcodes:
Hello! I wanted to check and see what the status of this HIP currently is.

This is much needed functionality and would greatly empower creators.

Ashe-Oro:
> 

I agree with this. The admin key could remove and rotate another key, but not add the key if it's not currently set. This allows additional trust when the ADMIN key is set, but the SUPPLY (for example) is not set or has been removed.

HIP/hip-0000-token-update-remove-keys.md
Ashe-Oro: @michielmulders let's work thru these edge cases together. 

On a slightly different note, i'm still a fan of being able to send any token back to the treasury as a type of "burn" mechanism. That way, nobody will be stuck with a token that they don't want and cannot get rid of. Altho I think that would be best in its own HIP.

HIP/hip-0000-token-update-remove-keys.md
Ashe-Oro: coming back to this...i agree that the existing key should not be required to sign. What if the owner no longer has access to it, but still has the ADMIN key? I believe the ADMIN key is currently the only one that needs to be signed with both existing and new keys.

hip-0000-token-update-remove-keys.md
Ashe-Oro: > Hello! I wanted to check and see what the status of this HIP currently is.
> 
> This is much needed functionality and would greatly empower creators.

still in community discussion. Seems there's still an outstanding issue around:

1. Should existing keys be required to sign when being rotated? 
2. Should a key be able to rotate itself without an ADMIN key?

bugbytesinc: 1. Yes - keys being rotated out should sign, which is consistent with the current implementation of the existing Hedera Network.
2. No - rotating keys is mutating the object, to be consistent with the current implementation of the existing Hedera Network, an admin key signals mutability, and therefore needs to participate in mutating an object configuration.

nkavian: Aside from "That's how Hedera Network does it", for what other possible reason(s) are the subkeys being required to sign? Is there some technical or non-technical reason for this design?

From a security perspective, that is a flawed approach; so would like to understand the above.

bugbytesinc: Hedera got this right in the original design, except the overlooked the need to turn existing mutable objects into immutable ones, hence the need for this HIP.

nkavian: From what you're describing, it sounds like you're treating the supply key as the "supply key admin" and the trust of from the user is targeted to solely the supply admin and not the super root admin.

What you're proposing is uncommon. When compared to either the real world or to the tech domain. Other blockchains treat the admin as you say the "super root admin"; an admin doesn't need the permission from a subkey. In a real world comparison, a manager does not need an employees consent to lay them off (whereas in your assertion, the employee must consent to the manager's actions).

---

Giving a counter example to your example: if a supply key can hold the admin hostage for any reason at any point in time; then that should give the holders of that token concern. That the true power is controlled by the KYC key, the freeze key, the supply key, etc, since those keys can at any time disrupt the the token and the admin doesn't have unilateral control over them.

michielmulders: I agree with @nkavian that the admin key is designed to be the super root user that has absolute control over the project. 

Example: You assign the supply key of your NFT project to one of your employees. This employee turns out to be malicious and leaves the project. When you don't have this admin super power, the project is stuck because the admin key can't rotate the supply key anymore. You need this type of super power initially to guarantee the safety over your project. Once you are confident that everything is running smoothly, you want to have the ability to remove the admin key. 

If the admin key can't change the supply key without its consent, we don't need this HIP I believe? @bugbytesinc

michielmulders: The key that's being rotated doesn't need to sign. Here's a code example I have where I create an NFT with an admin key and supply key. Later, I change the supply key to a new random key. **Only the admin key had to sign the transaction,** the old and new supply key don't need to sign. This behavior is in line with the expected "super user" admin behavior. Let me know if this is correct? @bugbytesinc 

Here's the TX ID on testnet where you can see a different key for the supply key: https://hashscan.io/testnet/token/0.0.3934667 

Full code snippet (JS) is here: https://gist.github.com/michielmulders/7448e536734b80edff86de64d3e11e43

HIP/hip-0000-token-update-remove-keys.md
michielmulders: Agree, should be able to return to treasury at all times.

hip-0000-token-update-remove-keys.md
bugbytesinc: > Nor the old, nor the new supply key need to sign for this transaction to be valid. So the above is in line with the existing behavior of the admin key having "super power" controls.

I stand corrected regarding admin key being a super root (there are other parts of the hedera system this is not the case, this is not one of them, so changing behavior would break backwards compatibility)  However the original hip (or this one) is still necessary if you want to make an component immutable after a period of bootstrapping.  Say, like freezing a supply of an NFT collection forever.

Ashe-Oro: just noting a few of the take aways here so we can reach consensus. 

1. ADMIN key should be able to remove/rotate *any* key without requiring that key's signature. 
Example: ADMIN key can remove SUPPLY key without requiring the SUPPLY key to sign the action. 

2. Any key can remove itself and (obviously) needs to sign the action. 

3.  When rotating the ADMIN key, both the old and new keys must sign the action (this is how Hedera current works)

....are we able to reach consensus on these 3 points?

bugbytesinc: #2 is not backwards compatible.  ADMIN still should approve key swaps.  This is the model for mutability for the whole of hedera services.

Ashe-Oro: the issue I have here is that if no ADMIN key is set, then a key is immutable. There's plenty of reasons that a SUPPLY key can be set, but no ADMIN key set (minimal permissions). Should the SUPPLY key not then be able to remove itself to continue to reduce counterparty risk?

michielmulders: I agree with Ashe's comment. If we want to follow the minimal permissions route, keys should be able to rotate itself. Adding to that, by being able to rotate themselves, we are not increasing the risk profile for the token.

michielmulders: @bugbytesinc Let me know if you can agree with this? Happy to hear your further input into this matter. We are open to set up a call to discuss things and reach consensus 👐

bugbytesinc: Perhaps @lbaird could provide insight into the reasoning for the original decision keep objects immutable that don't have admin keys.  Should a token that has no ADMIN key be able to have the SUPPLY key remove itself so there is no longer a SUPPLY key at all?

Ashe-Oro:
> **2. If the admin key is removed, can lower privilege keys then remove themselves?** No, it still remains a lower privilege key that doesn't have the authority to remove itself. It has the ability to change itself to a valid or invalid key. Changing itself to an invalid key, like the all-zeros key, has the same effect as removing a key.

I'm looking for community feedback on this point. To be clear, the current proposed solution still requires an ADMIN key to be set in order for a key (ie SUPPLY) to be properly removed. If no ADMIN key is set, the only option to achieve token supply immutability would be to use the SUPPLY key to set itself to a pubic key which has no private key (0x000000...0 or 0xDEADBEEF... or 0x1111111, etc). 

The general guidance would be to only use 0x00000...0 and for the Hedera ecosystem to consider this key to equate to immutability unless the ADMIN key is set (since the SUPPLY key could again be rotated). 

What are your thoughts? Should there exist a path for a SUPPLY key properly remove itself when no ADMIN key is set or is setting the SUPPLY key to 0x000000 acceptable? 

Here's a flow chart of "testing for immutability" as I understand it from the proposed solution. 

image


@justynspooner @HGraphPunks @bugbytesinc @rocketmay @Cooper-Kunz @mattsmithies @nkavian

kantorcodes:
> > **2. If the admin key is removed, can lower privilege keys then remove themselves?** No, it still remains a lower privilege key that doesn't have the authority to remove itself. It has the ability to change itself to a valid or invalid key. Changing itself to an invalid key, like the all-zeros key, has the same effect as removing a key.
> 
> I'm looking for community feedback on this point. To be clear, the current proposed solution still requires an ADMIN key to be set in order for a key (ie SUPPLY) to be properly removed. If no ADMIN key is set, the only option to achieve token supply immutability would be to use the SUPPLY key to set itself to a pubic key which has no private key (0x000000...0 or 0xDEADBEEF... or 0x1111111, etc).
> 
> The general guidance would be to only use 0x00000...0 and for the Hedera ecosystem to consider this key to equate to immutability unless the ADMIN key is set (since the SUPPLY key could again be rotated).
> 
> What are your thoughts? Should there exist a path for a SUPPLY key properly remove itself when no ADMIN key is set or is setting the SUPPLY key to 0x000000 acceptable?
> 
> Here's a flow chart of "testing for immutability" as I understand it from the proposed solution.
> 
> ```
> //testing if an NFT Collection is immutable
> immutable = 
>   supply == null 
>   || (supply == allZeros 
>       && (admin == null 
>           || admin == allZeros))
> ```
> 
> *the use of `supply` key is only used as a reference and this logic is shared by all “lower-privilege keys, not just the `supply` key
> 
> **the use of `allZeros` is in reference to any invalid key, not only to `0x0000000000000000000000000000000000000000`
> 
> ***the allZeros key is recommended to communicate an invalid key.
> 
> image
> 
> It should be mentioned that in either case, using an invalid key (such as 0x00000) or properly removing the key will require NFT apps to test against both paths to confidently display if an NFT Collection is immutable.
> 
> @justynspooner @HGraphPunks @bugbytesinc @rocketmay @Cooper-Kunz @mattsmithies @nkavian

My two cents are that we should have a pathway for any key to remove itself, but I am happy to start with the admin key being able to do that and revisiting in another hip if that means we can expedite this along. 

We've used smart contract based mints and always have to think about this scenario. Having the ability to remove keys even with the admin key set, would open many doors in terms of flexibility.

Ashe-Oro:
> My two cents are that we should have a pathway for any key to remove itself, but I am happy to start with the admin key being able to do that and revisiting in another hip if that means we can expedite this along.
> 
> We've used smart contract based mints and always have to think about this scenario. Having the ability to remove keys even with the admin key set, would open many doors in terms of flexibility.

After posting this, I had a call with Engineering and they we discussed a solution which would includes a key being able to remove itself as long as it's currently has the highest priority. 

Said differently, there are 2 levels of priority on a token HIGH and LOW. The ADMIN key has intrinsic high privilege and all other keys have low privilege.  They key here that ONLY HIGH PRIVILEGE KEYS CAN REMOVE KEYS INCLUDING ITSELF. 

If set, the ADMIN is the sole high priv key and is the only key that can remove all other keys and itself. It can also change any other key (including itself) as described in my last comment. 

The difference here is that if the ADMIN key is REMOVED...all other keys become high priv and can remove themselves in addition to changing themselves. That said, there's no path for a SUPPLY key to either change or remove a FREEZE key, etc. 

This empowers each key to manage itself and does not reduce the intrinsic high privilege of the ADMIN key since if it's set (even to an invalid key) then they minor keys cannot remove themselves. 

One last important thing to note...if the ADMIN key is set but invalid thus unable to call the removeKey function, the other keys (SUPPLY, WIPE, etc) are able to set themselves to 0x000000...0 to establish immutability since they will never reach the high priv status. While this solution works, I recommend using the cleaner removeKey functionality if you can since it should be the first query on an NFT Collection to determine immutability or not. 

tl;dr

- Any key considered "high privilege" can remove itself. 
- Only the ADMIN key has intrinsic high privilege. 
- If the ADMIN key is removed, all other keys are considered high privilege and can remove themselves individually (see #1) 
- If a key is NOT set, it cannot be set under any circumstance. 
- Any key can change itself at any time to a valid or invalid key (ie 0x000000...0) 
- Only the ADMIN key can change any other key (ie. SUPPLY, FREEZE, WIPE, KYC, etc) to a valid or invalid key (ie 0x000000...0)

nkavian:
> I'm looking for community feedback on this point. 

My feedback would be to avoid allowing child keys to pass their permissions to someone else.

A key has a responsibility to perform its core action. A `freezeKey` should be responsible to manage freeze and unfreeze.

The key should not be responsible to manage itself. It escalates/increases the amount of management required.

Imagine a freezeKey were stolen, and the key was allowed to reassign itself to a different malicious address.
 - The original freezeKey becomes dead, the owner can't change it to a different safe key in time.
 - If the admin is asleep then 8 hours or 2 weeks pass by with malicious activities occurring.

To mitigate this and other types of issues, more off-chain tech has to be built to monitor, detect, & alert a larger set of circumstances.

For example, moving from a fee-less blockchain to a system that charged rent for wallets increased the amount of off-chain monitoring to urgently verify every single hosted wallet; to have an early warning system if a wallet was going to expire with tokens still stored in it. Then the scanning has to occur on a daily basis (does the indexer really want all this scanning logic). Then the system becomes fragile when indexers, validators, and you name it get rate limited or have an outage.

Considering the impact on integrators is just as valuable as considering how cool it would be if a key holder could change themselves.

Sometimes the simpler solution is better.

michielmulders:
> 



> > I'm looking for community feedback on this point.
> 
> My feedback would be to avoid allowing child keys to pass their permissions to someone else.
> 
> A key has a responsibility to perform its core action. A `freezeKey` should be responsible to manage freezes and unfreezes.
> 
> The key should not be responsible to admin itself. It escalates/increases the amount of management required off-chain.
> 
> Imagine a freezeKey were stolen, and the key was allowed to reassign itself to a different malicious address.
> 
> * The original freezeKey becomes dead, the owner can't change it to a different safe key in time.
> * If the admin is asleep then 8 hours or 2 weeks pass by with malicious activities occurring.
> 
> I understand it's not a perfect example, but the point was that it's more complicated to handle. To mitigate this and other types of issues, more off-chain tech has to be built to monitor, detect, & alert a larger set of circumstances.
> 
> For example, moving from a fee-less blockchain to a system that charged rent for wallets increased the amount of off-chain monitoring to urgently verify every single hosted wallet; to have an early warning system if a wallet was going to expire with tokens still stored in it. Then the scanning has to occur on a daily basis (does the indexer really want all this scanning logic). Then the system becomes fragile when you get rate limited or indexers, validators, & you name it has an outage.
> 
> Considering the impact on integrators is just as valuable as considering how cool it would be if a key holder could change themselves.
> 
> Sometimes the simpler solution is better.

@nkavian When there's no admin key and the freeze key gets compromised, there's no difference between being able to remove itself or set itself to a valid or invalid key. In both cases, the key is unretrievable. 

If there is an admin key, then only the admin key can remove the key. If the freeze key decides to change itself to a malicious or invalid key, the admin key can still reinstate access by changing it to a legit, valid key. 

But if I understand you correctly, you don't want to give any administrative responsibilities for lower-privilege keys. In case the freeze key gets compromised, the admin key can change the freeze key to a valid key again.

bugbytesinc:
I wish this was two or more separate HIPS.  One explicitly for extending **_existing_** functionality (where all lower-level key rotations already require consent of ADMIN) to include the ability to remove a key to that list of functionality.

...and One for the discussion of whether or not lower-level keys can self-admin.

...and/or One for the discussion of self-admin when ADMIN does not exist and the object should otherwise be immutable.

I am in the camp that once the ADMIN key is removed, the object becomes immutable, just like present functionality.

It seems we are getting hung up the semantics of the last two and its stalling the whole thing.

nkavian:
@michielmulders I did try to point out that my example is not perfect; and I understand many people can find flaws in it. It was late.

However, my main point was about considering the effects on integrators. Hedera has already made other design choices that increased our monitoring layer and raised integration/maintenance complexity. I'm contrasting that with knowledge of 16 other blockchains we've integrated and 5 others we've evaluated for readiness. If child keys have semi-admin privileges, it's simply increasing that problem space.

Developer sentiment is painfully important and can affect organic growth; when they whisper a different blockchain/graph is easier to use.

> you don't want to give any administrative responsibilities for lower-privilege keys

Yes, that's my point of view.

Ashe-Oro:
> integrators

agree that this could be 2 separate HIPs.  It's a balancing act between (more scope & fewer) vs (less scope & more frequent).  I'll bring it up with Engineering and see which they prefer.

nkavian:
> NFT Risk Calculator: https://github.com/hashgraph/hedera-nft-utilities/tree/main/risk

Are there any 3rd party impartial risk scoring tools?

That's a great counter example you shared, thanks.

Ashe-Oro:
> Are there any 3rd party impartial risk scoring tools?

I'm not aware of any other risk score calculators in the Hedera ecosystem.

Ashe-Oro:
> > NFT Risk Calculator: https://github.com/hashgraph/hedera-nft-utilities/tree/main/risk
> 
> Are there any 3rd party impartial risk scoring tools?
> 
> That's a great counter example you shared, thanks.

also if you disagree or have suggestions to improve the risk score manager, please ping me ashe.oro@swirldslabs.com.

Ashe-Oro:
UPDATE: After giving time for community feedback on this HIP and thru several conversations with the SwirldsLabs Engineering team, I believe it's best to *NOT* include the ability for keys to remove themselves at this time. I agree with @bugbytesinc that this functionality can (and should) be broken out into it's own HIP. 

It is important to note that all NFTs which have an ADMIN key set are considered to carry HIGH risk since the entire NFT Collection can be deleted with the ADMIN key. 

This means that if you want to properly remove a key in the future, you'll need to set an ADMIN key and your NFT Collection will be considered HIGH risk until it's removed. 

If you choose NOT to include an ADMIN key, you have the option to use the lower priv key (ie SUPPLY) to set itself to 0x00000000..0 which essential the same as removing they key since there's no private key associated with the "long-zero" public key. 

With the 0x0000 key set as a low priv key AND no ADMIN key set, the NFT Collection can be seen as immutable. Again, here's the path to check for immutability. 

image

tl;dr
1. ADMIN key can properly remove itself
2. ADMIN key can update itself to either a valid or invalid key
3. ADMIN key can properly remove low priv key (ie SUPPLY, WIPE, KYC, etc)
4. Low priv key can update itself to either a valid or invalid key
5. If a key has been removed, it CANNOT be set/defined in the future. 
6. Low priv key CANNOT properly remove itself; only update itself. 

Any questions?

mgarbs (approved):

mgarbs (approved):

mgarbs (approved):

PR #550, opened 2022-08-24
Add instructions for ruby bundler in readme for local build

mgarbs:
Our readme is missing the install step of the local build instruction so this pr adds that.


sergmetelin (approved):
LGTM

PR #565, opened 2022-09-06
Define more granular NFT standard using HIP412

michielmulders:
**Description**:
This HIP tries to define a strict specification for "Web3 NFTs" (also referred to as "Collectible NFTs") using HIP412 as the base layer. The goal of this HIP is to make it easier for NFT tooling to interpret and parse NFT token metadata.

### DIFF
--- a/HIP/hip-0000-nft-standard.md
+++ b/HIP/hip-0000-nft-standard.md
@@ -152,7 +152,7 @@ Alternatively, you can use [Arweave](https://www.arweave.org/), receiving a simi
 
 **Optional (recommended)**
 
-**Description:** SHA-256 digest of the file pointed by the `image` property. The `image_integrity` property that contains a cryptographic hash of the representation of the resource the author expects to load. 
+**Description:** SHA-256 digest of the file pointed by the `image` property. The `sha256_checksum` property that contains a cryptographic hash of the representation of the resource the author expects to load. 
 
 For instance, an author may wish to load some image from a shared server. Specifying that the expected SHA-256 hash of https://example.com/image.jpeg is ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad means that the user agent can verify that the data it loads from that URL matches that expected hash before loading the NFT. This integrity verification significantly reduces the risk that an attacker can substitute malicious content.
 
--- a/HIP/hip-0000-nft-standard.md
+++ b/HIP/hip-0000-nft-standard.md
@@ -83,26 +83,32 @@ Below is the human-readable schema, presented to maximize clarity.
 }
 ```
 
-### Field Specific Rationale
+Below is the field-specific rationale.
 
-#### name
+### name
 
-**Type:** string 
+**Type:** string
+
 **Required**
+
 **Description:** Specific NFT name
 
 
-#### creator
+### creator
 
-**Type:** string 
+**Type:** string
+
 **Required**
+
 **Description:** Artist name
 
 
-#### creatorDID
+### creatorDID
 
 **Type:** string (URI)
+
 **Optional**
+
 **Description:** 
 
 This is an optional field to carry a decentralized identifier for the creator. This would allow a creator to subsequently lay claim to a creation even if the current ownership is different.
@@ -116,17 +122,21 @@ https://github.com/hashgraph/did-method/blob/master/did-method-specification.md
 https://w3c.github.io/did-core/
 
 
-#### description
+### description
 
-**Type:** string 
+**Type:** string
+
 **Required**
+
 **Description:** Human-readable description of the NFT.
 
 
-#### image
+### image
 
 **Type:** string (CID or URI)
+
 **Required**
+
 **Description:** CID or path to the NFT's **preview image** file. Each NFT should have a preview image set to allow NFT tooling to load a display image quickly. 
 
 For image-based NFTs, you can use a compressed image optimized for the web. For other types of NFTs, you can use any compressed image that gives a visual clue as to what to expect as a user. 
@@ -136,10 +146,12 @@ It's recommended to host your file on [IPFS](https://ipfs.io/) and use a service
 Alternatively, you can use [Arweave](https://www.arweave.org/), receiving a similar CID that looks like this: `ar://<hash>`.
 
 
-#### image_integrity
+### image_integrity
 
 **Type:** string (SHA-256)
+
 **Optional (recommended)**
+
 **Description:** SHA-256 digest of the file pointed by the `image` property. The `image_integrity` property that contains a cryptographic hash of the representation of the resource the author expects to load. 
 
 For instance, an author may wish to load some image from a shared server. Specifying that the expected SHA-256 hash of https://example.com/image.jpeg is ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad means that the user agent can verify that the data it loads from that URL matches that expected hash before loading the NFT. This integrity verification significantly reduces the risk that an attacker can substitute malicious content.
@@ -147,10 +159,12 @@ For instance, an author may wish to load some image from a shared server. Specif
 References: https://w3c.github.io/webappsec-subresource-integrity/
 
 
-#### type
+### type
 
 **Type:** string (MIME type)
+
 **Required**
+
 **Description:** Mime formatting shall follow the following format: **type/subtype**. As a rule, mime types are all lowercase. However apps should be programmed to accept any case for robustness.
 
 A list of common mime types can be found here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
@@ -158,11 +172,14 @@ A list of common mime types can be found here: https://developer.mozilla.org/en-
 Note that mime types for directories are not uniformly defined. Some IPFS CIDs point to directories rather than files, so this type is useful. This standard shall define it using the format: text/directory.
 
 
-#### format
+### format
 
 **Type:** string
+
 **Required**
+
 **Possible values:** 
+
 - `none`: When not following any standard (not recommneded)
 - `opensea`: When following the [OpenSea standard](https://docs.opensea.io/docs/metadata-standards)
 - `HIPXXX`: When following a standard defined by HIPs, e.g. `HIP412`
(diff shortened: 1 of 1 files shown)


HIP/hip-0000-nft-standard.md
```diff
@@ -0,0 +1,527 @@
...
+Below is the human-readable schema, presented to maximize clarity. 
+
+```json
+{
+    "name": "NFT name (required)",
+    "creator": "artist (required)",
+    "creatorDID": "DID URI (optional)",
+    "description": "human readable description of the asset (recommended)",
+    "image": "cid or path to the NFT's preview image file (required)",
+    "image_integrity": "SHA-256 digest of the file pointed by the image field (recommended)",
```
justynspooner, line 49: Suggest keeping this the same name as we use in the files array
```suggestion
    "sha256_checksum": "SHA-256 digest of the file pointed by the image field (recommended)",
```

HIP/hip-0000-nft-standard.md
```diff
@@ -0,0 +1,527 @@
...
+                "is_default_file": "(Type: boolean) indicates if the file is the main file for this NFT (conditionally optional - required when multiple files are listed)",
+                "metadata": "metadata object (optional)",
+                "metadata_uri": "uri to metadata (optional)",
+                "localization": [
+                    // optional localization array
+                    {
+                        "uri": "uri to file (required)",
+                        "locale": "language identifier (required)"
+                    }
+                ]
```
justynspooner, line 68: localization should appear at the root of the metadata as well so that the main name, description etc can be localized.

HIP/hip-0000-nft-standard.md
```diff
@@ -0,0 +1,527 @@
...
+            "trait_type": "name of trait (required)",
+            "value": "value for this trait (required)",
+            "max_value": "maximum possible value for this trait (optional)"
+        }
+    ]
+    // no additional root-level properties allowed - properties array should contain additional properties you may require
+}
+```
+
+Below is the field-specific rationale.
```
justynspooner, line 86: It would be useful to include the JSON Schema representation like we did in HIP 10 and 412 so that validators and parsers can make use of it.

HIP/hip-0000-nft-standard.md
```diff
@@ -0,0 +1,527 @@
...
+**Description:** CID or path to the NFT's **preview image** file. Each NFT should have a preview image set to allow NFT tooling to load a display image quickly. 
+
+For image-based NFTs, you can use a compressed image optimized for the web. For other types of NFTs, you can use any compressed image that gives a visual clue as to what to expect as a user. 
+
+It's recommended to host your file on [IPFS](https://ipfs.io/) and use a service like [Pinata](https://pinata.cloud/) to easily pin your file. Your CID should look like this: `ipfs://<hash>`.
+
+Alternatively, you can use [Arweave](https://www.arweave.org/), receiving a similar CID that looks like this: `ar://<hash>`.
+
+
+### image_integrity
```
justynspooner, line 149:
```suggestion
### sha256_checksum
```

HIP/hip-0000-nft-standard.md
```diff
@@ -0,0 +1,527 @@
...
+
+**Type:** string (SHA-256)
+
+**Optional (recommended)**
+
+**Description:** SHA-256 digest of the file pointed by the `image` property. The `image_integrity` property that contains a cryptographic hash of the representation of the resource the author expects to load.
```
justynspooner, line 155:
```suggestion
**Description:** SHA-256 digest of the file pointed by the `image` property. The `sha256_checksum` property that contains a cryptographic hash of the representation of the resource the author expects to load. 
```

HIP/hip-0000-nft-standard.md
```diff
@@ -0,0 +1,527 @@
...
+
+**Required**
+
+**Possible values:** 
+
+- `none`: When not following any standard (not recommneded)
+- `opensea`: When following the [OpenSea standard](https://docs.opensea.io/docs/metadata-standards)
+- `HIPXXX`: When following a standard defined by HIPs, e.g. `HIP412`
+
+**Description:** Describes the used standard to display NFT metadata. This property allows for faster parsing by NFT tooling because they don't have to figure out which standard they are dealing with.
```
justynspooner, line 187: I'd like to propose that we update this to have a reference to a published schema. This would let parsers very quickly identify not only the schema type but also the version of the schema being used over time as the standards evolve. I've added more examples in the main discussion here: https://github.com/hashgraph/hedera-improvement-proposal/discussions/413#discussioncomment-3589586

HIP/hip-0000-nft-standard.md
michielmulders: I wanted to make it a bit more specific for the image property but I'm happy to accept this name as well for clarity. The ARC0003 standard defined by Algorand used different naming for the preview image and additional files' checksums.

HIP/hip-0000-nft-standard.md
michielmulders: Right, didn't think of that!

HIP/hip-0000-nft-standard.md
michielmulders: @justynspooner I've added a JSON schema 👍

HIP/hip-0000-nft-standard.md
```diff
@@ -0,0 +1,795 @@
...
+Below is the human-readable schema, presented to maximize clarity. This document also includes a Formal JSON Schema definition to assist in data validation and describe the data in formal terms. **[See the Formal JSON Schema Definition](#formal-json-schema-definition) section below.**
+
+```json
+{
+	"name": "NFT name (required)",
+	"creator": "artist (required)",
+	"creatorDID": "DID URI (optional)",
+	"description": "human readable description of the asset (required)",
+	"image": "cid or path to the NFT's preview image file (required)",
+	"sha256_checksum": "SHA-256 digest of the file pointed by the image field (recommended)",
```
justynspooner, line 49: Is there a reason we use sha256 over md5? md5 is faster to generate from a file and since we're not using it for storing a password and it's only used as a checksum it should do the job fine. sha256 might be overkill for this.

HIP/hip-0000-nft-standard.md
```diff
@@ -0,0 +1,795 @@
...
+**Optional**
+
+**Description:** External URI pointing to an informational page about a collection or specific NFT information i.e. `www.mynft/collection/<serial>/`.
+
+
+### files
+
+**Type:** array
+
+**Required**
```
justynspooner, line 215: I don't think this should be required. If the NFT preview image is enough to represent the NFT and there is no need for any higher res imagery, then we shouldn't force them to add any other files here.

HIP/hip-0000-nft-standard.md
```diff
@@ -0,0 +1,795 @@
...
+
+
+### attributes.display_type
+
+**Type:** string - Possible display types (but other types are allowed):
+
+- `text` (**default value representation**)
+- `percentage` (for integer or number values)
+- `boost` (for integer or number values)
+- `datetime` (for string values)
```
justynspooner, line 313: Opensea have a good example of the `date` display_type. We can extend this to support `datetime` as well to include the time. I've added color as well to give some inspiration to how that could work.

https://docs.opensea.io/docs/metadata-standards#date-traits

```suggestion
- `datetime` (for a number which represents the unix timestamp in seconds )
- `date` (for a number which represents the unix timestamp in seconds )
- `color` (for a hexadecimal or rgb string color sequence such as "#00ff44" or "rgb(0,255,0)" )
```

HIP/hip-0000-nft-standard.md
```diff
@@ -0,0 +1,795 @@
...
+				}
+			},
+			"additionalProperties": true
+		},
+		"files": {
+			"type": "array",
+			"contains": {
+				"type": "object"
+			},
+			"minContains": 1,
```
justynspooner, line 660: As mentioned above, I don't think we need to force a files array if the image is enough for the NFT

HIP/hip-0000-nft-standard.md
```diff
@@ -0,0 +1,795 @@
...
+					"display_type": {
+						"type": "string",
+						"description": "Sets the representation of the value of the trait."
+					},
+					"value": {
+						"type": ["string", "integer", "number", "boolean"],
+						"description": "Value for trait."
+					},
+					"max_value": {
+						"type": "string",
```
justynspooner, line 716: This could be a numerical max too
```suggestion
						"type": ["string", "integer", "number"],
```

HIP/hip-0000-nft-standard.md
```diff
@@ -0,0 +1,795 @@
...
+	},
+	"required": [
+		"name",
+		"creator",
+		"image",
+		"type",
+		"format",
+		"files",
+		"properties",
+		"attributes"
```
justynspooner, line 761: I think these would be the minimum required properties for a very simple NFT. I think we're OK adding image and type as required now if we're pushing for it to be the preview image. I can't remember the react exact reason we made image optional in HIP 412, feel free to chime in if anyone sees a reason why this shouldn't be required.

```suggestion
		"name",
		"image",
		"type"
```

HIP/hip-0000-nft-standard.md
```diff
@@ -0,0 +1,795 @@
+---
+hip: 0000
+title: Define a granular NFT Standard (Extends HIP-412)
+author: Michiel Mulders <@michielmulders>
+working-group: Michiel Mulders <@michielmulders>, Ashe Oro <@Ashe_Oro>, Hash Axis <@hashaxis>, Rocket May <@rocketmay>, HGP Patches <@HGP_Patches>
+type: Informational
+needs-council-approval: No
+status: Draft
```
mgarbs, line 8:
```suggestion
status: Deferred
```

HIP/hip-0000-nft-standard.md
michielmulders: I would opt to force this because otherwise we will get a mix between people who use the preview image as the main image and some who don't. If you look at other specifications, they even provide guidelines for maximum dimensions for the preview image? We wanted to be a bit more specific so I don't think this hurts?

HIP/hip-0000-nft-standard.md
michielmulders: See other comment about forcing this. They can always set it to the same image if they really want to have the main image be the preview image. It's not a lot of extra metadata. I prefer forcing this.

HIP/hip-0000-nft-standard.md
michielmulders: As this will become a reference implementation under HIP412, this should be specific. But you are right about setting `properties` and `attributes` to be optional. The rest should be required.

rocketmay: “image” was originally listed as ‘required’, however HashAxis noted that in the case of file NFTs, creating both image and file would require two files to pin instead of one. At the scale of millions, this doubling of the number of pinned images is significantly impactful to the cost of hosting the files. For files that can have their thumbnails generated programmatically, an image is not required. Because of this, the “image” field was changed to Conditionally Optional.

HIP/hip-0000-nft-standard.md
michielmulders: I like this!

PR #574, opened 2022-09-09
build error array and throw if length greater than 0

mgarbs:

**Description**:
This PR builds an error array and throws if it has a length greater than 0. Its needed because people who submit PRs for their HIPs are fixing errors one by one (each time pushing a new commit and running against that validation check) With this PR, the validation check collects everything wrong with the header inside an error array and throws rather than throwing the moment it encounters a bad header.

This PR modifies `validateHIP.js` in order to collecting all errors in the header before throwing.


sergmetelin (approved):
LGTM

PR #584, opened 2022-09-27
hip-584 evm module

mustafauzunn:
Signed-off-by: Mustafa Uzun <mustafa.uzun@limechain.tech>

**Description**:

In order to be possible to both execute and simulate transactions, run by the EVM used inside `hedera-node`,
we need to extract a module encapsulating all of the EVM logic.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-584.md
+++ b/HIP/hip-584.md
@@ -5,9 +5,10 @@ author:  Ivan Kavaldzhiev <ivan.kavaldzhiev@limechain.tech>, Mustafa Uzun <musta
 type: Standards Track
 category: Service
 needs-council-approval: Yes
-status: Draft
+status: Review
 created: 2022-09-27
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/586
+updated: 2022-10-04
 ---
 
 ## Abstract
--- a/HIP/hip-584.md
+++ b/HIP/hip-584.md
@@ -1,0 +1,149 @@
+---
+hip: 584
+title: EVM Module
+author:  Ivan Kavaldzhiev <ivan.kavaldzhiev@limechain.tech>, Mustafa Uzun <mustafa.uzun@limechain.tech>
+type: Standards Track
+category: Service
+needs-council-approval: Yes
+status: Draft
+last-call-date-time:
+created: 2022-09-27
+discussions-to:
+---
+
+## Abstract
+
+Encapsulating Hedera smart contract EVM specific logic in a library.
+It would expose interfaces that will allow different implementations.
+Mirror node would build ethereum-related REST APIs on top of the library. 
+Mirror nodes could use the library to simulate a transaction and the output would be used to 
+populate the return value for the REST API that is invoked. 
+Consensus nodes would also use the library and keep their existing logic by providing specific interface implementations.
+
+## Motivation
+
+Extracting common Hedera smart contract EVM logic in a single place
+
+Enabling free read-only queries.
+
+Enabling free gas estimate for a given transaction.
+
+Enabling free transaction simulations.
+
+## Rationale
+
+In order to be possible to both execute and simulate transactions, run by the EVM used inside `hedera-node`,
+we need to extract a module encapsulating all of the EVM logic.
+
+The resulting library would allow us to fulfil the following use cases:
+
+- Execute transactions in a regular flow with committing changes to state
+- Execute transaction simulations that perform read-only operations
+- Execute transaction simulations that perform speculative writes without committing changes to state
+- Execute transaction simulations that perform speculative writes using historical data for transaction replay, 
+without committing changes to state
+
+## User stories
+
+- As a developer, I want to have a library with common Hedera smart contract EVM logic that I can reuse in different projects
+- As a user, I want to perform `eth_call` calls from the Ethereum JSON-RPC standard
+- As a user, I want to perform `eth_gasEstimate` calls from the Ethereum JSON-RPC standard
+
+## Specification
+
+### EVM Library Definition
+
+We want the EVM library to be used by multiple clients like the `hedera-services` and `hedera-mirror-node` repos
+and to be able to be extended for a future use from additional clients.
+
+To facilitate differences, we need to define interfaces, which would be inside the library 
+and each client of the library (in this case `hedera-node` and `mirror-node`) would provide its own implementation.
+
+### Mirror Node
+
+Information will be made available via new REST APIs
+
+#### Read-only queries REST API
+
+A new  `/api/v1/contracts/call` REST API will be added to execute read-only queries.
+
+The following JSON represents a typical request body:
+```json
+{
+  "block": "latest", 
+  "data": "0x81a73ad500000000000000000000000000000000000000000000000000000000000004e5", 
+  "from": "0x00000000000000000000000000000000000004e2", 
+  "gas": "0x76c0", 
+  "gas_price": "0x76c0", 
+  "to": "0x00000000000000000000000000000000000004e4", 
+  "value": "0x0"
+}
+```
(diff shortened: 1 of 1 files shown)


steven-sheehy (approved):
LGTM

HIP/hip-584.md
```diff
@@ -0,0 +1,127 @@
...
+- As a user, I want to perform `eth_gasEstimate` calls from the Ethereum JSON-RPC standard
+
+## Specification
+
+### EVM Library Definition
+
+We want the EVM library to be used by multiple clients like the `hedera-services` and `hedera-mirror-node` repos
+and to be able to be extended for a future use from additional clients.
+
+To facilitate differences, we need to define interfaces, which would be inside the library
```
Nana-EC, line 58: Should we not be listing the suggested major interfaces and their responsibilities here?

HIP/hip-584.md
```diff
@@ -0,0 +1,127 @@
...
+sidecar types. Otherwise, executing the endpoints would result in missing runtime bytecode for execution
+or using stale contract storage data.
+
+## Security Implications
+
+There would be some security implementations for `Mirror Nodes`. Since the invocation of `eth_call` and `eth_gasEstimate` RPC calls 
+would be free of charge and the `Mirror Nodes` don’t have throttle mechanism, some attack vectors would be possible. 
+The Nodes could be overloaded with huge amount of calls or calls invoking smart contract methods with huge gas usage, 
+both of which might take more system usage and slow down the network for other users. A rate limit mechanism will be implemented,
+so that the load put on Mirror Nodes to be balanced.
```
IvanKavaldzhiev, line 106: Mirror Nodes -> `Mirror Nodes`

HIP/hip-584.md
IvanKavaldzhiev: We may define the basic interface for interacting with the library. It will be inside `hedera-evm-api` module and `mirror-node` would use it to spin up a simulation of a transaction. The output would be `TransactionProcessingResult`. `hedera-evm-api` on its own side would consume and use the `hedera-evm` implementation module to perform the transaction. The dependency chain would look like this:
`besu` -> `hedera-evm` -> `hedera-evm-api `-> `hedera-node`/`hedera-mirror-web3`

Going further into more details about the other interfaces, IMO, might introduce more specific and unnecessary information for the context of a HIP.

HIP/hip-584.md
```diff
@@ -0,0 +1,128 @@
...
+Encapsulating Hedera smart contract EVM specific logic in a library.
+It would expose interfaces that will allow different implementations.
+Mirror node would build ethereum-related REST API on top of the library. 
+Mirror nodes could use the library to simulate a transaction and the output would be used to 
+populate the return value for the REST API that is invoked. 
+Consensus nodes would also use the library and keep their existing logic by providing specific interface implementations.
+
+## Motivation
+
+Extracting common Hedera smart contract EVM logic in a single place
```
shemnon, line 25: The motivation section should be prose not a bullet list.

```suggestion
Currently the EVM code is deeply embedded in consensus node logic.  However ongoing user experience work needs EVM execution to enable features such as cost-free execution of read-only smart contract queries, gas estimation, and transient simulation of read-write operations.

In order to maximize compatibility and reduce defects it would be best to run all these activities from the same codebase.  In order to do so a library needs to be extracted from the consensus logic and the consensus nodes and external programs such as the mirror node.
```

HIP/hip-584.md
```diff
@@ -0,0 +1,128 @@
...
+---
+
+## Abstract
+
+Encapsulating Hedera smart contract EVM specific logic in a library.
+It would expose interfaces that will allow different implementations.
+Mirror node would build ethereum-related REST API on top of the library. 
+Mirror nodes could use the library to simulate a transaction and the output would be used to 
+populate the return value for the REST API that is invoked. 
+Consensus nodes would also use the library and keep their existing logic by providing specific interface implementations.
```
shemnon, line 21:
```suggestion
This HIP describes the interfaces and modules that will allow the EVM implementation to be separated into a Java library. This will enable a number of future features such as different EVM implementations, EVM execution APIs on the mirror node, and transaction simulation and verification off of the consensus nodes.  The consensus nodes would continue to use this library.
```

HIP/hip-584.md
```diff
@@ -0,0 +1,128 @@
...
+- Execute transaction simulations that perform speculative writes without committing changes to state
+- Execute transaction simulations that perform speculative writes using historical data for transaction replay, 
+without committing changes to state
+
+## User stories
+
+- As a developer, I want to have a library with common Hedera smart contract EVM logic that I can reuse in different projects
+- As a user, I want to perform `eth_call` calls from the Ethereum JSON-RPC standard
+- As a user, I want to perform `eth_gasEstimate` calls from the Ethereum JSON-RPC standard
+
```
shemnon, line 51: I think we need a story for the debug use case, worded so it's just the enabling logic and not the full debug support

```suggestion
- As a user, I want a library that can be used to re-run historical EVM transactions accurately so that I can inspect and debug the execution of the bytecode logic.

```

HIP/hip-584.md
IvanKavaldzhiev: @shemnon, @Nana-EC  I think that saying `This HIP describes the interfaces and modules` would imply further and more detailed definitions about the implementation details for the EVM module. I think we agreed on being more abstract in terms of explanations in this HIP and if needed, link a more detailed design doc with interface definitions and examples at a later point.

HIP/hip-584.md
Nana-EC: Maybe this can be in a design doc committed into the services repo inside the `hedera-evm` repo

HIP/hip-584.md
```diff
@@ -0,0 +1,127 @@
...
+To facilitate differences, we need to define interfaces, which would be inside the library 
+and each client of the library (in this case `hedera-node` and `mirror-node`) would provide its own implementation.
+
+### Mirror Node
+
+Information will be made available via new REST API
+
+#### Read-only queries and gas estimate REST API
+
+A new  `/api/v1/contracts/call` REST API will be added to execute read-only queries and gas estimate.
```
Nana-EC, line 67: Note that this will be a `POST` since all current APIs are `GET`

Nana-EC (approved):
LG, one suggestion

HIP/hip-584.md
```diff
@@ -0,0 +1,127 @@
...
+
+There would be some security implementations for `Mirror Nodes`. Since the invocation of `eth_call` and `eth_gasEstimate` RPC calls 
+would be free of charge and the `Mirror Nodes` don’t have throttle mechanism, some attack vectors would be possible. 
+The Nodes could be overloaded with huge amount of calls or calls invoking smart contract methods with huge gas usage, 
+both of which might take more system usage and slow down the network for other users. A rate limit mechanism will be implemented,
+so that the load put on `Mirror Nodes` to be balanced.
+
+## How to Teach This
+
+Respective documentation will be added.
```
Nana-EC, line 110: Can be more specific with
- Service "EVM Modularization" Design Doc
- "Mirror EVM Archive Node" Design Doc
- Open API doc on Mirror Node
- docs.hedera.com sections for `/api/v1/contracts/call`

Nana-EC (changes requested):
Looking good, a couple more requests.

Nana-EC (approved):

HIP/hip-584.md
IvanKavaldzhiev: I believe we can link these corresponding documentations at a later point, so that they would more up-to-date with the latest implementation details for the evm modules.

mgarbs (approved):
LGTM

PR #585, opened 2022-09-27
Update name regex and push error to error array

mgarbs:
**Description**:
This PR modifies `validateHIP.js` and `regex.js` in order to support hyphens in names and not throw on a name check but add it to the error array to continue executing validation checks against the entire header.


sergmetelin (approved):
LGTM

PR #598, opened 2022-10-04
swap release column and last call column in display

mgarbs:


sergmetelin (approved):
LGTM

PR #610, opened 2022-11-01
update readme to bring attention to the HIP process

mgarbs:


pathornteng:
It seems that the test fails. Could you please recheck and rerun the test? @mgarbs

pathornteng (approved):
So this RP changes README.md and it does not need to get validated. LGTM

PR #632, opened 2022-11-28
[WIP] Hedera Account Service (HAS) System Contract

Nana-EC:
Signed-off-by: Nana Essilfie-Conduah <nana@swirldslabs.com>

**Description**:
Hedera Account Service (HAS) System Contract

In smart contract operations will provide `msg.sender` values as either Hedera addresses or virtual addresses. 
To support authorization flows that check addresses either explicitly in smart contract code the network should support in smart contract distinguishing operations that allow smart contract to distinug between the two forms.
Additionally, to support standard implicit checks the network should support system contract checks in a manner similar ECRECOVER for easy developer support.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:
Additional support for account evm equivalence features may be added to this system contract depending on EVM client feature expansion

**Checklist**


david-bakin-sl:
Selectors for methods:

```solidity
pragma solidity >=0.7.0 <=0.9.0;
contract HIP632 {
    function getVirtualAddresses(address) public returns (address[] memory) {
        return new address[](0);
    }

    function getHederaAddress(address) public view returns (address) {
        return address(uint160(0));
    }

    function isVirtualAddress(address) public view returns (bool) {
        return true;
    }

    function isAuthorized(address, bytes calldata messageHash, bytes calldata signatureBlob) public view returns (bool) {
        return true;
    }

    function isAuthorizedRaw(address, bytes calldata messageHash, bytes calldata signatureBlob) public view returns (bool) {
        return true;
    }
}

// via Remix:
//"getHederaAddress(address)": "a4e310ba",
//"getVirtualAddresses(address)": "e9482d42",
//"isAuthorized(address,bytes,bytes)": "b2526367",
//"isAuthorizedRaw(address,bytes,bytes)": "b2a31da4",
//"isVirtualAddress(address)": "d501235a"
```

se7enarianelabs:
What will happen if two different Hedera accounts have one shared virtual address?

lukelee-sl:
> What will happen if two different Hedera accounts have one shared virtual address?

I do not believe this state will be support.  This section from HIP 631 implies that a map will ensure that evmAddresses can be added as a virtual address to at most one Hedera account.
```
Add a new evmAddress → accountId map, for easy account lookups
Account creation
CryptoCreate transactions ...
CryptoTransfer transactions ...
In both transaction type cases the evmAddress → accountId map is consulted for an existence check. Should the virtual address be absent then creation may be attempted and the map should be updated with the evmAddress
```

PR #646, opened 2023-01-04
Fungible Token Metadata

mgarbs:

### DIFF
--- a/HIP/hip-646.md
+++ b/HIP/hip-646.md
@@ -2,19 +2,19 @@
 hip: 646
 title: Fungible Token Metadata Field
 author: May Chan (@rocketmay)
-working-group: Ashe Oro (@Ashe-Oro)
+working-group: Ashe Oro (@Ashe-Oro), Michael Garber (@mgarbs)
 type: Standards Track
 category: Service
 needs-council-approval: Yes
 status: Review
 created: 2023-01-04
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/627
-updated: 2023-01-04
+updated: 2023-01-04, 2023-01-05
 ---
 
 ## Abstract
 
-This HIP proposes the addition of the metadata field to Fungible Tokens (FT), taking after the Non-Fungible Token (NFT) metadata field which was added in HIP-17.
+This HIP proposes the addition of the metadata field to Fungible Tokens (FT), taking after the Non-Fungible Token (NFT) metadata field which was added in HIP-17. While HIP-17 metadata is scoped to a single NFT, here we propose adding metadata scoped to an entire token type.
 
 ## Motivation
 
@@ -30,7 +30,7 @@ The success of HIP-412 in standardizing the NFT space on Hedera demonstrates the
 
 As a token creator, I want a platform-standardized method to host information about my token in a decentralized fashion, allowing applications to query data such as token icon and website.
 
-As a wallet or explorer, I want to query a mirror node to obtain the metadata for any fungible token, similar to current non-fungible token calls, to simplify calls and implementation.
+As a wallet or explorer, I want to query a mirror node to obtain the metadata for any fungible token, similar to current non-fungible token calls, to simplify calls and implementation. Mirror nodes currently return metadata for individual NFTs; but for fungible token types, we want to get **token-level** metadata.
 
 As a developer, I want to store the metadata uri in a dedicated field, to avoid ambiguity of using an overloaded field such as 'memo'.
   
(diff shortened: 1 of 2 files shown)


HIP/hip-646.md
```diff
@@ -0,0 +1,90 @@
...
+needs-council-approval: Yes
+status: Review
+created: 2023-01-04
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/627
+updated: 2023-01-04
+---
+
+## Abstract
+
+This HIP proposes the addition of the metadata field to Fungible Tokens (FT), taking after the Non-Fungible Token (NFT) metadata field which was added in HIP-17.
```
tinker-michaelj, line 17: This may lead some readers to believe that HIP-17 gave metadata to non-fungible **token types**; instead of to individual NFT's. It also leaves it slightly unclear whether this HIP proposes to add `metadata` to both fungible and non-fungible token types; or only fungible types.

An extra sentence like,
```
While HIP-17 metadata is scoped to a single NFT, here we propose adding metadata scoped to an entire token type.
``` 
might help.

HIP/hip-646.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Rationale
+
+The success of HIP-412 in standardizing the NFT space on Hedera demonstrates the potential of standardized metadata functionality for FT's. Wallets, explorers and other applications can display relevant supplementary information about tokens to users in a decentralized fashion.
+
+## User stories
+
+As a token creator, I want a platform-standardized method to host information about my token in a decentralized fashion, allowing applications to query data such as token icon and website.
+
+As a wallet or explorer, I want to query a mirror node to obtain the metadata for any fungible token, similar to current non-fungible token calls, to simplify calls and implementation.
```
tinker-michaelj, line 33: Here we might also clarify that current mirror nodes return metadata for individual NFTs; but we want to also get token-level metadata (especially for fungible token types).

HIP/hip-646.md
```diff
@@ -0,0 +1,90 @@
...
+
+```
+   // Retrieves the metadata of a token
+   rpc getTokenInfo (Query) returns (Response);
+```
+
+### TokenMintTransactionBody
+Property `metadata` is updated to include all tokens.
+```
+message TokenMintTransactionBody {
```
tinker-michaelj, line 49: I think `TokenCreateTransactionBody` + `TokenUpdateTransactionBody` are more natural places to initialize and update token-level metadata.

HIP/hip-646.md
```diff
@@ -0,0 +1,90 @@
...
+    TokenID token = 1; // The token for which to mint tokens. If token does not exist, transaction results in INVALID_TOKEN_ID
+!   uint64 amount = 2; // Applicable to tokens of type FUNGIBLE_COMMON. The amount to mint to the Treasury Account. Amount must be a positive non-zero number represented in the lowest denomination of the token. The new supply must be lower than 2^63.
++   repeated bytes metadata = 3; // Applicable to **all tokens**, previously was only NON_FUNGIBLE_UNIQUE. A list of metadata that are being created. Maximum allowed size of each metadata is 100 bytes
```
tinker-michaelj, line 52: What's the motivation for a list of metadata?

tinker-michaelj (commented):
Very sensible and useful HIP.

Added some comments to avoid confusion about the implied scope of existing NFT-specific metadata.

HIP/hip-646.md
rocketmay: Agreed, this is good clarification

HIP/hip-646.md
rocketmay: +1

HIP/hip-646.md
rocketmay: Just a copy-paste error, good catch thank you

HIP/hip-646.md
tinker-michaelj: Np, I'd suggest:

1. Add a new field to [`TokenCreateTransactionBody`](https://github.com/hashgraph/hedera-protobufs/blob/main/services/token_create.proto),
```
message TokenCreateTransactionBody {
    ...
     /**
     * The token metadata, limited to a UTF-8 encoding of no more than 100 bytes;
     * will usually be a URI to a complete description (as NFT metadata generally
     * just links to a HIP-412 compliant JSON document)
     */
    string metadata = 23;
}
```

2. Permit updates by adding a new field to [`TokenUpdateTransactionBody`](https://github.com/hashgraph/hedera-protobufs/blob/main/services/token_update.proto),
```
message TokenUpdateTransactionBody {
    ...
    /**
     * If set, the token's new metadata (UTF-8 encoding max 100 bytes)
     */
    google.protobuf.StringValue metadata = 16;
}
```

3. Support getting the metadata directly from consensus nodes as you already mentioned in the [`getTokenInfo` response](https://github.com/hashgraph/hedera-protobufs/blob/main/services/token_get_info.proto),
```
message TokenInfo {
    ...
    /**
     * The token's metadata, if any.
     */
    string metadata = 27;
}
```

tinker-michaelj (approved):
LGTM! For completeness, we could disclaim it's out-of-scope to enhance the EVM system contracts for `TokenCreate`, `TokenUpdate`, and `GetTokenInfo` with metadata support at this time; but if token metadata becomes important for ecosystem interoperability, a future HIP may insist upon EVM support.

HIP/hip-646.md
mgarbs: Added this. Thanks, @tinker-michaelj

tinker-michaelj: Np, thanks @mgarbs!

pathornteng (approved):
LGTM but it's good to have someone from the engineering team do the final review from the implementation perspective.

HIP/hip-646.md
```diff
@@ -0,0 +1,142 @@
...
+message TokenUpdateTransactionBody {
+    ...
+    /**
+     * If set, the token's new metadata (UTF-8 encoding max 100 bytes)
+     */
+    google.protobuf.StringValue metadata = 16;
+}
+```
+
+3. Support getting the metadata directly from consensus nodes as you already mentioned in the [`getTokenInfo` response](https://github.com/hashgraph/hedera-protobufs/blob/main/services/token_get_info.proto),
```
tinker-michaelj, line 93: Should delete "as you already mentioned". 😁

SimiHunjan (approved):

PR #657, opened 2023-01-21
Mutable metadata fields for dynamic NFTs 

mgarbs:

### DIFF
--- a/HIP/hip-000.md
+++ b/HIP/hip-000.md
@@ -1,0 +1,115 @@
+---
+hip: 000
+title: Mutable metadata fields for dynamic NFTs
+author: Patches Punks <@HGraphPunks>
+working-group: Jason Fabritz <@bugbytesinc>, Sean Cuscus <@seancuscus>, Cooper Kunz <@Cooper-Kunz>, May Chan <@rocketmay>, Burstall Stowerling <@Burstall>, Ian Holsman <@web3-nomad>, Ashe Oro <@Ashe-Oro>, Michael Garber <@mgarbs>, Paul Sullivan <@paulgs9988>
+type: Standards Track
+category: Core
+needs-council-approval: Yes
+status: Draft
+created: 2023-01-04
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/607
+updated: 2023-01-21
+---
+
+
+## Abstract
+
+
+Currently, there is no native way to update the metadata of a non-fungible token (NFT) on the Hedera network. This HIP proposes the creation of a new key called the METADATA key, which will allow updating NFT metadata. This will provide the Hedera network with functional parity with other networks and enable developers to create dynamic NFTs with new interactions.
+
+
+## Motivation
+
+
+Dynamic NFTs are a valuable addition to the world of non-fungible tokens because they allow for more interactive and immersive experiences. With the ability to update the metadata of an NFT, developers can create game pieces that gain new equipment and stats, legal agreements that reflect changing parameters, or even NFTs that update their images based on the time of day. Dynamic NFTs also open up the possibility for NFTs to be used in a wider range of industries and applications, such as supply chain tracking, real estate, among others. The potential for dynamic NFTs is numerous, and their introduction will greatly enhance the capabilities of the NFT ecosystem.
+
+
+## Rationale
+
+
+To address this issue, the proposed solution is to create a new key type called the METADATA key. This key will allow NFTs in a collection to update their metadata.
+
+
+The use of the METADATA key to update NFT metadata will not affect the functionality of existing NFTs on the Hedera Network. It will be available for new and existing NFTs.
+
+
+## User stories
+
+
+- As a collector of NFTs, I want to be able to update the metadata of an NFT in my collection so that I can add new information or make changes to the existing metadata.
+
+
+- As a creator of NFTs, I want to be able to update the metadata of an NFT while leaving an auditable on-chain history of when the update happened and which key was used to sign for the update action so that I can retain as much transparency as possible. 
+
+
+- As a creator of NFTs, I want to use IPFS to manage my dynamic NFT’s metadata so that I don’t have to use a centralized file storage solution to enable the ability to edit.  
+
+
+- As a developer building applications on top of the Hedera Network, I want the ability to update the metadata of NFTs so that I can create dynamic NFTs and build more advanced use cases.
+
+
+## Specification
+
+
+The METADATA key will be a new key that allows for updating the metadata of an NFT in a new NftMetadataUpdate transaction.
+
+The NftMetadataUpdate transaction should accept the following parameters:
+	- identifier (string): the 0.0.<token-type>.<serial-no>
+- the new NFT metadata: the new metadata for the NFT
+
+
+The NftMetadataUpdate transaction should have a new token type field called "dynamic_metadata" of type bool, which is valid only for non-fungible tokens. By default, this field is set to false (meaning the metadata for this token type is immutable). If set to true, owners of this token type can change their NFT's metadata via the NftMetadataUpdate transaction.
+
+
+Example transaction:
+```js
+NftMetadataUpdate()
+     .setTokenId(tokenId)
+     .setMetaData(newMetaData)
+     .sign(METADATAKEY)
+     .execute(client);
+```
+
+
+Requirements: The NFT must exist on the Hedera network and should only be allowed to execute if it is signed by the new METADATAKEY. The new metadata must be a valid string. The update must be made atomically and be recorded on the Hedera network.
+
+## Backwards Compatibility
+
+
+The addition of the METADATA key will not affect the functionality of existing NFTs on the Hedera Network because it will be available for use on new and existing NFTs.
+
+## Security Implications
+
+
+The use of the METADATA key and creation of a NftMetadataUpdate transaction will not introduce any new security concerns.
+
+## How to Teach This
+TBD
+
+
+
+## Reference Implementation
+
+
+The reference implementation must be complete before any HIP is given the status of “Final”. The final implementation must include test code and documentation.
+
+
+## Rejected Ideas
+
+
+## Open Issues
+
+
+There are currently no open issues with this proposal.
+
+
+## References
+TBD
+
+
+
+## Copyright/license
+
+
+This document is licensed under the Apache License, Version 2.0 -- see [LICENSE](../LICENSE) or (https://www.apache.org/licenses/LICENSE-2.0)


HIP/hip-657.md
```diff
@@ -0,0 +1,115 @@
...
+
+
+The METADATA key will be a new key that allows for updating the metadata of an NFT in a new NftMetadataUpdate transaction.
+
+The NftMetadataUpdate transaction should accept the following parameters:
+	- identifier (string): the 0.0.<token-type>.<serial-no>
+- the new NFT metadata: the new metadata for the NFT
+
+
+The NftMetadataUpdate transaction should have a new token type field called "dynamic_metadata" of type bool, which is valid only for non-fungible tokens. By default, this field is set to false (meaning the metadata for this token type is immutable). If set to true, owners of this token type can change their NFT's metadata via the NftMetadataUpdate transaction.
```
Nana-EC, line 62: To confirm does this imply only the current owner can make the change.
If so might be good to call this out under Security section.
Explicitly note the permissions consideration that token creators and previous owners cannot make such changes

HIP/hip-657.md
```diff
@@ -0,0 +1,115 @@
...
+The reference implementation must be complete before any HIP is given the status of “Final”. The final implementation must include test code and documentation.
+
+
+## Rejected Ideas
+
+
+## Open Issues
+
+
+There are currently no open issues with this proposal.
```
Nana-EC, line 104: Q's

1. Can token creators make such an update when they do not own the NFT?
2. Can token creators set the NFT to be immutable/static in a way that future owners cannot make them dynamic?
3. Is the key type applicable at the token class of the individual NFT level?

HIP/hip-657.md
```diff
@@ -0,0 +1,115 @@
...
+## Motivation
+
+
+Dynamic NFTs are a valuable addition to the world of non-fungible tokens because they allow for more interactive and immersive experiences. With the ability to update the metadata of an NFT, developers can create game pieces that gain new equipment and stats, legal agreements that reflect changing parameters, or even NFTs that update their images based on the time of day. Dynamic NFTs also open up the possibility for NFTs to be used in a wider range of industries and applications, such as supply chain tracking, real estate, among others. The potential for dynamic NFTs is numerous, and their introduction will greatly enhance the capabilities of the NFT ecosystem.
+
+
+## Rationale
+
+
+To address this issue, the proposed solution is to create a new key type called the METADATA key. This key will allow NFTs in a collection to update their metadata.
```
Nana-EC, line 31: Is the implication a new key type [here](https://github.com/hashgraph/hedera-protobufs/blob/main/services/basic_types.proto#L576)?

HIP/hip-657.md
```diff
@@ -0,0 +1,115 @@
...
+## Abstract
+
+
+Currently, there is no native way to update the metadata of a non-fungible token (NFT) on the Hedera network. This HIP proposes the creation of a new key called the METADATA key, which will allow updating NFT metadata. This will provide the Hedera network with functional parity with other networks and enable developers to create dynamic NFTs with new interactions.
+
+
+## Motivation
+
+
+Dynamic NFTs are a valuable addition to the world of non-fungible tokens because they allow for more interactive and immersive experiences. With the ability to update the metadata of an NFT, developers can create game pieces that gain new equipment and stats, legal agreements that reflect changing parameters, or even NFTs that update their images based on the time of day. Dynamic NFTs also open up the possibility for NFTs to be used in a wider range of industries and applications, such as supply chain tracking, real estate, among others. The potential for dynamic NFTs is numerous, and their introduction will greatly enhance the capabilities of the NFT ecosystem.
```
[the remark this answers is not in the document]
Ashe-Oro: the network won't track the various states of the metadata (ie the history of the metadata values) but rather than the metadata was updated since it will be an on-chain transaction.

HIP/hip-657.md
Ashe-Oro: no, not a new key type but rather the addition of a new purposeful key on the token_id called `METADATA`. It can be of the same type(s) as the other keys on the token_id as currently defined.

HIP/hip-657.md
Ashe-Oro: as with the other keys on a token_id, the account which controls they associated pub/priv key pair is able to sign for the `NftMetadataUpdate` action. 

Unless i've misunderstood something, there's no requirement to have both custody of the token_id (or nft_id) && control the pub/priv key pair in order to make a successful `NftMetadataUpdate`. 

Since the METADATA key is defined on the token_id, rather than on the nft_id, individual NFT holders will not be able to update each NFT which they custody unless of course they also maintain the `METADATA` private key.

HIP/hip-657.md
Ashe-Oro: > Can token creators make such an update when they do not own the NFT?
Yes, absolutely. The account which controls the private key associated with the METADATA public key can update the METADATA field on any NFT within the Collection (token_id) independently from which account currently custodies the NFT asset.

Ashe-Oro: > Can token creators set the NFT to be immutable/static in a way that future owners cannot make them dynamic?

Similar to all other keys on a token_id, if it is not set at the time of token creation, then the NFT Collection (token_id) is considered immutable whereas a new key cannot be added. This is consistent with all other keys on the token_id. 

Depending on the outcome of [HIP-540](https://github.com/hashgraph/hedera-improvement-proposal/pull/540), the METADATA key can remove itself thus entering it into the state of "no METADATA key defined" which also is considered immutable since no key can be re-added once removed (or not initially set).

Ashe-Oro: > Is the key type applicable at the token class of the individual NFT level?

The token_id (token class), but has the ability to update the METADATA value for any NFT within that token class.

HIP/hip-657.md
HGraphPunks: Right, category of type in this context being: SUPPLY, WIPE, ADMIN, etc. Not the way the key is generated or its encryption / encoding.

HIP/hip-657.md
mgarbs: @Ashe-Oro is right. The intention here was that only the entity in control of this key would be able to call `NftMetadataUpdate`. Individual holders would be unable to update their metadata unless they knew that key.

HIP/hip-657.md
mgarbs: I can remove the word type because I can see how it might get misconstrued

HIP/hip-657.md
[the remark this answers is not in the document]
mgarbs: Wouldnt it be good to know what the updates were though? Let's say we represent a property as an nft and all its attributes are defined in the metadata (price, floorplan, additions, etc) When updates happen to the property, it would be good to look onchain to verify to see what they were, no?

HIP/hip-657.md
```diff
@@ -0,0 +1,115 @@
...
+- As a developer building applications on top of the Hedera Network, I want the ability to update the metadata of NFTs so that I can create dynamic NFTs and build more advanced use cases.
+
+
+## Specification
+
+
+The METADATA key will be a new key that allows for updating the metadata of an NFT in a new NftMetadataUpdate transaction.
+
+The NftMetadataUpdate transaction should accept the following parameters:
+	- identifier (string): the 0.0.<token-type>.<serial-no>
```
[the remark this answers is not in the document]
mgarbs: Added

HIP/hip-657.md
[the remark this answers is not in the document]
Ashe-Oro: if it's possible to query previous values of `metadata` then that would be ideal, but i'm not certain how that would work and don't have a solution to propose.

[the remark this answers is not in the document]
mgarbs: Yea, true. This part would be out of scope for this hip. We will discuss a separate mirror node hip at a later date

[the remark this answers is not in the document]
nickpoorman: I don't think this is outside the scope of this HIP. We should take into consideration Mirror Node exposing this data. @steven-sheehy looping you in. Do you have any thoughts or see any concerns with having to surface multiple versions of NFT metadata?

nickpoorman:
cc @tinker-michaelj looping you in for your thoughts.

Ashe-Oro:
ATTN: 
Hi All - we are moving this HIP into `LAST CALL` and I ask you to please tag anyone who hasn't participated in the discussion thus far. Anyone who's building in the Hedera NFT ecosystem should be aware of the addition of a `METADATA` key and it's downstream effects. 

I'm mostly curious the impact of our various NFT tooling (Marketplaces, etc) given the fact that the `METADATA` field was once mutable, and may now become immutable depending on if a key is set. 

An architecture build with an assumption that `METADATA` would remain immutable may need to be refactored to account for the new mutability and I want to make sure these team/projects are aware of this HIP.

Ashe-Oro (approved):
Last call!

HIP/hip-657.md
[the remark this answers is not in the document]
Ashe-Oro: ping @steven-sheehy . I believe this is the last remaining open discussion for this HIP. 

Can/should the mirror node track the change history for metadata field values? IMO it would be extremally helpful if possible. 

Here's an example: 

My house is represented as an NFT. I replace the roof with a new 30-year warranty which causes and update to my NFT's metadata. Being able to review the history of my home's various states throughout time helps new buyers audit work done on my house.

itsbrandondev:
so incredible to see this HIP getting closer to the finish line!

i work with folks in the music industry and some are interested in bridging to web3. the problem is that many complex royalty structures for music require mutable NFTs. this is a major painpoint, but one that Hedera is in a unique position to solve.

im currently helping lead the Hashgraph NFT Working Group in their efforts to establish music specific metadata standards for Hedera and HIP-657 would allow these standards to meet the needs of the music industry.

there are so many use cases a metadata key unlocks, but even just for music specifically, this would be an industry game changer for complex music royalties that artists depend on.

having this added to the network would accelerate tokenized music efforts substantially and trigger a cambrian explosion of new innovations on the network. we also want to prevent projects from spending resources developing workarounds to substitute the lack of a metadata key.

counting down the days till this is on mainnet!

AdriMsM91:
Good night, I'm **COO & CoFounder of Kabila**.

Different projects have asked us on countless occasions for the possibility of dynamically updating the metadata, it is something that is already in almost every BlockChain and most people who come to Hedera request. By not having this system, many people are **pointing to centralized** servers, which leads to **discussions between their holders**, since they don't trust centralization, likewise, to this day if we make a mistake while we're minting an NFT, we don't have a possible solution other than **Burn that NFT and if we burn that NFT the SerialNumbers will never match with the Max Supply** of the collection again.

**Dynamic Metadata solves many problems**, but at the same time it opens the doors to infinite new minting techniques, such as Pre Reveal.

It seems to us that it is something very important for the Hedera ecosystem. Hopefully we can see this implemented soon.

GEstates:
Would love to use dynamic nfts for tokenized real estate so investors can keep track of renovations done to the homes just by seeing there nfts update! Will revolutionize fractional real estate!

PR #658, opened 2023-01-23
Update search

mgarbs:
This PR expands the search function to check the content in all hips and moves the search element inside the page content in `default.html`

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -4,10 +4,10 @@
 {%- include head.html -%}
 
 <body>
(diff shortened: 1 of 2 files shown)


david-bakin-sl (approved):
LGTM! (Searching for the search term "hollow" that prompted this PR finds two uses now, none before.)

nickpoorman (approved):

nathanklick (approved):
Approving. This PR does not modify or introduce a new HIP. It is exempt from the normal approval processes.

PR #673, opened 2023-02-08
Proposed changes to HIP-583

rbair23:
This changeset is not yet complete. I need to also update HIP-631 ad HIP-632, but at least we can get started with comments on HIP-583. There are a number of important changes:

- I added two new ideas to the "rejected" section that I didn't understand or like. We can bring them back if we feel like this was done in error. Specifically, I don't understand why we would allow hollow account creation when using `CryptoCreate`, and I do not think we should automatically set the alias when an `ECDSA` key is used, but should let the wallet do this.
- I changed the terminology. I am not a fan of `public-address` because it is not precise enough, and I found it confusing. It implies that 0.0.10 for example isn't a public address. Instead I wanted to clarify that we have three types of alias, and in this HIP specifically we're looking at `key alias` and `evm address alias` as the two principal formats (with `long-zero alias` being the third. Although I wish to call it something else as well, but I didn't have a good idea).
- I consolidated the use cases a little and added a few

There are a few TODOs to be addressed, and I think we should put support for adding an alias via account update into the HIP, the same as HIP-32 does, even if we don't implement it right away (but we should do so very soon).


rbair23:
I went ahead an updated the language to support `CryptoUpdate`, although we do not have to implement it right away.

bugbytesinc:
I can't find in the new language explaining that the **_only_** way to prove ownership of a hollow account is to sign a transaction where the hollow account is the **payer** of the transaction (this is how hedera intends to implement the feature, see issue https://github.com/hashgraph/hedera-services/issues/4863)  

Which, when thinking of it, how is the owner of the new account expected to find the account ID since payer can only be expressed in 0.0.123 form?  It seems to be a bit of extra work involved.

HIP/hip-583.md
```diff
@@ -2,221 +2,268 @@
...
+cannot take part in transactions submitted by others requiring its signature.
 
-As an illustration, a wallet or exchange could create and fund a new Hedera account based on the `public-address` alone. Once the transaction goes through the owner of the `public-address` user may use a wallet of their choice to submit a signed transaction (e.g. `CryptoTransfer`, `ContractCreate`, `ContractCall`, `EthereumTransaction` or `TokenAssociation`) to claim ownership of the account and carry out the desired transaction. The ownership claim and account completion will be seamless and transparent to the user.
+Since Hedera accounts require keys to ensure security, an `ECDSA` key matching the derived `evm address alias` must be
+provided in a future signed transaction issued by the owner of the account to take control of the account. Due to the
+nature of `ECDSA secp256k1` keys, the public key can be extracted from the signature and the signed bytes. It can then
+go through the derivation function to generate the associated `evm address`, which can be compared with the `alias` to
+prove the match and account ownership. This process will complete the creation of the account, taking it from a
```
MiroslavGatsanoga, line 182: Just to note an implementation detail for HAPI transactions: for the completion of the hollow account currently we are expecting a full-prefix signature (i.e. we have the entire public key in the signature) instead of deriving it in a way similar to `ecRecover`.

HIP/hip-583.md
```diff
@@ -2,221 +2,268 @@
...
+- `Hedera-native wallets` - wallets that utilize the Hedera SDK or gRPC HAPI calls to interact with Hedera consensus
+  and mirror nodes. Historically, accounts created via these wallets are `ED25519` based and usually identify the
+  account via the `Hedera Account Id`.
+- `auto-create` - a unique account creation flow in which the `key alias` or `evm address alias` that a crypto-transfer
+  is addressed to is used to automatically create a new account (if none exists) without an explicit `CryptoCreate`
+  transaction being submitted.
+- `Hollow account` - a partially created account which lacks a public key but does have a valid `shard.realm.num`
+  accountId _and_ is auto-created with an `evm address alias`. Such an account may receive assets but may not take
+  part in transactions that require its signature, until it is promoted to being a "full" account.
+- `lazy-create` - a unique account creation flow in which the public address (extracted from a provided ECDSA public
```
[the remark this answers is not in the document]
rbair23: True. From a linguistics perspective I think of auto-create being used to create hollow accounts or full accounts.

HIP/hip-583.md
```diff
@@ -2,221 +2,268 @@
...
-- Expand `CryptoTransfer` and `CryptoCreate` transactions to support the input of an alias of the Ethereum account address style with no public key
+This HIP proposes an expansion to the ways in which an account with an `alias` can be created and referenced to ensure
+compatibility with users coming from existing chains with `ECDSA` keys.
 
-### Add Public key → alias validation
+It is thus proposed to:
+- Update `CryptoCreate` transactions to support setting an alias using an `evm address alias`
+- Update `CryptoCreate` transactions to support setting an alias using a `key alias`
+- Update `CryptoTransfer` transactions to support auto-account creation for accounts with an `evm address alias`
+- Update `CryptoUpdate` transactions to support setting an alias using an `evm address alias`
```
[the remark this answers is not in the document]
rbair23: You are right, I need to remove the "Note".

HIP/hip-583.md
```diff
@@ -2,221 +2,268 @@
...
+An `Account` also has a system-assigned account identifier of the form `shard.realm.num`, where each value is an 8-byte
+number. It may also be referred to with an identifier of the form `shard.realm.alias`, where the alias is of one of
+the following forms:
+
+1. `long-zero alias` - This is the Hedera account number styled as an Ethereum account hex string. The
+   Hedera account number is hex encoded and the value is prefixed with zeros to form a 20 byte address. This format
+   cannot be used to auto-create an account, but can be used wherever an `alias` can be used.
+2. `key alias` - This format is created by using the bytes of a simple cryptographic public key supported by Hedera
+   (such as `ED25519` or `ECDSA secp256k1`). It is not necessary for the account's key to match the `alias`, only that
+   the account owner has asserted cryptographic proof of ownership of the account `alias` key.
```
Nana-EC, line 61: Q: Should we add that on creation if if a key alias is provided that it will match the account key, or are we providing anew feature here and relaxing this rule? If so it'll mean you can create an account with key A and then key alias B in one go

HIP/hip-583.md
```diff
@@ -2,221 +2,268 @@
...
+TODO: Are all three of these base-32 encoded? Should use consistent encoding
+
+To provide greater context for this HIP some additional terms useful:
+
+- `Ethereum-native wallets` - wallets that utilize the Ethereum JSON-RPC APIs implemented by the JSON-RPC Relay.
+  Accounts created via these wallets are `ECDSA` derived keys and use the `evm address alias` format.
+- `Hedera-native wallets` - wallets that utilize the Hedera SDK or gRPC HAPI calls to interact with Hedera consensus
+  and mirror nodes. Historically, accounts created via these wallets are `ED25519` based and usually identify the
+  account via the `Hedera Account Id`.
+- `auto-create` - a unique account creation flow in which the `key alias` or `evm address alias` that a crypto-transfer
```
Nana-EC: auto-create is only for full key alias scenario
```suggestion
- `auto-create` - a unique account creation flow in which the `key alias` that a crypto-transfer
```

HIP/hip-583.md
```diff
@@ -2,221 +2,268 @@
...
+  is addressed to is used to automatically create a new account (if none exists) without an explicit `CryptoCreate`
+  transaction being submitted.
```
Nana-EC:
```suggestion
  transaction being submitted. In this case the the resulting account with have a valid account key set matching the key alias provided.
```

HIP/hip-583.md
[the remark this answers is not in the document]
rbair23: However, "lazy create" as a way of capturing "auto-create with evm address alias leading to hollow account" is a useful shortcut, otherwise I have to write all of that every time.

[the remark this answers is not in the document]
Nana-EC: I don't think `evm address alias` should be added to the `auto crate` definition as an `auto-create` should produce a full account not a hollow account.

HIP/hip-583.md
```diff
@@ -2,221 +2,268 @@
...
+1. As a user with an `ECDSA` derived account from another chain, I would like to create my Hedera account using that
+   same ECDSA public key and associated Ethereum compatible `evm address alias`.
+2. As a developer, I would like to create a new account with a `key alias` via the `CryptoCreate` transaction.
+3. As a developer, I would like to create a new account with an `evm address alias` via the `CryptoCreate` transaction.
+4. As a developer, I would like to transfer HBAR or tokens to a Hedera account using their `evm address alias`.
+5. As a Hedera user with an `Ethereum-native wallet`, I would like to receive HBAR or tokens in my account by sharing
+   only my `evm address alias`.
+6. As a Hedera user with a `Hedera-native wallet`,  I would like to transfer HBAR or tokens to another account using
+   only the recipient's `evm address alias`.
+7. As a Hedera user with an `ED25519` key on my account, I want to add an `evm address alias` to my account.
```
Nana-EC, line 113: Q: To clarify this is a new feature for HIP 583.
Essentially CryptoUpdate and pass in an evm address alias to unblock accounts with no evm address alias in state or memory?

HIP/hip-583.md
```diff
@@ -2,221 +2,268 @@
...
-- Expand `CryptoCreate` transactions to support setting an alias using a public-key
-- Expand `CryptoTransfer` and `CryptoCreate` transactions to support the input of an alias of the Ethereum account address style with no public key
+This HIP proposes an expansion to the ways in which an account with an `alias` can be created and referenced to ensure
+compatibility with users coming from existing chains with `ECDSA` keys.
 
-### Add Public key → alias validation
+It is thus proposed to:
+- Update `CryptoCreate` transactions to support setting an alias using an `evm address alias`
+- Update `CryptoCreate` transactions to support setting an alias using a `key alias`
+- Update `CryptoTransfer` transactions to support auto-account creation for accounts with an `evm address alias`
```
Nana-EC, line 123:
```suggestion
- Update `CryptoTransfer` transactions to support lazy-account creation for accounts with an `evm address alias`
```

HIP/hip-583.md
rbair23: OK, I left this language the same as was in the previous version, so it sounds like we did not implement it as specified in the HIP? From @bugbytesinc comment above, it sounds like we have an issue here to resolve.

HIP/hip-583.md
```diff
@@ -2,221 +2,268 @@
...
-Note, in this case the final value of the account alias will remain as the `public-address` and not the `public-key` as it would be in the `auto-create` case.
+When a user transfers value (hbars or tokens), they may address their destination account using the standard
+account ID format (`shard.realm.num`), or by using an alias format (`shard.realm.alias`). If the alias is an
+`evm address alias`, then the consensus node will first check to see if such an account exists. If it does,
+then it will use the normal rules for transferring the hbars or tokens (including checking `receiverSigRequired`,
+and all other rules). If the account *does not* exist, then the account will be auto-created with the same rules
```
Nana-EC, line 169: `auto-create` is used again here where `lazy-create` was used before.
Are we trying to remove the lazy-create term or is this a mixup?
Clarifying to confirm terminology implications

HIP/hip-583.md
Nana-EC: Yeah let's call this nuance out.
It doesn't affect work as this is the case when utilizing SDK's but it's good to note.

rbair23:
> I can't find in the new language explaining that the **_only_** way to prove ownership of a hollow account is to sign a transaction where the hollow account is the **payer** of the transaction (this is how hedera intends to implement the feature, see issue [hashgraph/hedera-services#4863](https://github.com/hashgraph/hedera-services/issues/4863))

Ah, thanks for this, it is new to me. I did not put language saying the *payer* has have the signatures because it seems to me that either the payer signature should match the key represented by the alias, or the transaction should be signed by the payer AND the key representing the alias. We just need the transaction to be signed by the key represented by the alias to assert that the person owning that key approves the transaction. But I don't see why it needs to be the payer.

> Which, when thinking of it, how is the owner of the new account expected to find the account ID since payer can only be expressed in 0.0.123 form? It seems to be a bit of extra work involved.

Oh! Why don't we support payer to use alias?

HIP/hip-583.md
```diff
@@ -2,221 +2,268 @@
...
+the `evm address alias` to be used.
 
-### Allow CryptoCreate to automatically set alias
+### Setting `alias` with `CryptoUpdate`
 
-When `CryptoCreate.key` is an `ECDSA` key, the network node may calculate the `public-address` and automatically set the accounts alias value. 
+`HIP-32` supports updating the `alias` of an account after the account has been created. This HIP extends that
+support to include `evm address alias`es. As with `CryptoCreate`, ownership of the private key of the `ECDSA` public
+key associated with the `evm address` **must** be asserted on the `CryptoUpdate` by signing the transaction with
+that private key. The key used does **not** have to be in state on any account.
```
Nana-EC, line 193: We should highlight the restriction here for an account that already has a key alias (ED or EC) set.
Multiple alias in state should be outside of this HIP.

HIP/hip-583.md
```diff
@@ -2,221 +2,268 @@
...
 
-- A special reserved `ECDSA` based `aliasKey` on all accounts. This would have ensured only a single key was used on contract interactions.
-- A special reserved `ECDSA` based `aliasKey` and an additional `evmAlias`. This would have provided complete separation of all EVM interactions with Hedera account keys.
-
-In a way, this section can be thought of as a breakout section of the Rationale section that focuses specifically on why certain ideas were not ultimately pursued.
+- A special reserved `ECDSA` based `aliasKey` on all accounts. This would have ensured only a single key was used on
+  contract interactions.
+- A special reserved `ECDSA` based `aliasKey` and an additional `evmAlias`. This would have provided complete separation
+  of all EVM interactions with Hedera account keys.
+- The ability to call `CryptoCreate` with just the `evm address alias` and not the public key. Use a `CryptoTransfer`
```
Nana-EC, line 254: Disagree here.
This was an explicit request by wallets to support their current flows.
Creating 0 balance accounts is important to them.
Additionally, we can assume developers would have access to the public key used to derive the evm address, but we can't presume all business practices will have the public key in the same place as the evm address.
Many web3 flows might preserve a list of evm address in one places and the private keys in another. They rarely deal with public keys as Hedera does

Nana-EC (commented):
Some feedback on naming consistencies, additional features and restrictions

HIP/hip-583.md
```diff
@@ -2,221 +2,268 @@
...
+are fundamentally different entities. An `Account` must have a key. The key for the account may be a simple key
+(typically, an `ED25519` public key), or it may be a compound key, such as a list of keys all of which must sign each
+transaction, or a threshold key such that N of the keys must be signed. These may be nested to create very rich and
+powerful accounts requiring multiple-signatures. Further, an `Account` may change the keys on the account over time.
+
+An `Account` also has a system-assigned account identifier of the form `shard.realm.num`, where each value is an 8-byte
+number. It may also be referred to with an identifier of the form `shard.realm.alias`, where the alias is of one of
+the following forms:
+
+1. `long-zero alias` - This is the Hedera account number styled as an Ethereum account hex string. The
```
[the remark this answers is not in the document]
rbair23: I am intentionally leaving them out. The shard and realm are already part of the AccountID, so duplicating them here seems dubious. The shard of the AccountId is 8 bytes, when encoded here it is only 4 bytes. So if encoding shard and realm in the long zero is actually important, we are basically wasting 4 bytes in the account ID that we'll never be able to use. And if we encode shard and realm into the long-zero, then we cannot effectively filter out long-zero aliases from contract create, or from the auto-create flow, meaning somebody can "squat" on a long-zero address.

HIP/hip-583.md
rbair23: Yes. I wasn't going to put this into the HIP, until I saw that HIP-32 already has language for supporting alias on account update! So we *have* to have this language in this HIP to clarify what happens on account update, since HIP-32 already mandates it (even if we haven't implemented it yet).

HIP/hip-583.md
```diff
@@ -2,221 +2,268 @@
...
 ```
 
-Additionally, if an account is created with an `ECDSA` public key the alias property should be set automatically at creation to ensure smart contract compatibility from creation. 
+#### Add Public key → alias validation
 
-### Account Creation via alias only (Lazy Create)
-
-![lazy-create-with-address.png](../assets/hip-583/lazy-create-with-address.png)
+If the `alias` is set during `CryptoCreate`, then ownership of the `alias` **must** be proven by the caller
+by signing the transaction with the key represented by the `alias`. It is not necessary for the `key` on the
```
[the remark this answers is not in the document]
rbair23: Yes. All we need is to verify that somebody who owns the key associated with the alias, approved the transaction.

HIP/hip-583.md
rbair23: Yes, I was trying to clarify the terms. Auto-creation is automatic-creation, which does not imply full account or hollow account. Any thing that is created automatically falls under the umbrella of auto-creation. Lazy creation means we don't create something until needed -- but actually we *are* creating something, we're creating a hollow account. So I changed the terminology so that auto-creation encompasses anything that is automatically created, and hollow creation is specifically the creation of a hollow account.

HIP/hip-583.md
rbair23: Agree!

HIP/hip-583.md
rbair23: I do not understand.

If the account has zero balance, why would I want to pay for the account's creation? If the account is not going to have zero balance, then I will need to do a crypto transfer to create the account, in which case it will be auto-created if I do the transfer first. If I have custody of the key, then I can create a normal account. If I do not have custody of the key, and I do no intend to transfer HBAR into the account, why would I pay to create the account?

HIP/hip-583.md
```diff
@@ -2,221 +2,266 @@
...
+1. As a user with an `ECDSA` derived account from another chain, I would like to create my Hedera account using that
+   same ECDSA public key and associated Ethereum compatible `evm address alias`.
+2. As a developer, I would like to create a new account with a `key alias` via the `CryptoCreate` transaction.
+3. As a developer, I would like to create a new account with an `evm address alias` via the `CryptoCreate` transaction.
+4. As a developer, I would like to transfer HBAR or tokens to a Hedera account using their `evm address alias`.
+5. As a Hedera user with an `Ethereum-native wallet`, I would like to receive HBAR or tokens in my account by sharing
+   only my `evm address alias`.
+6. As a Hedera user with a `Hedera-native wallet`,  I would like to transfer HBAR or tokens to another account using
+   only the recipient's `evm address alias`.
+7. As a Hedera user with an `ED25519` key on my account, I want to add an `evm address alias` to my account.
```
steven-sheehy, line 113: We could expand this use case or add a separate use case for when users have a complex key on the account and want to have an `evm address alias`.

HIP/hip-583.md
```diff
@@ -2,221 +2,266 @@
...
+are fundamentally different entities. An `Account` must have a key. The key for the account may be a simple key
+(typically, an `ED25519` public key), or it may be a compound key, such as a list of keys all of which must sign each
+transaction, or a threshold key such that N of the keys must be signed. These may be nested to create very rich and
+powerful accounts requiring multiple-signatures. Further, an `Account` may change the keys on the account over time.
+
+An `Account` also has a system-assigned account identifier of the form `shard.realm.num`, where each value is an 8-byte
+number. An `Account` may also be referred to with an identifier of the form `shard.realm.alias`, where the alias is of one of
+the following forms:
+
+1. `long-zero alias` - This is the Hedera account number styled as an Ethereum account hex string. The
```
steven-sheehy, line 56: Suggest `entity num alias` or `entity id alias`

HIP/hip-583.md
```diff
@@ -2,221 +2,266 @@
...
 type: Standards Track
 category: Service
 needs-council-approval: Yes
 status: Accepted
 last-call-date-time: 2022-10-17T07:00:00Z
 created: 2022-09-20
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/593
-updated: 2022-09-29, 2022-10-03, 2022-10-20, 2022-11-11
+updated: 2022-09-29, 2022-10-03, 2022-10-20, 2022-11-11, 2023-02-07
 requires: 32
```
steven-sheehy, line 16: Should this supersede 631 or how do we go about indicating 631 will not be implemented?

steven-sheehy (dismissed):
LGTM. Just some nits

HIP/hip-583.md
Nana-EC: I think there's a potential here, but it remains a hypothesis and for that reason it's best to not add.
I'm okay with this as a rejected idea.
if we get the data we can add it when propose it again when the time is right

HIP/hip-583.md
Nana-EC: Makes sense.
All DI frameworks would agree also :)

HIP/hip-583.md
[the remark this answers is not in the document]
Nana-EC: gotcha

HIP/hip-583.md
stoqnkpL: @rbair23 there are a couple of different scenarios for hollow account completion.
1. In the context of an `EthereumTransaction` flow we do extract the public `ECDSA` key from the raw Ethereum signed transaction and use that key for the completion of the hollow account. Reference: [SigsAndPayerKeyScreen](https://github.com/hashgraph/hedera-services/blob/ac9d19c615dfbbd9d3876cfe1fcb017373fb6ae6/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/logic/SigsAndPayerKeyScreen.java#L124). So that part is correct and implemented.
3. In the context of a `HAPI` transaction flow. There is no `ecrecover`ing for HAPI transactions. Instead, we look at the `payer` of incoming transactions, and try to match the full-prefix public key to the hollow account `EVM address alias`. I think @MiroslavGatsanoga was referencing this flow.

It has to be the payer. It was implemented like this due to performance considerations. Searching through the whole signature map of each incoming transaction and trying to find matching hollow accounts… is this something we want to do?

HIP/hip-583.md
```diff
@@ -2,221 +2,266 @@
...
+nature of `ECDSA secp256k1` keys, the public key can be extracted from the signature and the signed bytes. It can then
+go through the derivation function to generate the associated `evm address`, which can be compared with the `alias` to
+prove the match and account ownership. This process will complete the creation of the account, taking it from a
+`Hollow Account` to a normal complete account with a valid public key.
+
+Note, in this case the final value of the account alias will remain as the `evm address alias` and not the `key alias`.
+
+`EthereumTransaction` transactions where value is transferred  (i.e `tx.to` and `tx.value` are set) should also support
+this case. This will support users of `Ethereum-native wallets` and eth dev tools who are accustomed to creating and
+crediting an account's balance in one step. In this case th `tx.to` value of the RLP encoded transaction would dictate
```
stoqnkpL, line 185: nit: typo in `the`

HIP/hip-583.md
```diff
@@ -2,221 +2,266 @@
...
+the `evm address alias` to be used.
+
+### Setting `alias` with `CryptoUpdate`
+
+`HIP-32` supports updating the `alias` of an account after the account has been created. This HIP extends that
+support to include `evm address alias`es. As with `CryptoCreate`, ownership of the private key of the `ECDSA` public
+key associated with the `evm address` **must** be asserted on the `CryptoUpdate` by signing the transaction with
+that private key. The key used does **not** have to be in state on any account. As with `HIP-32`, there can only
+be a single `alias` per account (except for the `long-zero alias` which all accounts have), and it is immutable, and
+therefore an account that has an `alias` cannot have it updated.
```
stoqnkpL, line 195: Technically, an account can have up to three aliases linked to it (even before HIP-583):
1. the implicit `long-zero alias` which all accounts have
2. an `ECDSA` public key alias set via auto account creation
3. the public `EVM address` derived from that key

Only in some cases is the public `EVM address` alias actually stored in the `alias` field of the account in state, but it is important to note that this alias belongs to the account and the nodes are aware of it.

HIP/hip-583.md
```diff
@@ -2,221 +2,266 @@
...
+`CryptoCreate` transactions will see extra bytes for every account created with an alias.
 
-`CryptoCreate` transactions will see extra bytes for every account created with an alias
-
-The size of the bytes in each case depends on the governing account key type. For an `ECDSA` based account the alias will always be a valid Ethereum address and will be 20 bytes long. In the `ED25519` case the alias will be a valid protobuf serialized public key of 35 or 68 bytes depending on encoding.
+The size of the bytes in each case depends on the governing account key type. For an `ECDSA` based account the alias
+will always be a valid Ethereum address and will be 20 bytes long. In the `ED25519` case the alias will be a valid
+protobuf serialized public key of 35 or 68 bytes depending on encoding.
 
 ## Security Implications
```
stoqnkpL, line 231: Maybe it also makes sense to include the fee-charging strategy information for hollow account creation in different scenarios. Basically, Hedera will consume a larger amount of gas from the gas limit than is provisioned by the native EVM in case we're creating accounts.
We may also want to consider throttling creations in the EVM. It is possible to create 20+ accounts per second via smart contract calls even at current gas prices. Via HAPI transactions, creations are throttled at 2 per second I believe.

HIP/hip-583.md
Nana-EC: To be clear I think in the case where account key is A and key alias is B, we'd preserve existing network logic that checks that the account id key is present in the signature map and for alias the key is also present in the signature map.

HIP/hip-583.md
rbair23: 631 depends on this one, not the other way around. I think that means this one doesn't supersede the other, but that the other is out of date and needs to be updated.

HIP/hip-583.md
```diff
@@ -2,221 +2,266 @@
...
 
 ## User stories
 
-1. As a user with an `ECDSA` based account from another chain I would like to create my Hedera account using my `public-address`.
-
-2. As a developer, I would like to `auto-create` a new account using a public key via the `CryptoCreate` transaction.
-
-3. As a developer, I would like to `lazy-create` a new account using a `public-address` via the `CryptoCreate` transaction.
+1. As a user with an `ECDSA` derived account from another chain, I would like to create my Hedera account using that
+   same ECDSA public key and associated Ethereum compatible `evm address alias`.
```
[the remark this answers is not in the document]
rbair23: You are right! I was thinking "A user with an Ethereum-compatible wallet". I will update the user store to clarify.

HIP/hip-583.md
rbair23: Thanks!

HIP/hip-583.md
rbair23: Yes exactly.

HIP/hip-583.md
rbair23: OK, language added.

HIP/hip-583.md
rbair23: OK, I switched direction on this. I decided to **supersede** this language from HIP-32 to indicate that we do NOT support adding an alias to an existing account, with a section at the bottom explaining why.

HIP/hip-583.md
rbair23: I agree, if we are going to support update on accounts, there is more to be specified. I found a critical complication with update, where from the EVM perspective an updated account may have 2 valid aliases (the long-zero and the evm address alias) and now we have a new problem -- how to decide which the EVM should use for a particular contract call. So I opted to remove the update language entirely, including this user story.

HIP/hip-583.md
rbair23: I went with `account num alias` since I think it only applies to accounts (... and contracts... but they are accounts).

HIP/hip-583.md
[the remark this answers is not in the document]
rbair23: I think actually you would be OK using an Ethereum wallet and JSON-RPC relay if you use your long-zero address, AS LONG AS the contract you are calling isn't using `ecrecover`. If it is, then I agree, it won't work. For HIP-583, I think we accept that limitation, with the workaround being, you have to create a new account 😭 . But, since updating an existing account introduces a number of new wrinkles, including new work on the wallets, I think it is better to live with this limitation for the moment in order to onboard more users to Hedera and work on this in a new HIP.

HIP/hip-583.md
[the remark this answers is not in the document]
rbair23: I have updated the use case. Thanks!

HIP/hip-583.md
```diff
@@ -2,221 +2,266 @@
...
+  in the creation of a `Hollow account`.
+
+With an Ethereum-native wallet, you can transfer the native currency (in our case, HBAR), or tokens, to an `evm address`
+that does not yet exist. If nobody in the universe has the private key for that account, the value held by that account
+is effectively locked forever. If somebody does have the keys to that account, they can access the account and control
+it. Since Ethereum-native wallets do not understand the `0.0.10` style format of a Hedera account ID, it is necessary
+to support the creation of accounts with an `evm address alias` to enable compatibility between Ethereum-native
+wallets and Hedera-native wallets and all accounts.
 
 ## User stories
```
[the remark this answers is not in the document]
rbair23: I like it. I think from an implementation perspective it is covered by this use case I added today (before I saw your comment):

> 7. As a Hedera user with a `Hedera-native wallet`,  I would like to transfer HBAR or tokens to another account using
   only the recipient's `evm address alias`.

However I think it is good to get your user story as well since it may be useful for documentation and outreach and shows one reason why we are doing what we are doing.

HIP/hip-583.md
rbair23: I agree, this is said very well. I have added language to the HIP which I hope is as good as what you have here. You are exactly right.

HIP/hip-583.md
rbair23: You are right, we should list how we plan to pay for things. I have to admit, I do not know who is paying for what in different scenarios and we should include that in the HIP in a special section under the specification.

HIP/hip-583.md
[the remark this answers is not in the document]
stoqnkpL: You can't work with an Ethereum wallet and JSON-RPC relay without signing with an ECDSA key that can be recovered and the resulting EVM address matched to an existing alias belonging to an account in Hedera.

WilliamJordan33:
Hello Everyone,

I am having trouble sending my HBAR from an exchange to my Ledger. I have downloaded the hashpack, made an account and created the wallet. There are two checksums; one with decimals and numbers and one with decimals numbers a dash and letters at the end. When I am connected to the Ledger to the Hedera App, I open it, I click receive on Ledger live and the address that comes up is only the first checksum without the digits at the end. There is no way to change either of the checksums on Ledger Live or hashpack but a support robot that I emailed from the hashpack website told me to delete the last checksum digits which is impossible. There is a warning at the bottom saying that, "Your Hedera address cannot be confirmed on your ledger device" I sent 2 HBAR from the exchange to the checksum, remembered to put the MEMO along with the address that was on Ledger Live and my hashpack account and nothing came through. I sent another 2 HBAR from the exchange and nothing went through as well. I am furious because this checksum is not making things more secure but it is putting my assets at risk. All of my other crypto currencies that did not add this extra layer of unnecessary "security" are safe and secure on my Ledger. Ledger is not really compatible with the checksum and my assets are now at risk. For the love of God can someone please help me?

Sincerely,

William Jordan

WilliamJordan33:
Hello Everyone,

I am having trouble sending my HBAR from an exchange to my Ledger. I have downloaded the hashpack, made an account and created the wallet. There are two checksums; one with decimals and numbers and one with decimals numbers a dash and letters at the end. When I am connected to the Ledger to the Hedera App, I open it, I click receive on Ledger live and the address that comes up is only the first checksum without the digits at the end. There is no way to change either of the checksums on Ledger Live or hashpack but a support robot that I emailed from the hashpack website told me to delete the last checksum digits which is impossible. There is a warning at the bottom saying that, "Your Hedera address cannot be confirmed on your ledger device" I sent 2 HBAR from the exchange to the checksum, remembered to put the MEMO along with the address that was on Ledger Live and my hashpack account and nothing came through. I sent another 2 HBAR from the exchange and nothing went through as well. I am furious because this checksum is not making things more secure but it is putting my assets at risk. All of my other crypto currencies that did not add this extra layer of unnecessary "security" are safe and secure on my Ledger. Ledger is not really compatible with the checksum and my assets are now at risk. For the love of God can someone please help me?

Sincerely,

William Jordan

WilliamJordan33:
Another problem with hashpack is that there is not a direct way for its users to contact the creators which are you. That would be the most efficient way to solve problems because any of the problems that we encounter, we could just directly tell you guys. 

Another problem is that Ledger does not trust hashpack HIP-15. Ledger is the most well respected and trusted hardware wallets in the world. It is your responsibility to have a good relationship with Ledger and to ensure that the corresponding technologies can operate in a seamless confluence. 

My Ledger Live does not trust your checksums or the QR code option as well. I have lost crypto assets on your behalf and I had to go out of my way to enlighten you guys. You guys are partnered with Hedera which has the potential to be one of the most revolutionary companies in the history of the entire world but this revolution will not happen if you are not being clear, transparent and efficient with our security. 

Do not take this personally. I just want my assets to be safe, that it all.

HIP/hip-583.md
dimitar-dinev: Just to make sure we do not miss something:

The current signing requirements for `CryptoCreate` [only require a signature from the new account key, i.e. the `adminКey`, when the `receiverSigRequried` is being set to `true`.](https://github.com/hashgraph/hedera-services/blob/8d6869569f5874f3ad033907f417b0a7fb363fe6/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/sigs/order/SigRequirements.java#L775-L785) Otherwise, no signature is needed (except for the payer's, of course).

Are we aiming to change this?

HIP/hip-583.md
```diff
@@ -2,228 +2,461 @@
...
+`evm address alias`. However, it is **impossible** to auto-create an account using the `account-num alias`. If the
+account does not exist, the transfer will fail.
 
-`EthereumTransaction` transactions where value is transfered  (i.e `tx.to` and `tx.value` are set) should also support this case.
-This will support users of `Ethereum-native wallets` and eth dev tools who are accumstomed to creating and crediting an accounts balance in one step.
-In this case th `tx.to` value of the RLP encoded transaction would dictate the `public-address` to be used as an alias.
+##### `key alias`
 
-### Allow CryptoCreate to automatically set alias
+The `key alias` is a protobuf-encoded `Key` object. We only support `ED25519` and `ECDSA_secp256k1` keys at this time.
```
Nana-EC, line 312: I think this section `key alias` and `evm address alias` are repeats of a previous secton

Nana-EC (changes requested):
Looking good.
Some suggestions on wording and a request to remove the reject ida of `CryptoCreate` w `evm_address` from the other sections noting support

HIP/hip-583.md
```diff
@@ -10,220 +10,451 @@ status: Accepted
...
+When an attempt is made to create a new account with `CryptoCreate` with an alias, we check whether it is a `key alias`
+by checking if there are more than 20 bytes. We then:
 
-When `CryptoCreate.key` is an `ECDSA` key, the network node may calculate the `public-address` and automatically set the accounts alias value. 
+1. Deserialize the protobuf object. If the object is not a protobuf object, we reject the transaction.
+2. If the protobuf object is not an `ED25519` or `ECDSA_secp256k1` `Key`, we reject the transaction.
+3. If the `Key` is an `ED25519` key:
+    1. If the alias is already in use, we use that account and do not create a new one
```
stoqnkpL, line 317: We would actually reject the transaction since this is a `CryptoCreate` and not a `CryptoTransfer`

HIP/hip-583.md
```diff
@@ -10,220 +10,451 @@ status: Accepted
...
+    2. If the alias is not in use, we create a new account and use the given key as the key on the account
+4. If the `Key` is an `ECDSA_secp256k1` key:
+    1. If the alias is already in use, we use that account and do not create a new one
+    2. If the derived `evm address` from the key is already in use, we use that account and do not create a new one
+    3. If the alias is not in use, we create a new account and use the given key as the key on the account.
 
-## Backwards Compatibility
+If we do create a new account from either `CryptoCreate` or `CryptoTransfer`, and the `Key` is an `ECDSA_secp256k1` key,
+then we reserve in the aliases for **both** the raw protobuf-encoded `Key` bytes **and** the derived evm address.
```
stoqnkpL, line 325: If we create a new account with an `ECDSA_secp256k1` key, we no longer automatically reserve the raw protobuf-encoded `Key` bytes **and** the derived evm address aliases. Or do you mean only if the account has that key as an alias? In that case, it's correct.

HIP/hip-583.md
```diff
@@ -10,220 +10,451 @@ status: Accepted
...
-This will support users, developers and exchanges who prefer to create and credit an accounts balance in one step.
+It **is** possible to specify an `account-num alias` as the _target_ of a `CryptoTransfer` transaction. Indeed, for
+Ethereum-native wallets, this is the only way to transfer value to Hedera accounts that don't have an
+`evm address alias`. However, it is **impossible** to auto-create an account using the `account-num alias`. If the
+account does not exist, the transfer will fail.
 
-`EthereumTransaction` transactions where value is transfered  (i.e `tx.to` and `tx.value` are set) should also support this case.
-This will support users of `Ethereum-native wallets` and eth dev tools who are accumstomed to creating and crediting an accounts balance in one step.
-In this case th `tx.to` value of the RLP encoded transaction would dictate the `public-address` to be used as an alias.
+##### `key alias`
```
MiroslavGatsanoga, line 308: It seems like this section was copied from the `CryptoCreate` section above. I think we don't need `key alias` section for `CryptoTransfer` because those cases are described in HIP-32 and we are not making any changes in HIP-583 for `CryptoTransfer` with key aliases.

HIP/hip-583.md
stoqnkpL: With https://github.com/hashgraph/hedera-services/pull/4990 we're moving the account finalization logic to prehandle and updating it so that _any_ full-prefixed ECDSA public key from the sig map will be matched to _any_ existing hollow account in Hedera and used to finalize it.

bugbytesinc: It is my understanding from the preliminary implementation (on previewnet) that this is ONLY implemented for PAYER accounts.  Meaning that only the payer signature will be checked for ecrecover, any other signature will not be examined in this process.  So if you have a service that has nominated itself to be the payer in a TX which is still _also_ signed_ by an secp256k1 matching a hollow account participating in the TX, the TX _will fail because the secp256k1 key is not the payer_

Can someone please clarify the intent of the HIP and how it will be implemented?

HIP/hip-583.md
```diff
@@ -2,228 +2,461 @@
...
+This HIP allows _any_ new account on Hedera to have an `ECDSA` derived `alias`, compatible with Ethereum addresses, and
+permits the use of this `alias` for all operations supported by `alias` today. This HIP expands support in the
+auto-create flow for `ECDSA` derived addresses to be used as the `alias`. It also adds support to `CryptoCreate`
+transactions for creation with either a key-based `alias` or, an `ECDSA` derived `alias`.
+
+Rationale
+------------------------------------------------------------------------------------------------------------------------
+
+Fundamental in the design of the Hedera network are the concepts of _accounts_, account _identifiers_, and _keys_. These
+are fundamentally different entities. An `Account` must have a key. The key for the account may be a simple key
```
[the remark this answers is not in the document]
rbair23: I guess it is not even true then, because some accounts like `0.0.800` have no key at all. Maybe they aren't a full or complete account, but maybe that terminology isn't right, because it isn't hollow either. What if I just say:

> Today, a user `Account` must have a key

I wanted to avoid talking about the HIP proposal in the preamble, but just explain the situation today.

HIP/hip-583.md
```diff
@@ -2,228 +2,461 @@
...
+powerful accounts requiring multiple-signatures. Further, an `Account` may change the keys on the account over time.
+
+An `Account` also has a system-assigned account identifier of the form `shard.realm.num`, where each value is an 8-byte
+number. An `Account` may also be referred to with an identifier of the form `shard.realm.alias`, where the alias is of
+one of the following forms:
+
+1. `account-num alias` - This is the Hedera account number styled as an Ethereum account hex string. The
+   Hedera account number is hex encoded and the value is prefixed with zeros to form a 20 byte address. This format
+   cannot be used to auto-create an account, but can be used wherever an `alias` can be used. This is sometimes
+   referred to as the "long-zero" alias, since the first 12 bytes are 0.
```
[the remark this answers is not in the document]
rbair23: OK

HIP/hip-583.md
```diff
@@ -2,228 +2,461 @@
...
+Examples of each format are noted below:
+(Please NOTE: the `alias` is a binary format when encoded on-wire for protobuf. We represent the bytes as hexidecimal
+ASCII for clarity in the documentation, but in reality these are raw bytes)
+
+| Identifier Type           | Example                                                                  |
+|---------------------------|--------------------------------------------------------------------------|
+| Hedera Account ID         | 0.0.10                                                                   |
+| Hedera Account Num Alias  | 0.0.000000000000000000000000000000000000000a                             |
+| Hedera Key Alias          | 0.0.1220d75a980182b10ab7d54bfed3c964073a0ee172f3daa62325af021a68f707511a |
+| EVM Address Alias         | 0.0.b794f5ea0ba39494ce839613fffba74279579268                             |
```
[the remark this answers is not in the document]
rbair23: This table is listing the account IDs, which do not start with 0x (if you tried to include those bytes in the alias, they wouldn't work). But we could have a secondary table showing that?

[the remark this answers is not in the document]
rbair23: I'll just add an additional column.

HIP/hip-583.md
rbair23: Correct! Fixing the doc.

rbair23: Ah, I see the problem. This section is talking about CryptoTransfer, but mistakingly says it is CryptoCreate. I will correct.

HIP/hip-583.md
rbair23: Yes, only in the case that the account has the alias.

rbair23: Updating the language:

>If we do create a new account from either `CryptoCreate` or `CryptoTransfer`, and the `alias` is an `ECDSA_secp256k1` key, then we reserve in the aliases for **both** the raw protobuf-encoded `Key` bytes **and** the derived evm address. Essentially, this allows the account created with an `ECDSA_secp256k1` to be addressed by that alias, the `evm address alias`, and the `account-num alias`.

HIP/hip-583.md
rbair23: Done.

HIP/hip-583.md
rbair23: Done

HIP/hip-583.md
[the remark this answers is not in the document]
rbair23: Done

HIP/hip-583.md
rbair23: It is not an exactly copy-paste, there are important differences in the behavior. While HIP-32 does define this capability. I wanted to be explicit with how this works with `evm address alias`, which is new in this HIP. You will notice, for example:

> 4. If the `Key` is an `ECDSA_secp256k1` key:
>     1. If the alias is already in use, we use that account and do not create a new one
>     2. If the derived `evm address` from the key is already in use, we use that account and do not create a new one
>     3. If the alias is not in use, we create a new account and use the given key as the key on the account.

The second bullet explicitly calls out the behavior with evm address aliases involved. I any event, I want to be explicit here to help define the test cases that must be written and verified before the feature is complete.

HIP/hip-583.md
```diff
@@ -10,220 +10,452 @@ status: Accepted
...
+
+From the perspective of the Ethereum ecosystem, the `account-num alias` and `evm address alias` are both
+directly usable when prefixed with `0x`. Examples of each format are noted below:
+(Please NOTE: the `alias` is a binary format when encoded on-wire for protobuf. We represent the bytes as hexidecimal
+ASCII for clarity in the documentation, but in reality these are raw bytes)
+
+| Identifier Type           | Account ID Example (Protobuf)                                            | Ethereum Address Example (String)          |
+|---------------------------|--------------------------------------------------------------------------|--------------------------------------------|
+| Hedera Account ID         | 0.0.10                                                                   | N/A                                        |
+| Hedera Account Num Alias  | 0.0.000000000000000000000000000000000000000a                             | 0x000000000000000000000000000000000000000a |
```
rbair23, line 90: We need to document clearly that in the extremely rare chance that you have a private key that produces an evm address with 12 leading zero bytes (96 bits!!)... I'm sorry. It won't work on our network properly.

HIP/hip-583.md
```diff
@@ -10,220 +10,452 @@ status: Accepted
...
+expectations. While this is an important user flow we want to support, we defer this support to a future HIP. A full
+specification for this capability will likely require the ability for the user to select which alias is their default,
+and to specify on a per-call basis the alias they want to use when interacting with a smart contract.
+
+User stories
+------------------------------------------------------------------------------------------------------------------------
+
+1. As a user with an `ECDSA` derived account from another chain who uses an Ethereum-compatible wallet, I would like to
+   create my Hedera account using that same ECDSA public key and associated Ethereum compatible `evm address alias`.
+2. As a developer, I would like to create a new account with a `key alias` via the `CryptoCreate` transaction.
```
rbair23, line 160: Is this actually something we want? Should we deprecate support for `key alias` instead, and not dig the hole any deeper?

HIP/hip-583.md
```diff
@@ -10,220 +10,452 @@ status: Accepted
...
+3. As a developer, I would like to create a new account with an `evm address alias` via the `CryptoCreate` transaction.
+4. As a developer, I would like to transfer HBAR or tokens to a Hedera account using their `evm address alias`.
```
rbair23, line 162: Maybe separate out into three user stories, one that says using HAPI, one using EthereumTransaction, one using code in a smart contract.

HIP/hip-583.md
```diff
@@ -10,220 +10,452 @@ status: Accepted
...
+   the account owner has asserted cryptographic proof of ownership of the account `alias` key in addition to the
+   account key. The key is protobuf-encoded and those raw bytes are included in the `alias`.
+3. `evm address alias` - This format is created by using the rightmost 20 bytes of the 32 byte `Keccak-256` hash of an
+   `ECDSA secp256k1` public key. This calculation is in the manner described by the Ethereum Yellow Paper.
+   Note, that the recovery id is not formally part of the public key and not included in the hash. This is
+   calculated on the consensus nodes using the `ECDSA` key provided in the auto create `CryptoTransfer` transaction.
+   While both `evm address alias` and `account-num alias` are 20 byte values, they can be differentiated by the fact
+   that `account-num alias` is always prefixed with 12 bytes. The likelihood of randomly generated ECDSA keys, being
+   hashed, having 12 sequential zeros is very low. Those keys, should they be found, cannot be used on the network.
+
```
rbair23, line 81: Add a 4th point that all accounts have `account-num alias` and _may_ have one additional alias that is either `key alias` or `evm address alias`.

HIP/hip-583.md
```diff
@@ -10,220 +10,452 @@ status: Accepted
...
+Specification
+------------------------------------------------------------------------------------------------------------------------
+
+This HIP proposes an expansion to the ways in which an account with an `alias` can be created and referenced to ensure
+compatibility with users coming from existing chains with `ECDSA` keys.
+
+It is thus proposed to:
+- Update `CryptoCreate` transactions to support setting an alias using an `evm address alias`
+- Update `CryptoCreate` transactions to support setting an alias using a `key alias`
+- Update `CryptoTransfer` transactions to support hollow-account creation for accounts with an `evm address alias`
```
rbair23, line 181: Should we support creation of contracts with an alias? We would need to update the ContractCreateTransaction.

This gives us a special wrinkle in the system. If I do a crypto transfer and create a hollow account, how do I know that should be an account vs. a contract? How do we make sure the contract isn't accidentally created as an account this way?

HIP/hip-583.md
```diff
@@ -10,220 +10,452 @@ status: Accepted
...
+5. As a Hedera user with an `Ethereum-native wallet`, I would like to receive HBAR or tokens in my account by sharing
+   only my `evm address alias`.
+6. As a Hedera user with an `Ethereum-native wallet`, I would like to send HBAR or tokens from my account to another
+   account on Hedera using a valid 20-byte address.
+7. As a Hedera user with a `Hedera-native wallet`,  I would like to transfer HBAR or tokens to another account using
+   only the recipient's `evm address alias`.
+8. As a user with an Ethereum-native wallet, I want to be able to only share my EVM public address to an exchange, and
+   that exchange to create a hollow account for me by transferring funds to the address.
+
+Specification
```
rbair23, line 172: We need a section that discusses throttles. How many hollow account creations can we do per second? Do hollow account creations count against account creation throttle bucket? When we upgrade a hollow account to a full account, does it count against the account update throttle? Or account creation throttle? This needs to be specified.

HIP/hip-583.md
```diff
@@ -10,220 +10,452 @@ status: Accepted
...
+#### Proof of Ownership
 
-Since Hedera accounts require keys to ensure security, a matching `ECDSA` `public-key` must be provided in a future signed transaction issued by the owner of the `public-address`. Due to the nature of `ECDSA` keys the `public-key` and `public-address` may be extracted from the signature of a transaction signed by the `ECDSA` private key that owns the new account. This process will complete the creation of the account, taking it from a `Hollow Account` to a normal complete account with a valid `public-key`.
-Note, in this case the final value of the account alias will remain as the `public-address` and not the `public-key` as it would be in the `auto-create` case.
+If the `alias` is used during `CryptoTransfer` as the _target_ of the transfer, then there is no need to prove that
+anyone controls that account. If an `alias` is used in the `AccountID` of the payer, then ownership must be asserted,
+just as is done for any other transaction today.
```
rbair23, line 300: Add some text to indicate proof of ownership is required if you are the source of the transfer, not just the payer.

HIP/hip-583.md
```diff
@@ -10,220 +10,452 @@ status: Accepted
...
+2. Otherwise, we create a `Hollow account` and set the `alias` to be the `evm address alias`
+
+Since we reserve the `evm address alias` and since an `ECDSA_secp256k1` `key alias` checks for collisions with the
+`evm address alias`, we effectively reserve the `key alias` with the corresponding `ECDSA_secp256k1` key slot as well.
+This prevents collisions between compatible `key alias` and `evm address alias`.
+
+Backwards Compatibility
+------------------------------------------------------------------------------------------------------------------------
+
+Existing accounts are unaffected by the expansion and the additional alias field on the transaction protobuf is already
```
rbair23, line 348:
```suggestion
Existing accounts are unaffected by the expansion and the alias field on the transaction protobuf is already
```

HIP/hip-583.md
```diff
@@ -10,220 +10,452 @@ status: Accepted
...
+supported.
 
 `CryptoTransfer` auto-account creation is un-impacted and will still support the provision of the public key.
```
rbair23, line 351:
```suggestion
`CryptoTransfer` auto-account creation is not impacted and will still support the provision of the public key. 
```

HIP/hip-583.md
```diff
@@ -10,220 +10,452 @@ status: Accepted
...
 
-### Mirror Node Impications
+### Mirror Node Implications
 
-The Mirror Node can already ingest account creation transactions with alias values.
-This will only increate the applicable parent trnasactions that can cause this flow.
-Where it was limited to CryptoTransfer it will now include CryptoCreate
+The Mirror Node can already ingest account creation transactions with alias values. This will only increase the
+applicable parent transactions that can cause this flow. Where it was limited to CryptoTransfer it will now include
+CryptoCreate, and in both cases it will now support `evm address` formatted aliases.
```
rbair23, line 357: We need to verify that the mirror node has the ability to find info by evm address alias in all cases (I think it does, but we need to know).

There is some additional complexity if you are processing on raw transactions. Suppose you are processing all transactions, and filtering them if they apply to the accounts you care about. Some transactions may be based on `shard.realm.num`, and some may be based on alias. Such programs have to know what real account number is associated with each alias.

Also, the mirror node needs to be prepared for hollow accounts. If a thousand transactions have been made by sending value to a hollow account, and then it is made "real", then queries on the mirror node using the `shard.realm.num` must return all the previous transactions as well.

rbair23: It would be good to have an API on mirror node where, if I give any one of my alias forms, it will return all alias forms that are supported. For example, suppose I have account `0.0.10000` and it has a `key alias` that is an ECDSA key. If I query that mirror node API, I should see `0.0.10000`, the long-zero alias form, the `key alias` form, and the `evm address alias` form associated with that `key alias`! This can be a special query.

HIP/hip-583.md
MiroslavGatsanoga: Are we debating to drop almost all CryptoCreate changes with the exception of CryptoCreate with EVM address and provided key?

Nana-EC: This was explicitly requested by wallets as they value the ability to create zero balance accounts. They can do today via public key but want to extend business practice to include evm address

lukelee-sl:
I've updated the pr with information from @stoqnkpL from the two sync meeting that we've had.  I need to make another pass to remove the option of creating accounts with `key-alias` using `CryptoCreate` once I get confirmation that this use case is not use/important to native wallet creators.

HIP/hip-583.md
```diff
@@ -10,220 +10,518 @@ status: Accepted
...
+Add the `alias` property to the `CryptoCreateTransactionBody` [proto file](https://github.com/hashgraph/hedera-protobufs/blob/main/services/crypto_create.proto)
 
-Post the implementation of [HIP 32](https://hips.hedera.com/hip/hip-32), only accounts created via `auto-create` using the `CryptoTransfer` transaction have aliases set. Wallets and exchanges often utilize the `CryptoCreate` transaction and therefore have not fully integrated alias logic into their UX flows. This limits full applicability of Smart Contract scenarios for Hedera accounts created using `Hedera-native wallets` and web3 tool compatibility for new users coming from external chains using `Ethereum-native wallets`. 
+```proto
+message CryptoCreateTransactionBody {
+	...
+	/**
+     * The bytes to be used as the account's alias. The bytes must be formatted in one of the following ways:
+     * 
+     * 1. It will be the serialization of a protobuf Key message for an ED25519/ECDSA primitive key type</li>
```
MiroslavGatsanoga, line 193: This comment should be updated to not include the case for key alias (should be done in the actual protobufs as well).

HIP/hip-583.md
```diff
@@ -10,220 +10,518 @@ status: Accepted
...
-   * At most only one account can ever have a given alias on the network.
-   *
-   * If a transaction creates an account using an alias, any further crypto transfers to that alias will simply be deposited
-   * in that account, without creating anything, and with no creation fee being charged.
-	 * If an account with an alias is deleted the alias will be freed up for use.
-   */
-	bytes alias = 18;
-}
-```
+If we do create a new account from either `CryptoCreate` or `CryptoTransfer`, and the `alias` is an `ECDSA_secp256k1`
```
MiroslavGatsanoga, line 339: The mention of `CryptoCreate` when it comes to `ECDSA_secp256k1` should be removed.

HIP/hip-583.md
```diff
@@ -10,220 +10,518 @@ status: Accepted
...
 - Add support to check if an account exists with a given alias
 
 ### Performance Implications
 
-The mass adoption of alias setting on new and existing accounts will have minor but notable memory implications on the Hedera ecosystem
+`CryptoCreate` transactions will see extra bytes for every account created with an alias.
 
-`CryptoCreate` transactions will see extra bytes for every account created with an alias
+The size of the bytes in each case depends on the governing account key type. For an `ECDSA` based account the alias
+will always be a valid Ethereum address and will be 20 bytes long. In the `ED25519` case the alias will be a valid
```
MiroslavGatsanoga, line 434: The `ED25519` alias part can be removed since it will not be supported for `CryptoCreate`.

HIP/hip-583.md
```diff
@@ -10,220 +10,518 @@ status: Accepted
...
+Overall, we anticipate no significant performance impact from this feature.
 
-## Security Implications
+Security Implications
+------------------------------------------------------------------------------------------------------------------------
 
-The `lazy-create` account flow requires additional checks be added to ensure transactions involving an account with only a `public-address` are not processed without a matching `public-key`. 
+The `hollow-create` account flow requires additional checks be added to ensure transactions involving an account with
+only an `alias` and no key (i.e. Hollow accounts) are not processed. The `CryptoCreate` flow must verify the transaction
+is signed by the key represented by the alias (both `key alias` and `evm address alias`).
```
MiroslavGatsanoga, line 444: Here we should have only the `evm address alias` left and remove the `key alias`.

MiroslavGatsanoga (dismissed):

Nana-EC (approved):

Nana-EC (dismissed):
LG.
Please add you name to `author`, move Richards to `author` and add Stoyans to working group.

Nana-EC (approved):

lukelee-sl (approved):

PR #679, opened 2023-02-15
Bucket restructure for stream data

mgoelswirlds:
Adding the bucket restructuring HIP.

**Description**:
We propose changing the current stream data bucket folder structure from node account ID based to node ID based. Additionally, we propose incorporating shard into the path to be future compatible with sharding when it's enabled. We are also proposing a rollup mechanism to reduce cloud storage duplication and improve historical sync times of mirror nodes.

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-679.md
+++ b/HIP/hip-679.md
@@ -176,7 +176,10 @@ After a significant amount of time has passed after launching Phase three, we wi
 The existing mirror nodes would need to implement phase one.
 
 ### Dev ops Implications
-After the completion of phase one and having being notified, dev ops will need to implement phase two.
+After the completion of phase one and having been notified, dev ops will need to implement the bucket folder permissioning for specific nodes for phase two. Any related config changes would also be a dev ops responsibility, followed by the rollout of phase two.
+
+### Engineering team Implications
+Updates to the consensus node software and any code changes to mirror.py for phase two would be implemented by the engineering team.
 
 ## Backwards Compatibility
 
--- a/HIP/hip-679.md
+++ b/HIP/hip-679.md
@@ -1,5 +1,5 @@
 ---
-hip: 
+hip: 679
 title: Bucket restructuring for changing stream data folder structure
 author: Mugdha Goel (@mgoelswirlds)
 working-group: Richard Bair (@rbair23), Nick Poorman (@nickpoorman), Jasper Potts (@jasperpotts), Dan Alvizu (@dalvizu), Steven Sheehy (@steven-sheehy)
--- a/HIP/hip-679.md
+++ b/HIP/hip-679.md
@@ -9,7 +9,7 @@ needs-council-approval: Yes
 status: Draft
 created: 2023-01-09
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/678
-updated: 2023-02-21
+updated: 2023-03-15
 ---
 
 ## Abstract
--- a/HIP/hip-679.md
+++ b/HIP/hip-679.md
@@ -6,7 +6,7 @@ working-group: Richard Bair (@rbair23), Nick Poorman (@nickpoorman), Jasper Pott
 type: Standards Track
 category: Core
 needs-council-approval: Yes
-status: Draft
+status: Review
 created: 2023-01-09
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/678
 updated: 2023-03-15
--- a/HIP/hip-679.md
+++ b/HIP/hip-679.md
@@ -7,12 +7,9 @@ type: Standards Track
 category: Mirror, Service
 needs-council-approval: Yes
 status: Draft
-last-call-date-time: 
 created: 2023-01-09
-discussions-to: 
-requires: 
-replaces: 
-updated: 
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/678
+updated: 2023-02-15
 ---
 
 ## Abstract
@@ -209,7 +206,9 @@ Use the ledger id instead of the network name in the bucket path. This idea was 
 We do not know of any open concerns with this proposal.
 
 ## References
-This is the link to the discussion https://github.com/hashgraph/hedera-improvement-proposal/discussions/678 for this HIP.
+
+* [Mirror Node GitHub](https://github.com/hashgraph/hedera-mirror-node)
+* [Zstd gz comparison](https://engineering.fb.com/2016/08/31/core-data/smaller-and-faster-data-compression-with-zstandard)
 
 ## Copyright/license
 
--- a/HIP/hip-679.md
+++ b/HIP/hip-679.md
@@ -9,7 +9,7 @@ needs-council-approval: Yes
 status: Draft
 created: 2023-01-09
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/678
-updated: 2023-02-15
+updated: 2023-02-21
 ---
 
 ## Abstract
--- a/HIP/hip-679.md
+++ b/HIP/hip-679.md
@@ -4,7 +4,7 @@ title: Bucket restructuring for changing stream data folder structure
 author: Mugdha Goel (@mgoelswirlds)
 working-group: Richard Bair (@rbair23), Nick Poorman (@nickpoorman), Jasper Potts (@jasperpotts), Dan Alvizu (@dalvizu), Steven Sheehy (@steven-sheehy)
 type: Standards Track
-category: Mirror, Service
+category: Core
 needs-council-approval: Yes
 status: Draft
 created: 2023-01-09
--- a/HIP/hip-641.md
+++ b/HIP/hip-641.md
@@ -1,0 +1,216 @@
+---
+hip: 
+title: Bucket restructuring for changing stream data folder structure
+author: Mugdha Goel (@mgoelswirlds)
+working-group: Richard Bair (@rbair23), Nick Poorman (@nickpoorman), Jasper Potts (@jasperpotts), Dan Alvizu (@dalvizu), Steven Sheehy (@steven-sheehy)
+type: Standards Track
+category: Mirror, Service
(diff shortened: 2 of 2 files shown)


HIP/hip-679.md
```diff
@@ -0,0 +1,215 @@
...
+    
+     ```
+     The above compressed tar file includes the following files in node id based folders:
+     - Record stream files (.rcd.zst)
+     - Signature files (.rcd_sig)
+     - Sidecar files (_01.rcd.zst)
+     
+     NOTE: File `2022-07-13T08_46_08.041986003Z.rcd.zst` appears in two separate node folders. These are files with distinct content and when downloaded by the mirror node one of the two file will be selected based on the consensus majority.
+     
+     The aggregation in the new files would be size/time based, as will be determined by performance testing.
```
jasperpotts, line 169: Are we going to try multi levels of roll up? Hour/Day/Week/Month/Year etc?

HIP/hip-679.md
```diff
@@ -0,0 +1,215 @@
...
+    
+Example:
+    accountBalances/balance0.0.3/
+    eventsStreams/events_0.0.3/
+    recordstreams/record0.0.3/
+```
+    
+Proposed Structure:
+```
+    {network}/{shard}/{nodeID}/balance/
```
tannerjfco, line 63: there are 2 different identifiers for nodes, which do we intend to use here?

1. there is an internal platform identifier that starts from 0, which is how nodes are identified in logs
2. there is a node "name" that is defined in config.txt as well as in the mTLS key for the node (this is the relevant [config.txt field description](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/config.txt#L104) and this name would be "Alice" in the [example configuration](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/config.txt#L26)

we typically use #1 to talk about nodes. we're also interested in updating #2 to match #1, but this isn't done yet. if we intended to use #2 here, such an update should become prerequisite to enabling use of this new structure

HIP/hip-679.md
```diff
@@ -0,0 +1,215 @@
...
+    {network}/{shard}/{nodeID}/event/
+    {network}/{shard}/{nodeID}/record/
+    
+Example:
+
+    Standard network:
+    mainnet/0/3/balance/
```
tannerjfco, line 70: IMO, we should consider including the label of these numbers so that it's clear just from looking at the folder names what they identify. otherwise, you would need this HIP or other documentation to understand what they describe

e.g. `mainnet/shard0/node3/balance/`

HIP/hip-679.md
```diff
@@ -0,0 +1,215 @@
...
+    mainnet/0/3/event/
+    mainnet/0/3/record/
```
tannerjfco, line 72: last question on these ids - do we need to consider padding w/ zeros to allow for sorting as expected?

e.g. without padding you could get a sort like

1
10
11
2
3
4
...etc

vs w/ padding

01
02
03
04
05
06
07
08
09
10

a big catch though is that you have to know how many digits you'll need. this can make it easier for humans to navigate the buckets, but we don't have any such padding for the current account id delineation so we already have the former today, and I'm not aware that it's been a significant issue

HIP/hip-679.md
```diff
@@ -0,0 +1,215 @@
...
+ - If pathType is `AUTO`:
+    - Attempt to list signatures from  a path with node account IDs
+    - If successful, update a per node timestamp and boolean to ensure account ID is used until pathRefreshInterval elapses
+    - If no results, attempt to list signatures from a path with node IDs
+    - If successful, set `pathType` as `NODE_ID` so the node ID path is used going forward
+   
+In the existing scenario the default pathType would be `ACCOUNT_ID`, which would then be changed to AUTO when we are close to approaching Phase two.
+
+#### Phase Two:
+Update the consensus nodes to write stream files to the new `{network}/{shard}/{nodeID}/*` nodeId based path.
```
tannerjfco, line 111: I believe this may require updates to mirror.py, as I don't believe it currently recurses directories. it may be possible to account for in the deployment and target mirror.py to watch specifically the `{nodeID}` folder, provided there are no further sub-directories after this point

HIP/hip-679.md
```diff
@@ -0,0 +1,215 @@
...
+- Once the limit has been reached, aggregate the file and write it to the new bucket location. 
+- Delete the existing files that have been read from the current bucket location.
+
+After a significant amount of time has passed after launching Phase three, we will remove the properties in mirror node to default the downloading from the node ID based path.
+
+### Mirror Node Implications
+The existing mirror nodes would need to implement phase one.
+
+### Dev ops Implications
+After the completion of phase one and having being notified, dev ops will need to implement phase two.
```
tannerjfco, line 179: DevOps will be responsible for executing the rollout of this phase, but I don't believe that is the right assignment for overall implementation. This change will require updates to consensus node software, which would be implemented by services team. This could also require updates to mirror.py, which would be implemented by engineering as well (currently RE). If there are changes required to the deployment manifest (e.g. docker-compose changes) this part would currently go to DevOps to implement

HIP/hip-679.md
```diff
@@ -0,0 +1,215 @@
...
+    accountBalances/balance0.0.3/
+    eventsStreams/events_0.0.3/
+    recordstreams/record0.0.3/
+```
+    
+Proposed Structure:
+```
+    {network}/{shard}/{nodeID}/balance/
+    {network}/{shard}/{nodeID}/event/
+    {network}/{shard}/{nodeID}/record/
```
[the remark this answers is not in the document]
mgoelswirlds: We considered adding the realm also initially. But as per Leemon, ```
we could ignore realm , as if it is ever implemented it would be invisible to streaming files.
For a round, there will be one state file per shard, and it will contain all the realms in that shard. Similarly, a record file or event file or balances file will contain info for all the realms in that shard.
So the only important factors are the ledger ID (e.g., mainnet vs public testnet), and the shard (always zero at the moment). 
```

HIP/hip-679.md
mgoelswirlds: Just trying to understanding how that would be helpful? Currently the idea is to not do that , but as you see the rollup name is to the minute.

HIP/hip-679.md
mgoelswirlds: We intend to use the node identifier from 1 which as you rightly pointed is the internal platform identifier to keep it consistent.

steven-sheehy: > node identifier from 1

HIP/hip-679.md
steven-sheehy: There should not be any consensus nodes changes unless we decide to have consensus nodes compress using zstd instead of gzip as Jasper has suggested. There will be mirror.py changes but that has only recently changed responsibilities from DevOps back to engineering in the past week or so, as I understand it. So we can update the this section to reflect that.

HIP/hip-679.md
steven-sheehy: I don't think there's any need for recursion. The watch folder name and logic can stay exactly as is. It's only the S3 destination name that is changing, which should hopefully just be a config tweak.

HIP/hip-679.md
steven-sheehy: It's not necessary, we list for signatures within a particular node now not across. Padding would make it easier to visually look at the buckets, but would not help programmatic access by the mirror nodes. I don't think we should optimize for anything except programmatic access so we should keep it as is without padding like how it is in production.

HIP/hip-679.md
steven-sheehy: Disagree. We should not be optimizing for manual interaction with the bucket structure. Adding such prefixes helps visually but only increases the complexity of the mirror node code to have parse out the prefix. And it's not like we're adding a prefix for the other variable parts of the path like network name and stream type: `networkMainnet/shard0/node3/streamBalance`.

If you think of it as a REST API, we don't prefix variables when invoking urls like `/api/v1/accounts/account0.0.3` as it's implied from the specification `/api/v1/accounts/{id}` that the last URL component is the account ID.

HIP/hip-679.md
steven-sheehy: Though, depending upon whether it's a config or code change to mirror.py it might be either DevOps or engineering.

tannerjfco: I'm not sure what the update needs to be, but this statement still needs to be addressed

jasperpotts: My hope is longer term we will upload directly from services app rather than depending on external script. We can compress the stream as we send it and it can never touch the disk.

I did some experiments with zstd it can offer a nice increase in compression. It takes a little more CPU and a bit longer compared to Gzip to get there. If we build a dictionary that is reused for every file then it is less CPU and faster than Gzip and more compression. Just a bit more complex to code and manage. 

It would be nice to go from GZ to ZStd but not critical in short term

HIP/hip-679.md
jasperpotts: If we are running an importer that is trying to catch up, I found I was performance limited by number of API calls per second to storage bucket. There is one per file, so less files = faster. My thought that having a roll up of a complete days data in a single compressed file would super speed up bulk import.

HIP/hip-679.md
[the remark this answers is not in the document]
jasperpotts: Ok makes sense.

HIP/hip-679.md
mgoelswirlds: As mentioned in the HIP, as part of engineering design and with the help of performance testing , we still need to determine what the optimum size of the rollup would be.

HIP/hip-679.md
mgoelswirlds: Can this be resolved as of now?

HIP/hip-679.md
mgoelswirlds: yes that is accurate. Above was a typo.

mgarbs (approved):

PR #680, opened 2023-02-16
Feature/supply key adapt metadata of nft

HGraphPunks:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-660.md
+++ b/HIP/hip-660.md
@@ -1,0 +1,63 @@
+---
+hip: 660
+title: Enable supply key to adapt metadata of an NFT in the treasury wallet
+author: Patches (@HGraphPunks)
+working-group: Adrian Marin (@AdriMsM91), Ashe Oro (@Ashe-Oro), Michael Garber (@mgarbs)
+type: Standards Track
+category: Core
+needs-council-approval: Yes
+status: Draft
+last-call-date-time: null
+created: 2023-02-16
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/660
+updated: 2023-02-16
+---
+
+## Abstract
+
+Currently, there is no native way to update the metadata of a non-fungible token (NFT) on the Hedera network. This HIP proposes the extension of the SUPPLY key that will allow updating NFT metadata of NFTs in the treasury wallet. This will provide the Hedera network with enhanced abilities to create dynamic NFTs that are gaurenteed immutable outside of the treasury wallet, and mutable inside the treausry wallet. 
+
+## Motivation
+
+Though other methods can be used like burning and reminting a NFT, this is suboptimal for a lot of use cases where serial number corresponds to data / imagery and loses the ability to see adaptations of serial number over time. 
+
+There should be a way to adapt the metadata, but users know that the NFT can't be updated when it's in their wallet / distributed on the network.
+
+
+## Rationale
+
+ This ability of adapataion would gaurentee the data is immutable (if decentralized points is used) and the value earned from the data won't change even if the creator is disgruntled or malicious. Hedera network is uniquely positioned to create a method in which an NFT can be gaurenteed immutable outside the treasury wallet and mutable while inside the treasury wallet.
+
+
+## User stories
+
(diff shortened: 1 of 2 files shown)


HGraphPunks:
@mgarbs let me know if I need to adapt anything on this PR to get it past the apporvals

Ashe-Oro:
The Engineering team has expressed quite a bit of concern over making a
once immutable NFT mutable by overloading the supply key.

HGraphPunks:
> The Engineering team has expressed quite a bit of concern over making a once immutable NFT mutable by overloading the supply key.

What would be the best venue to engage in dialogue about the opposing opinions? I haven't seen any feedback that wasn't responded to in the discussions

bugbytesinc:
While I empathize with the problem this is trying to solve (_the perceived value of an NFT’s serial number is greater than its metadata_) I do not believe this HIP should be implemented.

This changes the rules for existing NFTs and the tokenomic assumptions around them for both creators and owners.

Once the sister HIP (metadata key) is implemented, why not just offer to burn and re-mint fungible metadata NFTs instead.  You can always swap serial number for serial number?

HGraphPunks:
> While I empathize with the problem this is trying to solve (_the perceived value of an NFT’s serial number is greater than its metadata_) I do not believe this HIP should be implemented.
> 
> This changes the rules for existing NFTs and the tokenomic assumptions around them for both creators and owners.
> 
> Once the sister HIP (metadata key) is implemented, why not just offer to burn and re-mint fungible metadata NFTs instead. You can always swap serial number for serial number?

I would agree with you 100% if the only focus and benefit was to create a method for backwards compatibility for adaptation of NFTs without burning.

The functionality that Hedera is uniquely positioned to implement is to guaranteed immutability in distribution and mutable data in treasury wallet. 

I believe use cases from gaming to legal contracts, to PFP projects that would benefit from being able to guarantee no manipulation of data outside of the owner sending their NFT to the treasury wallet. It also is a step closer to secure methods of adaptation than just the boolean with the METADATA key: ALL data can be edited all the time or none of the data can be edited ever. This addition would innovate a new state that's flexibile, and has trustless protection to holders. 

**Addressing the security concerns:**
I understand changing the scope of a key in production that affects many projects that are already minted. I agree changing rules under users on a DLT should be done rarely and with extreme caution. Mitigation of malicious actions that could be performed is paramount to trust in the network. 

That's coupled with the fact that almost all token Ids have a supply key since that's required to mint does make this more of a risk.

All that said, I see the risks as **extremely low.** 

**Points of attack from a malicious actor:**
1. Trick users to send their NFT back to the treasury, then creator changes the data
If a creator is malicious, the user is in danger of not even receiving a NFT back from sending it to the treasury. If the creator wants to degrade data on the NFT, they could but it requires the user to sign a transaction first to hand over custody. 

2. NFTs already in treasury wallet are changed before sent to users who are owned them
This can be an issue if the user sees on the mainnet / explorers that this is the NFT they want. Creator could manipulate it right before, devaluing the NFT compared to what the user expected the value to be for the purchase. 

This requires the seller to be a malicious creator, to hold the NFT, and to devalue serial numbers of their own project for short term gains. The ramifications of this action would also cause a lot of uncertainty in whatever platform or creator was doing this.

My main retort to security issues:
The risk that adding update functionality to the SUPPLY key is extremely low, and unlikely to affect users in production. All risks defined above are not additions to risks holders have today with creators using HTTP and centralized servers to get functionality they want with the current state of Hedera Network. 

A HTTP uri can change any data at any time for any NFT (much like a METADATA key). Most users don't understand the storage implications and developers are already using HTTP to get the benefits of this HIP which expose holders to more risk IMHO. 

I do believe there's a lot of benefits and innovation this HIP would bring. A new dynamic of more granular adaption of NFTs and more trustless immutability to users.

mgarbs:
While metadata mutation for NFTs has potential benefits, it's important to consider the drawbacks. Some of the things that have been discussed with engineering:
Introducing this feature could undermine trust in NFTs, as users expect immutability. 
Tracking provenance and history could become more challenging. 
Potential for fraud and manipulation is introduced, even though the chance is very small, it still exists and would be introduced by this hip.

mgarbs:
Being replaced with - https://github.com/hashgraph/hedera-improvement-proposal/pull/850

PR #682, opened 2023-02-22
Fix backwards compatibility clause

michielmulders:
**Description**:
The HIP-657 mentions that it affects both old and new NFTs while it will only apply to new NFTs.


Ashe-Oro (approved):
good catch.

PR #683, opened 2023-02-23
SOC2 Proposal HIP

voycey:
**Description**:
This has been discussed at length in the Guardian Tech Call and between members of HBAR Foundation and Swirlds labs. There likely needs to be some back and forth still on this but pushing this out to the wider community for discussion.

(Apologies - there are no instructions on what to call the renamed template - can set the filename and location to whatever is required)


mgarbs (commented):
You may have dragged this over and renamed it. Can we re-add the template back to the root directory?

voycey:
I have re-added that - it could have been solved during the merge by cherry picking the updated HIP file only - probably a better way of keeping this repo clean than doing a full merge.

mgarbs:
Thank you for submitting this HIP. After careful consideration, may we table this HIP and revisit it later once we have more clarity on a new implementation of archival data functionality?

Currently, we are exploring ways to archive data instead of expiring them from the network when rent is not paid. As a result, certain reports, including the SOC2 report, may not be necessary. We are working on the implementation details, such as how the data will be resurrected from the archive, what the associated fees will be, and whether there will be temporary unarchive fees. Let us revisit this when we have worked that out. Sound good?

voycey:
Hi Michael,

Unfortunately not, we need to move ahead with this as data is able to be removed from the network (whether it's rent, GDPR right to be forgotten or even illicit content the reason isn't important), as such, our SOC2 auditors are going to be asking for attestation from Hedera as to what controls Hedera have implemented for this.

This isn't just TYMLEZ specific, this will be required for any business building on the Hedera network

mgarbs:
> data is able to be removed from the network.

This isn't the case, though. Rent was never activated on mainnet. All I'm suggesting here is that we revisit this hip later because we understand your concerns among other people and are now _archiving_ the data instead of deleting it. When it's archived it will be put into cold storage, later retrievable for a fee. Engineering is working out all the details of how that is implemented. If we look at this later, we will have a better idea of where to go.

voycey:
> _archiving_ the data instead of deleting it

Michael, the fact that deletion was/is even an option is the reason that a SOC2 from Hedera is required

voycey:
@mgarbs to add to this - the ability for access to the network to be arbitrarily restricted to participants and businesses is another reason why SOC2 attestation is required, today's experience shows that. Our mainnet deployments are currently non-functional because of an opaque decision that was made to pause the network.

voycey:
@mgarbs Has this actually been implemented in some form? I am no longer actively developing on Hedera but from a compliance standpoint this would still be required for many enterprise deployments based on the ability to pause the network - documented controls over who has this ability need to be provided for any serious deployment

PR #702, opened 2023-03-25
Update hip-406.md

mgarbs:
This PR adds clarification for staking rewards and consensus weight explanations by:

- Adding a new scenario with Carol, who stakes without claiming a reward, to better illustrate reward distribution.

- Replacing the term "ignored" with a clearer explanation that excess coins are not rewarded and don't affect consensus weight.

These updates enhance the clarity of the documentation.


pathornteng (approved):
LGTM

PR #729, opened 2023-05-05
Add HIP for contract nonce externalization

MiroslavGatsanoga:
**Description**:
In order to provide for a better developer experience for [eth_getTransactionCount](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gettransactioncount) externalize the contract accounts nonce values from Consenus Node to Mirror Node.

This HIP proposes to:
- Externalize contract accounts nonces in transaction records
- Migrate all existing contract nonce values from Consenus Node to Mirror Node
- Expose the correct contract nonce values through Mirror Node REST API


HIP/hip-000.md
```diff
@@ -0,0 +1,116 @@
...
+- Tracking nonces for EOAs for `EthereumTransactions` is not changed (currently we don't have a specific protobuf to externalize the nonce but the nonce value is passed as part of the `EthereumTransaction` body, so Mirror Node can read the value from there)
+- Tracking nonces for EOAs for `non-EthereumTransactions` is not supported because there is no use case for it in Hedera
+
+## User stories
+
+1. As a user, when I make a request to `eth_getTransactionCount` endpoint I want to be able to get correct nonce values for contracts.
+  
+## Specification
+
+No changes to Services state will be needed because currently the nonces are currently being computed correctly there. We need to externalize these changes to Mirror Node so that it can keep a correct state representation for contract accounts.
```
steven-sheehy, line 46: So services has always been tracking the contract nonce and will have an accurate existing value? So the first time a mirror node gets a contract nonce it could be some high value (2+) since that contract could've created many child contracts previously.

HIP/hip-000.md
```diff
@@ -0,0 +1,116 @@
...
+Update `ContractFunctionResult`:
+
+```protobuf
+message ContractFunctionResult {
+    ...
+
+    /**
+     * A list of updated contract account nonces containing the new nonce value for each contract account
+     */
+    repeated ContractNonceInfo contract_nonces = 14;
```
steven-sheehy, line 76: There was a reason why we can't track this on each child's function result and just have a `ContractFunctionResult.nonce`, but I don't remember what it was. Can you clarify why by updating the docs?

HIP/hip-000.md
```diff
@@ -0,0 +1,116 @@
...
+}
+```
+
+### Mirror Node Changes
+
+Mirror Node should start using the new `contract_nonces` field and update the state of the contract accounts with it.
```
steven-sheehy, line 82: We should also include the mirror node REST API changes.

steven-sheehy (changes requested):

HIP/hip-000.md
MiroslavGatsanoga: > So services has always been tracking the contract nonce and will have an accurate existing value?

Yes, it looks like we increment the nonce in [ContractCreateTransitionLogic](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCreateTransitionLogic.java#L209), [ContractCallTransitionLogic](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/txns/contract/ContractCallTransitionLogic.java#L148) and [AbstractEvmRecordingCreateOperation](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-evm/src/main/java/com/hedera/node/app/service/evm/contracts/operations/AbstractEvmRecordingCreateOperation.java#L122).

> So the first time a mirror node gets a contract nonce it could be some high value (2+) since that contract could've created many child contracts previously.

Yes, given the above, I think that should be the case.

HIP/hip-000.md
MiroslavGatsanoga: Had a discussion with @stoqnkpL about that earlier today and the main consideration regarding the `ContractFunctionResult.nonce` approach is that in `ContractFunctionResult` we have the contract that is the target but we don't have the sender (which is the one that should have it's nonce incremented).
Also in the case of revert in some of the child transactions we are not sure if there will be a record produced for them, that's why I think the `contract_nonces` should be populated in the parent record.
I will update docs with this information.

HIP/hip-000.md
MiroslavGatsanoga: I think that for `eth_getTransactionCount` the `JSON-RPC Relay` uses the `/api/v1/accounts/{idOrAliasOrEvmAddress}` which returns a response with `ethereum_nonce` field in it. I thought that  this field will just start to return correct values for contract accounts and no API changes will be needed. Can you elaborate on what REST API changes should we additionally include?

HIP/hip-729.md
```diff
@@ -0,0 +1,117 @@
...
+needs-council-approval: Yes
+status: Draft
+created: 2023-05-05
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/730
+updated: 2023-05-09
+---
+
+## Abstract
+
+The Hedera Smart Contract Service (HSCS) supports the deployment of contracts. A deployed contract should have a nonce value that depends on the number of other contracts that were created during its own creation.
```
Nana-EC, line 17: nit: something like this maybe
```suggestion
The Hedera Smart Contract Service (HSCS) supports the deployment of contracts. A deployed contract `A` should have a nonce value that reflects the number of other contracts that were created since `A`'s creation.
```

HIP/hip-729.md
```diff
@@ -0,0 +1,117 @@
...
+    1. to check whether an account exists at an address, it checks whether the nonce property stored for this address is different than 0 (it also checks the balance and code properties for the address as well, but this is out of scope for this discussion).
+    2. when deploying new contracts via a `CREATE` operation, the address for an Ethereum contract is **deterministically** computed from the address of its creator (`sender`) and how many transactions the creator has sent (`nonce`). The `sender` and `nonce` are RLP encoded and then hashed with Keccak-256.
+
+Ethereum nodes expose a JSON-RPC endpoint that can be queried for each account’s nonce value: [eth_getTransactionCount](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gettransactioncount)
+
+In Hedera there is `JSON-RPC Relay` (as described in [HIP-482](https://hips.hedera.com/hip/hip-482)) that provides the same `eth_getTransactionCount` JSON-RPC endpoint, but currently it always returns a value of 1 for contract accounts. The desired behavior would be to have a nonce value of `1 + n` where `n` is the number of times the contract created other contracts through `CREATE` or `CREATE2` operations.
```
Nana-EC, line 30: Updated to highlight where the data is wrong and the chain effect.
```suggestion
In Hedera the Mirror Node which stores historical entity and transaction information has no knowledge of contract nonces and always returns a 0 value. As a result, the `JSON-RPC Relay` (as described in [HIP-482](https://hips.hedera.com/hip/hip-482)) which exposes the `eth_getTransactionCount` JSON-RPC endpoint has to query the Consensus node and not the Mirror Node to return nonce values for contract accounts. The desired behavior would be to have a nonce value of `1 + n` where `n` is the number of times the contract created other contracts through `CREATE` or `CREATE2` operations exposed to the Mirror Node.
```

HIP/hip-729.md
```diff
@@ -0,0 +1,117 @@
...
+In general, when we talk about account nonces we will need to consider all of those transaction types for both types of accounts (i.e. EOAs and contracts):
+
+- Contract nonce updates for both cases of `EthereumTransactions` and `non-EthereumTransactions` (i.e. `ContractCall` or `ContractCreate`) will need to be tracked and externalized to Mirror Node
+- Tracking nonces for EOAs for `EthereumTransactions` is not changed (currently we don't have a specific protobuf to externalize the nonce but the nonce value is passed as part of the `EthereumTransaction` body, so Mirror Node can read the value from there)
+- Tracking nonces for EOAs for `non-EthereumTransactions` is not supported because there is no use case for it in Hedera
+
+## User stories
+
+1. As a user, when I make a request to `eth_getTransactionCount` endpoint I want to be able to get correct nonce values for contracts.
```
Nana-EC, line 43:
```suggestion
1. As a user, when I make a request to the Mirror Node `api/v1/contracts/{contractId}` I observe a valid non-zero contract nonce value.
2. As a user, when I make a request to `eth_getTransactionCount` endpoint I want to be able to get correct nonce values for contracts.
```

HIP/hip-729.md
```diff
@@ -0,0 +1,117 @@
...
+     * A list of updated contract account nonces containing the new nonce value for each contract account
+     */
+    repeated ContractNonceInfo contract_nonces = 14;
+}
+```
+
+### Mirror Node Changes
+
+Mirror Node should start using the new `contract_nonces` field and update the state of the contract accounts with it.
+
```
Nana-EC, line 84:
```json
{
  ...,
  "nonce": 1,
  ...
}
```

HIP/hip-729.md
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * A list of updated contract account nonces containing the new nonce value for each contract account
+     */
+    repeated ContractNonceInfo contract_nonces = 14;
+}
+```
+
+### Mirror Node Changes
+
+Mirror Node should start using the new `contract_nonces` field and update the state of the contract accounts with it.
```
Nana-EC, line 83: I believe Mirror Node already has the field in the db `entity` table, it just doesn't update it or expose it for contracts
```suggestion
Mirror Node should start using the new `contract_nonces` field and update the state of the contract accounts with it. Additionally, the `api/v1/contract/{contractIdOrAddress}` endpoint response should be updated to return the contract `nonce` value
```

Nana-EC (commented):
Some suggestions to expand details

HIP/hip-000.md
MiroslavGatsanoga: Added to `Rejected Ideas` section.

HIP/hip-729.md
```diff
@@ -0,0 +1,120 @@
...
+The Hedera Smart Contract Service (HSCS) supports the deployment of contracts. A deployed contract should have a nonce value that depends on the number of other contracts that were created during its own creation.
+Currently this is not the observed behavior in Hedera and this HIP aims to fix that.
+
+## Motivation
+
+In Ethereum, contract accounts have nonces, as specified by the [yellow paper](http://gavwood.com/paper.pdf) and [EIP-161](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-161.md).
+
+The EVM uses the nonce value internally, for several different operations, for example:
+    1. to check whether an account exists at an address, it checks whether the nonce property stored for this address is different than 0 (it also checks the balance and code properties for the address as well, but this is out of scope for this discussion).
+    2. when deploying new contracts via a `CREATE` operation, the address for an Ethereum contract is **deterministically** computed from the address of its creator (`sender`) and how many transactions the creator has sent (`nonce`). The `sender` and `nonce` are RLP encoded and then hashed with Keccak-256.
```
bguiz, line 26: consider adding a 3rd bullet point:

```
    3. when deploying new contracts via a `CREATE2` operation, ...
```

HIP/hip-729.md
```diff
@@ -0,0 +1,120 @@
...
+
+Ethereum nodes expose a JSON-RPC endpoint that can be queried for each account’s nonce value: [eth_getTransactionCount](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gettransactioncount)
```
bguiz, line 28:
```suggestion
Ethereum nodes expose a JSON-RPC endpoint that can be queried for each account’s nonce value: [`eth_getTransactionCount`](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gettransactioncount)
```

HIP/hip-729.md
```diff
@@ -0,0 +1,120 @@
...
+
+In Hedera there is `JSON-RPC Relay` (as described in [HIP-482](https://hips.hedera.com/hip/hip-482)) that provides the same `eth_getTransactionCount` JSON-RPC endpoint, but currently it always returns a value of 1 for contract accounts. The desired behavior would be to have a nonce value of `1 + n` where `n` is the number of times the contract created other contracts through `CREATE` or `CREATE2` operations.
+
+## Rationale
+
+As described in [HIP-482](https://hips.hedera.com/hip/hip-482), the `JSON-RPC Relay` works with `EthereumTransactions` that it forwards to the consensus nodes. Apart from the `EthereumTransaction` type the consensus nodes can also handle other types of transactions (e.g. `ContractCall` or `ContractCreate`) that are not coming from the `JSON-RPC Relay` but can be used by developers by other means.
```
bguiz, line 34:
```suggestion
As described in [HIP-482](https://hips.hedera.com/hip/hip-482), the `JSON-RPC Relay` works with `EthereumTransactions` that it forwards to the consensus nodes. Apart from the `EthereumTransaction` type, the consensus nodes can also handle other types of transactions (e.g. `ContractCall` or `ContractCreate`) that are do not come from the `JSON-RPC Relay` but can be used by developers by other means.
```

HIP/hip-729.md
MiroslavGatsanoga: Added.

HIP/hip-000.md
MiroslavGatsanoga: Added.

HIP/hip-729.md
```diff
@@ -0,0 +1,117 @@
+---
+hip: 729
+title: Contract Accounts Nonces Externalization
```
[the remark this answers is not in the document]
MiroslavGatsanoga: Updated.

HIP/hip-729.md
MiroslavGatsanoga: Updated.

HIP/hip-729.md
MiroslavGatsanoga: Updated.

HIP/hip-729.md
MiroslavGatsanoga: Updated.

HIP/hip-729.md
MiroslavGatsanoga: Updated.

HIP/hip-729.md
MiroslavGatsanoga: Updated.

HIP/hip-729.md
MiroslavGatsanoga: The section starts with _"The EVM uses the nonce value internally, for several different operations, for example:"_, AFAIK `CREATE2` does not use the nonce, right?
Do you mean that we should state here that nonce is not used in `CREATE2` e.g. something like:
```
3. when deploying new contracts via a `CREATE2` operation, the derivation of new contract addresses does not involve the nonce. Instead, a contract address can be derived using the deployer address, bytecode, and salt.
```

bguiz: Yeah exactly, just being explicit to say this scenario is not included in the use of `nonce`,
because later on we have this text:

```
The desired behavior would be to have a nonce value of `1 + n` where `n` is the number of times the contract created other contracts through `CREATE` or `CREATE2` operations exposed to the Mirror Node.
```

but I assume that this refers to nonce as an *output* of the operation,
and not and *input* into the operation.

steven-sheehy (dismissed):

HIP/hip-729.md
MiroslavGatsanoga: Added.

HIP/hip-729.md
```diff
@@ -0,0 +1,129 @@
...
+Currently this is not the observed behavior in Hedera and this HIP aims to fix that.
+
+## Motivation
+
+In Ethereum, contract accounts have nonces, as specified by the [yellow paper](http://gavwood.com/paper.pdf) and [EIP-161](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-161.md).
+
+The EVM uses the nonce value internally, for several different operations, for example:
+    1. to check whether an account exists at an address, it checks whether the nonce property stored for this address is different than 0 (it also checks the balance and code properties for the address as well, but this is out of scope for this discussion).
+    2. when deploying new contracts via a `CREATE` operation, the address for an Ethereum contract is **deterministically** computed from the address of its creator (`sender`) and how many transactions the creator has sent (`nonce`). The `sender` and `nonce` are RLP encoded and then hashed with Keccak-256.
+    3. when deploying new contracts via a `CREATE2` operation, the derivation of new contract addresses does not involve the nonce. Instead, a contract address can be derived using the deployer address, bytecode, and salt.
```
Nana-EC, line 27: nit: having it as a 3rd point makes it read as if the nonce is utilized in `CREATE2`.
The content is good but i'd make a separate sentence right after number 2 instead of number 3

HIP/hip-729.md
```diff
@@ -0,0 +1,129 @@
...
+
+Ethereum nodes expose a JSON-RPC endpoint that can be queried for each account’s nonce value: [`eth_getTransactionCount`](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gettransactioncount)
+
+In Hedera the Mirror Node which stores historical entity and transaction information has no knowledge of contract nonces and always returns a 0 value. As a result, the `JSON-RPC Relay` (as described in [HIP-482](https://hips.hedera.com/hip/hip-482)) which exposes the `eth_getTransactionCount` JSON-RPC endpoint has to query the Consensus node and not the Mirror Node to return nonce values for contract accounts. The desired behavior would be to have a nonce value of `1 + n` where `n` is the number of times the contract created other contracts through `CREATE` or `CREATE2` operations exposed to the Mirror Node.
```
Nana-EC, line 31:
```suggestion
In Hedera the Mirror Node which stores historical entity and transaction information has no knowledge of contract nonces and always returns a 0 value. As a result, the `JSON-RPC Relay` (as described in [HIP-482](https://hips.hedera.com/hip/hip-482)) which exposes the `eth_getTransactionCount` JSON-RPC endpoint has to query the Consensus node and not the Mirror Node to return nonce values for contract accounts. Querying the consensus node comes at a cost which differs from EVM use cases. Tthe desired behavior would be to have a nonce value of `1 + n` where `n` is the number of times the contract created other contracts through `CREATE` or `CREATE2` operations exposed to the Mirror Node and clients that consume the Mirror Node APIs such as the JSON RPC relay.
```

HIP/hip-729.md
```diff
@@ -0,0 +1,129 @@
...
+
+Mirror Node should start using the new `contract_nonces` field and update the state of the contract accounts with it.  Additionally, the `/api/v1/contracts` and `api/v1/contract/{contractIdOrAddress}` endpoints responses should be updated to return the contract `nonce` value:
+```json
+{
+  ...,
+  "nonce": 1,
+  ...
+}
+```
+
```
Nana-EC, line 93: Please incorporate @SimiHunjan 's suggestion to add a section for `SDK` and `Mirror Node Explorer` changes

Nana-EC (changes requested):
Wording suggestions and highlight of suggested missing sections.
Specification details look good

HIP/hip-729.md
MiroslavGatsanoga: Updated.

HIP/hip-729.md
MiroslavGatsanoga: Updated.

Nana-EC (approved):

mgarbs (approved):

PR #733, opened 2023-05-08
add tooltips to all page and individual hips

mgarbs:
image

This PR adds tooltips on the page displaying all hips and gives information on what the individual statuses mean. It also adds tooltips to the individual hips page to give users more information about what some of the headers mean that users have struggled with


pathornteng (approved):
I would suggest moving Javascript and CSS to separate files for a better project structure. But it's a minor comment. LGTM

PR #739, opened 2023-05-18
Include Latest Consensus Timestamp in HAPI Responses

nickpoorman:
Adds a HIP for including the latest consensus timestamp in HAPI responses.


nickpoorman:
After some discussion, we've determined the latest consensus timestamp would best be queried from the Mirror Node. Closing this HIP in favor of that solution.

PR #742, opened 2023-05-24
Add HIP for share scheduleId code (#731)

franfernandez20:
**Description**:
Provide a solution to share a scheduleId

### DIFF
--- a/HIP/hip-731.md
+++ b/HIP/hip-731.md
@@ -1,0 +1,59 @@
+---
+hip: 731
+title: Standarize the way to share scheduleId transactions
+author: Fran Fernandez <fran@kabila.app>, Adrian Marin <adrian@kabila.app>
+working-group: Fran Fernandez <fran@kabila.app>, Adrian Marin <adrian@kabila.app>, Rafael Lujan <rafa@kabila.app>
+type: Standards Track
+category: Application
+needs-council-approval: No
+status: Draft
+created: 2023-05-24
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/731
+---
+
+## Abstract
+
+This proposal aims to standardize the way of sharing the codes of a Schedule Transaction.
+
+## Motivation
+
+Right now if someone tries to share the id of a Schedule Transactions, for example from one dApp to different wallets, it is not supported as they are using different algorithms to obfuscate the generated code. We believe that this is a high barrier for users.
+
+## Rationale
+
+This is a new specification, so should not have alternative design.
+
+## User stories
+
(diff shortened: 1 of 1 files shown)


HIP/hip-731.md
mgarbs, line 1: Looks great! Do we have a community consensus here? Let's invite @rocketmay and others who should be looped into this hip discussion

AdriMsM91: I don't think we have a community consensus, Tyler CTO from HashPack answered in the HIP. I'll ping him also in Discord.

AdriMsM91: Hello @mgarbs any update?

mgarbs: Hi @AdriMsM91! Are we all good on community consensus? Technically since its an informational hip we don't need it and devs can choose to follow it or not but I do want to put it out there one last time. Might be good to get @rocketmay 's input too here.

AdriMsM91: For sure

teacoat: This looks good to me

AdrianKBL: I think we're good to go @mgarbs !

jrwingate: I agree with BugBytesInc - whats the value in this - can someone explain the critical path for why this is needed?

HIP/hip-731.md
```diff
@@ -0,0 +1,59 @@
...
+
+Right now if someone tries to share the id of a Schedule Transactions, for example from one dApp to different wallets, it is not supported as they are using different algorithms to obfuscate the generated code. We believe that this is a high barrier for users.
+
+## Rationale
+
+This is a new specification, so should not have alternative design.
+
+## User stories
+
+1. As an user that wants to sell nfts, I want to trade some nfts for hbar, so I want to generate a schedule transaction with nfts serial numbers and how much the receiver will pay. I want to get my code and share it with the buyer so he can execute it.
```
mgarbs, line 28: When you say code are you talking about the memo?

HIP/hip-731.md
AdrianKBL: Ey thanks for your inputs @Volind @bugbytesinc @jrwingate 

Dropping here the discussion we had and why we need it (Last post).

https://github.com/hashgraph/hedera-improvement-proposal/discussions/731

So today after many hours of investigation we discovered that from the mirror node we can retrieve the data, we didn't know this before. Also this isn't easy and I'm pretty sure somebody new to Hedera won't get how to do this.

So this HIP is not needed anymore, but one function from the SDK to actually can decode the transaction_body from the https://testnet.mirrornode.hedera.com/api/v1/schedules/0.0.14847288 would be interesting at least.

HIP/hip-731.md
franfernandez20: No, for code we mean an identifier of the scheduled transaction.

PR #751, opened 2023-06-08
Add initial version of HTS locking HIP

sam-at-toko:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-751.md
+++ b/HIP/hip-751.md
@@ -8,7 +8,7 @@ category: Service
 needs-council-approval: Yes
 status: Draft
 created: 2023-06-07
-updated: 2023-07-14
+updated: 2023-07-19
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/759
 ---
 
--- a/HIP/hip-751.md
+++ b/HIP/hip-751.md
@@ -8,8 +8,8 @@ category: Service
 needs-council-approval: Yes
 status: Draft
 created: 2023-06-07
-discussions-to:
-updated: 2023-06-26
+updated: 2023-07-14
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/759
 ---
 
 ## Abstract
--- a/HIP/hip-751.md
+++ b/HIP/hip-751.md
@@ -1,13 +1,15 @@
 ---
 hip: 751
 **title: Lockable Fractional Amounts of Fungible Tokens on Hedera**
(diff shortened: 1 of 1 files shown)


sam-at-luther:
fyi @atul-hedera

web3-nomad:
another way of achieving this is implemented in the cosmos SDK where they have created a 'vesting' account.. so the logic would be at the account level instead of the token one.

https://docs.cosmos.network/v0.45/modules/auth/05_vesting.html

sam-at-luther:
I've added a discussion [here](https://github.com/hashgraph/hedera-improvement-proposal/discussions/759).

sam-at-luther:
> another way of achieving this is implemented in the cosmos SDK where they have created a 'vesting' account.. so the logic would be at the account level instead of the token one.
> 
> https://docs.cosmos.network/v0.45/modules/auth/05_vesting.html

This iss along the lines of the “VestingWallet” discussed in the "Feature Support in other Languages" section.

eleonora-Toko (approved):
Initial version of HIP proposing a lockable functionality on amounts of fungible tokens 🎉.

HIP/hip-751.md
```diff
@@ -0,0 +1,337 @@
...
+
+I WANT to have the option to enable the transfer lock function for a portion of my fungible token supply with a predetermined execution time,
+
+SO THAT I can schedule the unlocking of tokens in advance without the need for immediate manual execution.
+
+**_Scenario_** -  _Token Issuer_: Executing native events on the locked portion of a token supply
+
+AS A _Token Issuer_,
+
+I WANT to be able to execute events (metadata update, mint, burn, transfers and more) on my token even when the lock functionality is enabled,
```
mhess-swl, line 143: For clarification, it sounds like a token _issuer_ can still transfer locked tokens, but token _holders_ can't exchange these locked tokens. Did I get that right?

eleonora-Toko: @mhess-swl - Yes, that is correct. The owner of the LockKey can execute events on the tokens even while they are in a locked state, including transfers.  The lock primitive, as we envision it, targets a specific account, the account that should be the token holder will not be able to successfully execute the transfer till the lock has expired.

mhess-swl: Understood, thank you!

Stephanie-Yi:
Replaced by https://github.com/hashgraph/hedera-improvement-proposal/pull/796.

mgarbs:
This hip was rewritten into 796 and since it was never merged we dont need it

PR #758, opened 2023-06-16
Unify ContractInfo and AccountInfo Queries

Nana-EC:
**Description**:
This proposal addresses the disparity between `ContractInfo` and `AccountInfo` query results.
As contracts are also accounts developers should be able to retrieve the subset of account information when querying a node, currently this is not the case.
To address this the `ContractInfo` query will be updated to return the missing account related information.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-0000-contractinfo.md
+++ b/HIP/hip-0000-contractinfo.md
@@ -1,0 +1,154 @@
+---
+hip: 
+title: Unify ContractInfo and AccountInfo Queries
+author: Nana Essilfie-Conduah (@nana-ec)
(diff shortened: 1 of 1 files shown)


mgarbs:
@Nana-EC what do we want to do with this? May I close it?

Nana-EC:
> @Nana-EC what do we want to do with this? May I close it?

@mgarbs no, we should revisit and see if could use an update.
I wrote this a while back the HIP process has advanced so might be worth freshening it up and re-socializing before deciding.
I don't think the need has changed but maybe timing

PR #765, opened 2023-07-11
NFT Collection Metadata Field

Ashe-Oro:
**Description**:

New HIP to add a dedicated metadata field for NFT Collections.


mgarbs (approved):

michielmulders:
> **should ADMIN key be able to update this metadata field, or should it require the metadataKey (ie HIP-657 - https://github.com/hashgraph/hedera-improvement-proposal/blob/main/HIP/hip-657.md)

Can we say that it should require the metadata key to stay consistent, and don't complicate the "admin power"?

PR #789, opened 2023-08-14
Remove all updated dates except latest

mgarbs:
This pull request focuses on enhancing the data management within the system, specifically regarding the tracking of updated dates among all hips. By eliminating all but the most recently updated dates, the system ensures a more streamlined and efficient representation of this data. 

Note to the reviewer: this pr will cause the validation checker to fail for not having today's date added as the updated date. That's okay. That is expected.


pathornteng (approved):
LGTM

PR #796, opened 2023-08-30
Add Lockable Fractional Amounts of Fungible Tokens on Hedera

Stephanie-Yi:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-796.md
+++ b/HIP/hip-796.md
@@ -1,5 +1,5 @@
 ---
-hip: 799 \
+hip: 796 \
 Title: Lockable Fractional Amounts of Fungible Tokens on Hedera\
 Author: Sam Wood <sam.wood@luthersystems.com>, Stephanie Yi <stephanie.yi@luthersystems.com>,
 Eleonora Odorizzi <eleonora.odorizzi@toko.network>, Vicky Lio <vicky.lio@toko.network>\
@@ -9,9 +9,10 @@ type: Standards Track\
 Needs-council-approval: Yes\
 Category: Service\
 Status: Idea\
-Created: \
-Discussions-to: \
-Updated:
+Created: 2023-08-30\
+Discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/797\
+replaces: 751, 752\
+Updated: 2023-08-30\
 
 ---
 
--- a/HIP/hip-799.md
+++ b/HIP/hip-799.md
@@ -1,0 +1,727 @@
+---
+hip: 799 \
+Title: Lockable Fractional Amounts of Fungible Tokens on Hedera\
+Author: Sam Wood <sam.wood@luthersystems.com>, Stephanie Yi <stephanie.yi@luthersystems.com>,
+Eleonora Odorizzi <eleonora.odorizzi@toko.network>, Vicky Lio <vicky.lio@toko.network>\
+Working Group: Atul Mahamuni <@atul-hedera>, Leemon Baird <@lbaird>, Richard Bair <@rbair23>, Jasper
+Potts <@jasperpotts>\
+type: Standards Track\
+Needs-council-approval: Yes\
+Category: Service\
+Status: Idea\
+Created: \
+Discussions-to: \
+Updated:
+
+---
+
+## Abstract
+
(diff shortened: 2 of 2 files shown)


sergmetelin (approved):
lgtm

HIP/hip-796.md
```diff
@@ -0,0 +1,728 @@
...
+- A query to the mirror node for her account balance shows the equivalent of:
+
+```
+{
+…
+tokens: {[
+    {
+        token_id: 0.0.123456,
+        balance: [
+            {partition_id: 1, partition_id: 0.0.200001, balance: 10, locked: 5 },
```
david-bakin-sl, line 204: `partition_id` here should be `partition_number`

HIP/hip-796.md
```diff
@@ -0,0 +1,728 @@
...
+```
+{
+    id: 0.0.123456,
+    name: AcmeToken,
+    symbol: ACME,
+    lockKey: 0x111abc..,
+    partitionKey: 0x222abc..,
+    partitionMoveKey: 0x333ace…,
+    partitions: [
+        {partition_number:1, metadata: “ipfs://xxx”, partition_id: 0.0.200001},
```
david-bakin-sl, line 187: What is the purpose of `partition_number`?  `partition_id` is a sufficient primary key.  If `partition_number` is required how is it assigned (appears nowhere else in this document)?

HIP/hip-796.md
```diff
@@ -0,0 +1,728 @@
...
+}
+```
+
+- On 1 June 2024 the holder of the `lock-key` signs a transaction to unlock the tokens in Tranche-A.
+- Alice can attempt to transfer tokens in an individual partition on 1 June 2024.
+- If she transfers tokens from partition-1, that transfer will succeed.
+- If she transfers tokens from partition-3, that transfer will return an appropriate error informing her that her tokens are locked.
+- To make the transfer, she will use the simple form of the current crypto transfer as the existing tokenTransferList. If she wants to transfer tokens from partition-3, she (or the SDK) queries the mirror nodes or the wallets and maps partition-3 to the TokenID of `0.0.200003` and fills that in the token transfer list of the cryptoTransfer transaction.
+
+## Specification
```
david-bakin-sl, line 219: This proposed HIP has a lot of moving parts and though the wording seems clear I can’t help but think it would be more solid if it included some diagrams and/or pseudocode to make things especially concrete.  There’s a lot of room for implementation fuzziness to sneak in to a big pile of words.  At least if there were diagrams and/or pseudocode a contradiction between the two could be noticed and addressed - it might be either the words or the diagram/code - whereas as it stands things might slip though the implementation.  And not just contradictions, but also holes.

In the same way, additional tables would be useful, e.g., which combination of parameters to APIs are _valid_ and which are not?  E.g., some of these APIs have 4 "buckets" on 2-dimensions:  fungible vs non-fungible, and token-definition-id vs partition-id.  It's not clear that _all_ parameters of some of these APIs are acceptable to all 4 buckets.  (And I'm _not_ talking about the difference between "amount" and "serial#", that's clear enough.)

HIP/hip-796.md
```diff
@@ -0,0 +1,728 @@
...
+
+| Partition Key? | Partition-Move-Key? | Supply-Key? | Initial Supply? | Behavior                                                                                                                                                             |
+| -------------- | ------------------- | ----------- | --------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| N              | Y                   | \*          | \*              | Acts like a normal token-def, no partition behavior                                                                                                                  |
```
david-bakin-sl, line 345: Why is this line _not_ `N * * *`?

david-bakin-sl (commented):
Very complicated HIP.  The feature is sound and desirable but as usual - and here particularly - the devil is in the details.

HIP/hip-796.md
Stephanie-Yi: Might be best to remove the first `partition_id` since the second `partition_id: 0.0.200001` is identifiable enough for this partition.

PR #800, opened 2023-08-30
HIP-794: Add reference implementation

steven-sheehy:
**Description**:

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-794.md
+++ b/HIP/hip-794.md
@@ -2,19 +2,20 @@
 hip: 794
 title: Sunsetting Balance File
 author: Michael Heinrichs <michael@swirldslabs.com>
+working-group: Steven Sheehy <@steven-sheehy>
 type: Standards Track
 category: Service
 needs-council-approval: Yes
 status: Last Call
 last-call-date-time: 2023-09-12T07:00:00Z
 created: 2023-08-28
-discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/pull/794
-updated: 2023-08-29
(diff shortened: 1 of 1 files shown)


netopyr (approved):
LGTM

HIP/hip-794.md
```diff
@@ -26,39 +27,64 @@ Calculating and providing balance files requires iterating over all accounts. Th
 
 ## User stories
 
-As a user, I want a stable and responsive system that does not degrade as the usage scales out.
+* As a user, I want a stable and responsive system that does not degrade as the usage scales out.
+
+* As a mirror node operator, I want to provide accurate balance information to users.
   
 ## Specification
```
SimiHunjan, line 34: Do we need to mark this field as deprecated and note in the this section? 

https://github.com/hashgraph/hedera-protobufs/blob/main/streams/account_balance_file.proto

Is there anywhere else we need to flag as deprecated or add comments that it is no longer being produced after the change goes live?

steven-sheehy: > Do we need to mark this field as deprecated and note in this section?

Not sure we should. Mirror nodes still need to read this file for historical data and for bootstrapping from the last well known file.

> Is there anywhere else we need to flag as deprecated or add comments that it is no longer being produced after the change goes live?

The PR notes that `Any references to balance files in Hedera documentation should be de-emphasized or removed.` I didn't search out those references hoping the docs maintainers might.

sergmetelin (approved):
LGTM

PR #801, opened 2023-08-31
HIP-xxx Add support for debugging EVM transactions

stoqnkpL:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


HIP/hip-xxx.md
```diff
@@ -0,0 +1,165 @@
...
+
+- The type of action performed, e.g. `CALL`, `CREATE`, `PRECOMPILE`, `SYSTEM`
+- The specific operation type of a call, e.g. `OP_CALL`, `OP_DELEGATECALL`, `OP_CREATE`, etc
+- Detailed information for each call, e.g. `caller`, `recipient`, `gas`, `input`, etc
+- Information about the result of each call: `output`, `revert_reason`, `error`
+- Information about the depth of each call that can be used to reconstruct the transaction execution. The original action is at depth=0.
+
+Debug transaction with `callTracer`.
+
+The Hedera Consensus nodes produce these traces for each transaction and externalise them in the `CONTRACT_ACTION` sidecar record files. The Mirror nodes that have DEBUG support enabled should ingest the sidecar records. When a DEBUG request with `callTracer` is handled by the Mirror node, it will rely on the data saved in DB, which is the `contract_result` table for information related to the top level transaction and `CONTRACT_ACTION` sidecars info for the nested calls.
```
steven-sheehy, line 95: We should remove this API if it duplicates an existing API.

HIP/hip-xxx.md
```diff
@@ -0,0 +1,165 @@
...
+
+#### **Endpoint 2:** `/api/v1/contracts/results/{transactionIdOrHash}/traceOpcodes`
```
steven-sheehy, line 97: Prefer `/api/v1/contracts/results/{transactionIdOrHash}/opcodes` since trace is a verb and REST API paths are nouns.

HIP/hip-xxx.md
```diff
@@ -0,0 +1,165 @@
...
+memory: <bool> 
+
+// Boolean flag controlling whether to return storage information. Defaults to `false`.
+
+storage: <bool>
+```
+
+Disabling stack, memory and storage traces will significantly reduce the size of the opcode related data. This would be helpful for reducing the network bandwidth. 
+
+**Output**: detailed information for each executed op code.
```
steven-sheehy, line 117: We need to have an example json response that shows all the excepted fields. Doesn't need to be a large output.

steven-sheehy (changes requested):

HIP/hip-xxx.md
stoqnkpL: I added an example output.

HIP/hip-xxx.md
stoqnkpL: updated.

HIP/hip-xxx.md
stoqnkpL: removed.

HIP/hip-xxx.md
```diff
@@ -0,0 +1,191 @@
...
+Example output:
+```
+{
+    "jsonrpc": "2.0",
+    "id": 0,
```
steven-sheehy, line 112: This is an example of the JSON-RPC output. We should just return what's in the result field for the REST API.

HIP/hip-xxx.md
```diff
@@ -0,0 +1,191 @@
...
+    "result": {
+        "gas": 52139,
+        "failed": false,
+        "returnValue": "0000000000000000000000000000000000000000000000000000000000000001",
```
steven-sheehy, line 116: We should probably add 0x prefix to hex values like the rest of the API

steven-sheehy (changes requested):

HIP/hip-xxx.md
```diff
@@ -0,0 +1,191 @@
...
+		tracer: "callTracer"
+		// Options for `opcodeLogger`. Ignored for `callTracer`. Disabling some of the metrics collected will reduce the output.
+	    disableMemory: true,
+	    disableStack: true,
+	    disableStorage: true,
+	  },
+	],
+	"id": 0
+}
+```
```
Nana-EC, line 75: Needs an example of the response format we're going for in either case

Nana-EC (changes requested):
Nice.
Missing response examples for the new APIs on the relay and mirror node

HIP/hip-xxx.md
IvanKavaldzhiev: I have added examples for all cases.

HIP/hip-xxx.md
IvanKavaldzhiev: I adapted the response to be compatible with the REST APIs format.

HIP/hip-xxx.md
IvanKavaldzhiev: I updated the examples for the REST API to include prefixes.

HIP/hip-xxx.md
```diff
@@ -0,0 +1,921 @@
+---
+hip: 801
+title: Add support for debugging EVM transactions
```
Nana-EC, line 3: I think we should be explicit here since there are many other debug API's that exist that aren't covered here but will be in the future
```suggestion
title: Add support for debug_traceTransaction RPC API
```

Nana-EC (approved):
LG.
Suggestion on title

HIP/hip-xxx.md
```diff
@@ -0,0 +1,921 @@
...
+            "gas": 4419,
+            "gasCost": 3,
+            "depth": 1,
+            "stack": [
+                "0x00000000000000000000000000000000000000000000000000000000094f6253",
+                "00000000000000000000000000000000000000000000000000000000000000a5",
+                "0000000000000000000000008fa74f0a79112c8c46c8e6739eb8cc929083298c",
+                "0000000000000000000000000000000000000000000000000000000000000000",
+                "0000000000000000000000000000000000000000000000000000000000000014",
+                "0000000000000000000000000000000000000000000000000000000000000014"
```
steven-sheehy, line 689: Not all binary data has `0x` prefix here and below

HIP/hip-xxx.md
```diff
@@ -0,0 +1,921 @@
...
+Example output:
+
+```
+{
+    "gas": 52139,
+    "failed": false,
+    "returnValue": "0x0000000000000000000000000000000000000000000000000000000000000001",
+    "structLogs": [
+        [...],
+        {
```
steven-sheehy, line 677: Is the `[...],` just to show that there can be many such `structLogs` entries or is it supposed to be some array within the array? It's confusing so we should remove it everywhere. It should be obvious that arrays can have multiple entries.

HIP/hip-xxx.md
```diff
@@ -0,0 +1,921 @@
...
+            [...]
+        ]
+}
+```
+
+For providing output formatted by `opcodeLogger`, the Mirror node will re-execute the transaction using the state from the `contract_state_changes` sidecars produced by the consensus nodes. In this way, we can have a track on all of the storage/memory information and the entire trace of opcodes that were executed during the replay.
+
+⚠️ The output generated from the opcodeLogger is very verbose. The /traceOpcodes endpoint will be disabled by default and can be enabled by users of the Hedera Local node or users running their own Mirror Nodes.
+
+### **Performance considerations**
```
steven-sheehy, line 881: nit: No need to bold a header

HIP/hip-xxx.md
```diff
@@ -0,0 +1,921 @@
...
+                "reason": null
+            },
+            [...]
+        ]
+}
+```
+
+For providing output formatted by `opcodeLogger`, the Mirror node will re-execute the transaction using the state from the `contract_state_changes` sidecars produced by the consensus nodes. In this way, we can have a track on all of the storage/memory information and the entire trace of opcodes that were executed during the replay.
+
+⚠️ The output generated from the opcodeLogger is very verbose. The /traceOpcodes endpoint will be disabled by default and can be enabled by users of the Hedera Local node or users running their own Mirror Nodes.
```
steven-sheehy, line 879: I believe `/traceOpcodes` should be changed to `/opcodes`

HIP/hip-xxx.md
```diff
@@ -0,0 +1,921 @@
...
+
+### **Performance considerations**
+
+Since the output generated by this tracer might be quite large, we could add a `flag` for enabling different type of tracers, based on the usage and setup of the Mirror Archive Node. This would give us control and flexibility. A user would be able to enable `opcodeLogger` in a local environment and get the full response data. The default tracer enabled by default on production, could be the `callTracer`.
```
steven-sheehy, line 883: It would be nice to have worst case numbers to define "large". e.g how many max opcodes can be returned and the largest size of any particular entry with all flags enabled. I believe Danno mentioned in a call but I can't recall and would be good to have here so we can make informative decisions.

HIP/hip-xxx.md
```diff
@@ -0,0 +1,921 @@
...
+                "gasCost": 100,
+                "depth": 1,
+                "stack": null,
+                "memory": [
+                    "0x0000000000000000000000000000000000000000000000000000000000000000",
+                    "0000000000000000000000000000000000000000000000000000000000000000",
+                    "0000000000000000000000000000000000000000000000000000000000000080"
+                ],
+                "storage": {
+                    "0x0000000000000000000000000000000000000000000000000000000000000000": "0x0000000000000000000000000000000000000000000000000000000000000014"
```
steven-sheehy, line 800: Worth considering if it's worthwhile performance wise to remove leading zeros for the REST API response. Though it would mean the relay would have to re-add them I believe to be spec conformant?

HIP/hip-xxx.md
```diff
@@ -0,0 +1,921 @@
...
+            ],
+            "memory": [
+                "0x0000000000000000000000000000000000000000000000000000000000000000",
+                "0000000000000000000000000000000000000000000000000000000000000000",
+                "0000000000000000000000000000000000000000000000000000000000000080"
+            ],
+            "storage": {
+                "0x0000000000000000000000000000000000000000000000000000000000000000": "0x0000000000000000000000000000000000000000000000000000000000000014"
+            },
+            "reason": null
```
steven-sheehy, line 699: Should have an example where reason is not null so we know its type.

HIP/hip-xxx.md
```diff
@@ -0,0 +1,921 @@
...
+Disabling stack, memory and storage traces will significantly reduce the size of the opcode related data. This would be helpful for reducing the network bandwidth. 
+
+**Output**: detailed information for the transaction execution with `stack`, `memory` and `storage` set to `true`
+Example output:
+
+```
+{
+    "gas": 52139,
+    "failed": false,
+    "returnValue": "0x0000000000000000000000000000000000000000000000000000000000000001",
```
steven-sheehy, line 674: There's a lot of concerns with the API that I would push to address if we can never change it again.

* The REST API always uses snake_case for field names.
* The primary array field `structLogs` should be named after the endpoint `opcodes`
* Should include `contract_id`
* Considering pagination again

Are we intending this just as a disabled by default, only intended for passthrough consumption via the relay sort of API? And are we okay with breaking it later if that scope increases and we want to enable it by default?

steven-sheehy (changes requested):

HIP/hip-xxx.md
IvanKavaldzhiev: Yes, if we remove the leading zeros, the relay should populate them. We would gain performance benefits on the mirror-node side, but the overall performance might be the same, due to the extra computation on the relay side.

HIP/hip-xxx.md
IvanKavaldzhiev: Added

HIP/hip-xxx.md
IvanKavaldzhiev: Removed

HIP/hip-xxx.md
IvanKavaldzhiev: Fixed

HIP/hip-xxx.md
IvanKavaldzhiev: Fixed

HIP/hip-xxx.md
IvanKavaldzhiev: I ran a transaction with 1024 nested calls (the maximum allowed by the EVM) with all flags enabled. The output size for the opcodes logger info was 432MB, which seems quite a lot. For a single nested call the size was around 500 KB and the max pc counter is 1895. So for a main + nested transaction we get around 1895 op codes.

The memory of a given contract is limited to 2^256 slots of 32 bytes of entries.
The storage I believe has no limit and you store theoretically even store GB of data if you are ready to pay for this. So it seems we have no control or predictability of the storage size.

I suggest allowing the `/opcodes` endpoint to work only for local environments and additionally add pagination logic (again only locally). Pagination would mean executing the transaction again for each pagination request and fetching different portion of the output, so it won't be so efficient in terms of computation power, but could save us output size.

@Nana-EC , @steven-sheehy what are your thoughts?

HIP/hip-xxx.md
IvanKavaldzhiev: Added example, it's a hex encoded string of the error message.

HIP/hip-xxx.md
IvanKavaldzhiev: I believe the idea is to disable the /opcodes endpoint by default since it could be very memory and computation heavy API. It could be enabled by the user on a local instance, but stays disabled on production environments, controlling this via a feature flag. We could design it in such a way, that if a paid provider services wants to maintain Mirror Archive Node with /opcodes API - they could just change the feature flag and enable it for users.

In any case, I like the idea of pagination, since output could be hundreds of MBs.

HIP/hip-xxx.md
Nana-EC: I don't think it'd scale.
The example likely includes slots that relate to simple data type variables but not all will be and the majority within a class may not have leading zeroes

HIP/hip-xxx.md
steven-sheehy: Still some `0x` missing in first example

HIP/hip-xxx.md
steven-sheehy: What would be the pagination key if we added it? pc?

I assume there's a way to halt execution after reaching the page size to optimize earlier pages, but not continuing from there on the next page?

Since it's all local we could just not add pagination now. Then later if it's necessary we could add it in a backwards compatible way via an extra query param and a `links` section. Also, if we decide to enable this in an environment like testnet we could also consider limiting requests based upon the gas in the db. We query the gas and if it's within some reasonable range query for the opcodes and return it. Otherwise we return an error or partial data

steven-sheehy (changes requested):

HIP/hip-801.md
```diff
@@ -0,0 +1,895 @@
...
+
+## Backward Compatibility
+
+Mirror Nodes that will support the new REST APIs should have enabled importing of CONTRACT_BYTECODE , CONTRACT_STATE_CHANGE and CONTRACT_ACTION sidecar types. Otherwise, executing the endpoints would result in missing runtime bytecode for execution or using stale contract storage data. In addition, the mirror node should have full historical support (keeping state from the first block), so that it could replay transactions from the entire lifetime of the system.
+
+## Security Implications
+
+There would be some security implementations for Mirror Nodes. Since the debug_trace RPC calls would be free of charge and the Mirror Nodes don’t have throttle mechanism, some attack vectors would be possible. The Nodes could be overloaded with huge amount of calls or calls invoking smart contract methods with huge gas usage, both of which might take more system usage and slow down the network for other users. A rate limit mechanism will be implemented, so that the load put on Mirror Nodes to be balanced.
```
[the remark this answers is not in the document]
IvanKavaldzhiev: This is what we meant, that we don't have a gas based throttle. I will elaborate more on this in the description.

Since the initial purpose of the most heavy API, which is the `/opcodes` one would be mainly for local usage, probably we shouldn't put any limits on the response size for the moment.

I believe the size of the response might increase quicker than the gas increase, since for the same gas cost (executing the same opcode like MLOAD or SLOAD) we can load data with different size, depending on the contracts. So if we decide to have a limit mechanism, apart from the rate limit, it should be based on the response size itself.

HIP/hip-xxx.md
IvanKavaldzhiev: I agree with you. For now we can postpone the idea of pagination and come back in case of need. 

We can additionally limit requests based on the potential `gasCharged` value we might introduce inside the importer. This would be a good measurement of how much op codes would be executed and relatively how big the response size would be.

HIP/hip-801.md
```diff
@@ -0,0 +1,895 @@
...
+        "failed": false,
+        "return_value": "",
+        "opcodes": [
+            {
+                "pc": 541,
+                "op": "SLOAD",
+                "gas": 4519,
+                "gas_cost": 100,
+                "depth": 1,
+                "stack": null,
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Fixed

HIP/hip-801.md
```diff
@@ -0,0 +1,895 @@
...
+    ]
+}    
+```
+
+**Output**: detailed information for the transaction execution with `stack` set to `false`
+Example output:
+
+```
+{
+        "contract_id": "0x8fa74f0a79112c8c46c8e6739eb8cc929083298c",
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Fixed

HIP/hip-801.md
```diff
@@ -0,0 +1,895 @@
...
+                "reason": null
+            }
+    ]
+}    
+```
+
+**Output**: detailed information for the transaction execution with `stack` set to `false`
+Example output:
+
+```
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Added

HIP/hip-801.md
```diff
@@ -0,0 +1,895 @@
...
+{
+	"jsonrpc": "2.0",
+	"method": "debug_traceTransaction",
+	"params": [
+      // The transaction hash.
+	  "0x123...", 
+	  {
+		// The tracer to use. Accepted values: `callTracer` and `opcodeLogger`. Optional field. Default value is `callTracer`.
+		tracer: "callTracer"
+		// Options for `opcodeLogger`. Ignored for `callTracer`. Disabling some of the metrics collected will reduce the output.
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Removed

steven-sheehy (approved):
LGTM

mgarbs (approved):

PR #803, opened 2023-09-11
Rename `total_reward` as `max_total_reward` in HIP-786

tinker-michaelj:
**Description**:
 - Per title.


HIP/hip-786.md
```diff
@@ -108,14 +108,14 @@ message NodeStakeUpdateTransactionBody {
...
-  int64 total_reward = 14;
+  int64 max_total_reward = 14;
 }
 ``` 
 
 ### Mirror nodes
 
 If a mirror node API exposes network staking properties, it _should_ expand to include this new information
-and _should_ replace the `staking_reward_rate` field with `total_reward`. For example,
+and _should_ replace the `staking_reward_rate` field with `max_total_reward`. For example,
```
steven-sheehy, line 118: We can't remove `staking_reward_rate` as it would break backwards compatibility plus it was populated for historical data. We should remove the `replace` wording and just mention we're adding new fields and consumers should update to use the new field going forward. Similarly, we should remove the `remove` wording from the REST response below.

mgarbs (approved):

PR #817, opened 2023-09-29
Change postion to relative and add z indexes

mgarbs:
This PR fixes a bug in the UI that had the menu crossing under the search container when expanded.
Before:
image

After:
image


sergmetelin (approved):
LGTM

PR #838, opened 2023-12-04
HIP-820 suggested change `hedera_signTransactionBody` -> `hedera_signTransaction`

hgraphql:
**Description**:

This PR modifies `hedera_signTransactionBody` in order to utilize functionality already provided by the sdks. By using a [`TransactionList`](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/sdk/transaction_list.proto) object for request params and responses, we unlock additional functionality and ease of use for developers implementing the spec. The ease of use improvements include being able to call 2 builtin functions to serialize / deserialize methods.


hgraphql:
@bugbytesinc after further review, looks like the go, javascript, and java sdks use the `TransactionList` construct for their toBytes / fromBytes methods

can you review this PR and let me know if there are any requested changes?

hgraphql:
@mgarbs merging this PR is dependent on @bugbytesinc review

bugbytesinc (changes requested):
The use of `TransactionList` adds complexity and additional security issues that runs contrary to the original design of this HIP.  Using `Transaction` may be acceptable instead.

hgraphql:
> The use of `TransactionList` adds complexity and additional security issues that runs contrary to the original design of this HIP. Using `Transaction` may be acceptable instead.

what impact might this have on the implementation utilizing the `toBytes` / `fromBytes` methods of the SDKs?

would both dApps and Wallets that utilize these methods in the javascript, go, or java sdk would need to further process the output of these functions?

bugbytesinc:
Upon further review of the javascript sdk, the `fromBytes` method (and other assumptions made by the sdk) contains vulnerabilities where a maliciously crafted protobuf can be used to trick the user into signing a rogue transaction they did not intend to approve.  Therefore, I would recommend to all developers to not use the SDKs transaction’s `fromBytes` at all for externally created protobuf.  This problem exists for both `TransactionList` and `Transaction` encodings of protobuf.  Therefore, I cannot in good conscience approve changes to this HIP that change the protobuf wire format from `SignedTransaction` or `TransactionBody`.

hgraphql:
> Upon further review of the javascript sdk, the `fromBytes` method (and other assumptions made by the sdk) contains vulnerabilities where a maliciously crafted protobuf can be used to trick the user into signing a rogue transaction they did not intend to approve. Therefore, I would recommend to all developers to not use the SDKs transaction’s `fromBytes` at all for externally created protobuf. This problem exists for both `TransactionList` and `Transaction` encodings of protobuf. Therefore, I cannot in good conscience approve changes to this HIP that change the protobuf wire format from `SignedTransaction` or `TransactionBody`.

Hi Jason, thanks for further review. Could you provide more clarity on this? I just haven't seen this before and would love to have a better understanding of how this could happen and a user could be misled into approving something they did not intend. Basically, it'd be really helpful if you could outline the userflow where this could happen. Something like:

- user goes to malicious app
- user interacts and a request is made to a wallet
- **because transactionlist was used as a payload, requested transaction display is misleading**
- user approves transaction unwittingly

I definitely don't want to be under the understanding that these methods are the suggested way to serialize transaction information and send it over a public network such as the internet if they are not.

hgraphql:
closing as captured in other PRs

PR #839, opened 2023-12-04
HIP-820 transaction updates to follow SDK practices and terminology

hgraphql:
**Description**:

This PR aims to simplify implementation of the HIP by utilizing formats and features used by the SDKs. The Java, Go, and Javascript official sdks return a transaction response that can be serialized and transmitted over the network. This change suggests utilizing this payload to send back a response from the Hedera network upon executing a transaction. The format looks like this: https://docs.hedera.com/hedera/sdks-and-apis/sdks/transactions/submit-a-transaction

Utilizing this payload, the sdks can then be utilized to get more information about the transaction, such as getting a receipt or record by easily deserializing the payload.

**Error Handling**
Currently, the sdks throw an exception or error when a transaction returns a precheck code signaling something did not succeed with the transaction. dApps and Wallets will already need to handle error responses (often in try/catch blocks) when an `error` object is return by the JSON-RPC payload. More information can be found here - https://www.jsonrpc.org/specification#response_object

By passing an error object when `precheck` fails implementers will more readily be able to follow a error control flow, vs. receiving a success response when it might be expected to return an error. Most end user `success response` experience will occur when a transaction is successful, having error cases be more likely in the event of failure precheck codes.

**Motivation**
Overall, this PR aims to reduce the burden of dApp and Wallet implementers in a shared spec. The intention is for largest ease of use with commonly used tooling - i.e. the SDKs, vs. having the most compact payload over the wire and needing specialized or new implementations on both the dApp and Wallet side.


bugbytesinc:
@bumblefudge curious to your thoughts on the intent of how to do this the Wallet Connect way.

HIP/hip-820.md
```diff
@@ -71,38 +71,34 @@ When a dApp requires only the signature from the controller (wallet), it can use
...
... message and then submit it to the network, it can use the `hedera_signTransactionAndSend` method.  This method accepts a base64 encoded protobuf representation of the Hedera API [`TransactionList`](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/sdk/transaction_list.proto) message as input.  The controller decodes the message, interprets the contents, and if valid and approved, appends it’s signature to the [`SignatureMap`](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/services/basic_types.proto#L780) structure identified as the [sigMap](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/services/transaction_contents.proto#L40C13-L40C13) property of the [`SignedTransaction`](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/services/transaction_contents.proto#L31) repackages the message as appropriate for transmission to the Hedera network and sends it to the network.  The controller shall wait for the initial response from the Hedera node and return the resulting Node Precheck Code generated by the submission.
```
bugbytesinc, line 75: `signAndSendTransaction` was chosen to more accurately express what the method does.  It does not wait for execution, it returns precheck values, which does not necessarily imply execution.  Also, it is good to use "Send" as to be distinct from previous efforts that may still be utilized in the WC system.  I do not believe this method name should be changed.

HIP/hip-820.md
```diff
@@ -71,38 +71,34 @@ When a dApp requires only the signature from the controller (wallet), it can use
...
 
 #### Parameters
-`signedTransaction` – a base64 encoding of the protobuf endoded Hedera API [`SignedTransaction`](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/services/transaction_contents.proto#L31) message.  The message must include the [bodyBytes](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/services/transaction_contents.proto#L35C11-L35C11) field representing the transaction, but inclusion of a [sigMap](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/services/transaction_contents.proto#L40C13-L40C13) is optional.  If included, and if it is pre-populated with signatures, the pre-existing signatures must also be included in the message sent to the Hedera Node.
+`transaction` – a base64 encoding of the protobuf endoded Hedera API [`TransactionList`](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/sdk/transaction_list.proto) message.
```
bugbytesinc, line 80: `TransactionList` is not a HAPI construct, it is an artifact of the Hedera JS SDK.  This object is not utilized by all of the SDKs.  One of the goals for this protocol is to be SDK agnostic.  This works against that goal.  Furthermore, this is a vestiture of pre-open access thinking.  We do not want to ask a wallet to sign multiple duplicate transactions and pay the network fees to send those transactions.  The `TransactionList` and `Transaction` protobuf messages are not necessary as the `SignedTransaction` contains just enough information to complete the task and nothing extra.

HIP/hip-820.md
```diff
@@ -71,38 +71,34 @@ When a dApp requires only the signature from the controller (wallet), it can use
...
+### hedera_executeTransaction
+When a dApp only requires the services of the controller to act as a relay to the Hedera network for submitting an already signed transaction, it can use the `hedera_executeTransaction` method.  This method accepts a base64 encoded protobuf representation of the Hedera API [`SignedTransaction`](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/services/transaction_contents.proto#L31) message as input.  The controller decodes the message, interprets the contents, and if valid and approved, transmits the message unaltered to the Hedera network.  The controller shall wait for the initial response from the Hedera node and return the resulting Node Precheck Code generated by the submission.
```
bugbytesinc, line 92: Not interested in changing property name for same reasons stated above.

HIP/hip-820.md
```diff
@@ -71,38 +71,34 @@ When a dApp requires only the signature from the controller (wallet), it can use
...
+`signedTransaction` – a base64 encoding of the protobuf endoded Hedera API [`TransactionList`](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/sdk/transaction_list.proto) message.  The message must include the [bodyBytes](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/services/transaction_contents.proto#L35C11-L35C11) field representing the transaction, and the [sigMap](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/services/transaction_contents.proto#L40C13-L40C13) field containing one or more valid signatures.
```
bugbytesinc, line 97: Same reasons against this change as stated above.

bugbytesinc (changes requested):
Not commenting yet about `precheckCode`, still trying to get a feel for how the Wallet Connect protocol expects us to handle the situation in our case.

HIP/hip-820.md
hgraphql: @bugbytesinc understood and appreciate that opinion.

the cause for updating to "execute" vs. "send" is that it more closely aligns with the SDK terminology of execute.

```go
//Sign with client operator private key and submit the transaction to a Hedera network
txResponse, err := transaction.Execute(client)
```
```javascript
const txResponse = await transaction.execute(client);
```
```java
TransactionResponse txResponse = transaction.execute(client);
```

let's do a straw poll on this, please thumbs up which naming structure is preferred

HIP/hip-820.md
hgraphql: @bugbytesinc the intention is to be able to utilize the `fromBytes` methods in the SDKs to easily reconstruct a transaction. This should be transferable between sdks.

a list is utilized in go, java, and javascript in one way or another. If there's a better description of the payload that can also utilize these methods, please let me know, thanks!

`go` - https://github.com/hashgraph/hedera-sdk-go/blob/d3bf16d478e3a1226d2078ebdc6a7a0d5232c5e1/transaction.go#L95

`java` - https://github.com/hashgraph/hedera-sdk-java/blob/7ffc43474cd1c8532cd3d0d7a4ef740a618bd697/sdk/src/main/java/com/hedera/hashgraph/sdk/Transaction.java#L239

`javascript` - https://github.com/hashgraph/hedera-sdk-js/blob/fbcca47f9653bc0f53edb56b116300f48082bde1/src/transaction/Transaction.js#L197

The intention here is for ease of onboarding, vs. the smallest possible payload

Let's have a straw poll here as well to get other's feedback. Thumbs up on the preferred way of doing this

HIP/hip-820.md
hgraphql: @bugbytesinc understood. Let's see what other opinions there are and we can decide as a group how to move forward. Again, the idea here is to be more in line with the SDK's naming structure, which for new developers and those already utilizing the sdks, might relieve some of the burden of cognitive load.

here's a reference: https://github.com/hashgraph/hedera-improvement-proposal/pull/839#discussion_r1414368506

HIP/hip-820.md
hgraphql: cool, safe to say we can arrive at the same conclusion as this thread - https://github.com/hashgraph/hedera-improvement-proposal/pull/839#discussion_r1414311878

HIP/hip-820.md
```diff
@@ -71,38 +71,43 @@ When a dApp requires only the signature from the controller (wallet), it can use
...
+#### Error
+In the event of a precheck validation failure, the controller shall return an error object vs. a result object in the response of the JSON-RPC payload. For more information see <https://www.jsonrpc.org/specification#response_object>.
+
+`code` - the number `-1`
+
+`message` - a string giving a short description of the error, such as "transaction <tranactionId> failed precheck with status 10"
+
+`data` - an integer representing the precheck status code
+
```
hgraphql, line 99: @bugbytesinc how does this look?

bugbytesinc:
Still waiting on feedback from the wallet connect community with regards to the best way to deal with precheck code return.  Also see posts elsewhere on `TransactionList` which is an unsatisfactory construct to transfer the transaction.

HIP/hip-820.md
Volind: I understand both points of view. But there is another name that could be also considered: "call" (used in Signer)

hgraphql: Great, thanks for the feedback @Volind 

Sounds like there's not a strong opinion either way. While I'd have a preference for what I've suggested, it's ultimately not up to me, and I definitely support not making this update. Just wanted to provide an alternative.

So as I tally:
Jason - strong opinion in favor of not accepting this change
Tyler & Volind - good with either option

We can keep this open for the rest of the day to allow for collecting of any other strong opinions

hgraphql:
> Still waiting on feedback from the wallet connect community with regards to the best way to deal with precheck code return. Also see posts elsewhere on `TransactionList` which is an unsatisfactory construct to transfer the transaction.

@bugbytesinc Sounds good on the precheck issue.

Could you provide more feedback on whether or not you'd expect implementers to use the `toBytes` and `fromBytes` of the SDKs as outlined in this PR.

I imagine the nuance between the two is not as clear to everyone reading this. Though I think it would be beneficial to for others to fully understand how this decision might affect their implementations. I believe it'd be much easier to come to consensus on this, with clarity on the workarounds one might need to deploy when using the sdks.

The `toBytes` return a `TransactionList` from what I can tell and `fromBytes` can take either. In practice, I imagine wallets wouldn't reject a request because it passed a `TransactionList` from using these methods vs a `Transaction`. In the end, the decisions wallets are likely to make on whether or not to accept this would likely come down to the decision on what's best for end users and dApps.

If the understanding is that it's likely dApps and wallets are going to use these methods in their implementations, not conforming to this spec and we want to keep the spec more in line with how you'd like it, that works for me.

From our conversations, I've gathered your suggestion is to NOT use the SDKs for this, is that correct? Or was I misunderstanding?

bugbytesinc:
Upon further review of the javascript sdk, the `fromBytes` method (and other assumptions made by the sdk) contains vulnerabilities where a maliciously crafted protobuf can be used to trick the user into signing a rogue transaction they did not intend to approve.  Therefore, I would recommend to all developers to not use the SDKs transaction’s `fromBytes` at all for externally created protobuf.  This problem exists for both `TransactionList` and `Transaction` encodings of protobuf.  Therefore, I cannot in good conscience approve changes to this HIP that change the protobuf wire format from `SignedTransaction` or `TransactionBody`.

bugbytesinc:
> Could you provide more feedback on whether or not you'd expect implementers to use the `toBytes` and `fromBytes` of the SDKs as outlined in this PR.
> 

A dApp to create the bytes of a SignedTransaction with the following:
```js
let signedTransaction = tx._makeSignedTransaction(AccountId.fromString('0.0.3'));
let bytes = HashgraphProto.proto.SignedTransaction.encode(signedTransaction).finish();
```
A dApp to create the byes of a TransactionBody with the following:
```js
let transactionBody = tx._makeTransactionBody(AccountId.fromString('0.0.3'));
let bytes = HashgraphProto.proto.TransactionBody.encode(transactionBody).finish();
```

bugbytesinc:
A wallet could rehydrate a SignedTransaction with something like the following:
```js
let signedTransactionBytes = ...
let transactionProto = HashgraphProto.proto.Transaction.encoode({ signedTransactionBytes }).finish();
let transaction = Transaction.fromBytes(transactionProto);
```

(However, my reccomendation to all wallets *in general* is to consider using protobuf primitives for security reasons)

hgraphql:
closing per discussion and captured in other PRs

bumblefudge:
> Still waiting on feedback from the wallet connect community with regards to the best way to deal with precheck code return. Also see posts elsewhere on `TransactionList` which is an unsatisfactory construct to transfer the transaction.

@bugbytesinc @hgraphql Apologies for ghosting, I've been traveling and/or sick continuously since Istanbul!  I think the precheck codes are great, 9000 is appropriate to overload, all makes sense to me.  One [probably naïve idea came to mind](https://github.com/WalletConnect/walletconnect-specs/pull/117#discussion_r1428631024) reviewing it, but definitely nothing blocking.  

Was that all you were asking about, or were there other things you wanted my feedback/gut-check on? Sadly the `fromBytes`/`oneof` stuff is over my head/paygrade, can't help there 😅

hgraphql:
> > Still waiting on feedback from the wallet connect community with regards to the best way to deal with precheck code return. Also see posts elsewhere on `TransactionList` which is an unsatisfactory construct to transfer the transaction.
> 
> @bugbytesinc @hgraphql Apologies for ghosting, I've been traveling and/or sick continuously since Istanbul! I think the precheck codes are great, 9000 is appropriate to overload, all makes sense to me. One [probably naïve idea came to mind](https://github.com/WalletConnect/walletconnect-specs/pull/117#discussion_r1428631024) reviewing it, but definitely nothing blocking.
> 
> Was that all you were asking about, or were there other things you wanted my feedback/gut-check on? Sadly the `fromBytes`/`oneof` stuff is over my head/paygrade, can't help there 😅

no worries, thanks for following up!

really good to know 9000 is appropriate to overload

I think that wraps this question up. We were looking to ascertain the best practices for handling hedera network level errors - sending back an `error` vs. `response` in the json-rpc payload

since sending back a 9000 code in the json-rpc payload is appropriate to not overlap with WalletConnect codes, we're moving forward with that to fit closely with common developer workflows needing to handle flows in try/catch blocks

PR #840, opened 2023-12-05
HIP-820 signMessages discussion

hgraphql:
Add an original suggestion to more closely align the `signMessages` function with the signature provider API

**Description**:
This PR is for discussion. In practice, the signer functionality supports multiple messages. This PR is to see if the WalletConnect function should more closely align with that functionality:

https://docs.hedera.com/hedera/sdks-and-apis/sdks/signature-provider/signer

### DIFF
--- a/HIP/hip-820.md
+++ b/HIP/hip-820.md
@@ -7,7 +7,7 @@ type: Standards Track
 category: Application
 needs-council-approval: No
 status: Review
-created: 2023-10-05
+created: 2023-10-06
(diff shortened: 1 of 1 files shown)


bugbytesinc (changes requested):
This should not be changed.  `hedera_signedMessage` is modeled after the ETH equivalent in wallet connect.  Furthermore the `SignatureMap` contains important metadata regarding the keys which sign the message.  This change would break supporting multi-signature scenarios.

Volind:
I would also vote for keeping SignatureMap

hgraphql:
closing per group discussion

PR #844, opened 2023-12-12
Handling and externalisation improvements for account nonce updates

stoqnkpL:
**Description**:

Introduce rules for when to update an Ethereum transaction signer's nonce and a new field for externalising the nonce update in the record stream. This will help resolve issues where the Consensus nodes and the Mirror nodes are out of sync in regards to what the current value of an account's nonce is.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-844.md
+++ b/HIP/hip-844.md
@@ -45,7 +45,7 @@ message ContractFunctionResult {
 
     /**
      * If not null this field specifies what the value of the signer account nonce is post transaction execution. 
-     * For transactions that don't update the signer nonce, this field should be null.
+     * For transactions that don't update the signer nonce (like HAPI ContractCall and ContractCreate transactions) this field should be null.
      */
     Int64Value signer_nonce = 15;
 }
--- a/HIP/hip-844.md
+++ b/HIP/hip-844.md
@@ -6,9 +6,10 @@ working-group: Nana Essilfie-Conduah <nana@swirldslabs.com>, Danno Ferrin <danno
 type: Standards Track
 needs-council-approval: Yes
 category: Service
-status: Review
+status: Last Call
 created: 2023-12-12
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/845
+updated: 2024-01-12
 ---
 
 ## Abstract
--- a/HIP/hip-844.md
+++ b/HIP/hip-844.md
@@ -69,6 +69,8 @@ When importing transactions from the record stream, update the value of the sign
 
 When importing legacy transactions that don't have this field, the Mirror node should consider whether the transaction result is SUCCESS or CONTRACT_REVERT_EXECUTED and increment the signer nonce in these cases. Some statuses such as INSUFFICIENT_GAS can be returned both before and after entering the EVM, there is no way for the Mirror node to know whether the signer nonce should be updated. This was the root cause of multiple issues and the reason for the new protobuf field.
 
+The Mirror node will continue to serve outdated nonce information for some accounts and will provide accurate nonces for accounts that have signed valid EthereumTransactions post this HIP implementation.
+
 ## Security Implications
 
 Having consistency between Mirror node and Consensus node states will improve the overall network security and improve the user experience.
--- a/HIP/hip-844.md
+++ b/HIP/hip-844.md
@@ -7,7 +7,6 @@ type: Standards Track
 needs-council-approval: Yes
 category: Service
 status: Review
-last-call-date-time: 
 created: 2023-12-12
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/845
 ---
--- a/HIP/hip-844.md
+++ b/HIP/hip-844.md
@@ -1,5 +1,5 @@
 ---
-hip: xxx
+hip: 844
 title: Handling and externalisation improvements for account nonce updates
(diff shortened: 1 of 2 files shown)


HIP/hip-844.md
```diff
@@ -0,0 +1,86 @@
...
+Add a new protobuf field to contract_call_local.proto:
+
+```
+message ContractFunctionResult {
+    
+    [...]
+
+    /**
+     * If not null this field specifies what the value of the signer account nonce is post transaction execution. 
+     * For transactions that don't update the signer nonce, this field should be null.
```
steven-sheehy, line 48: To be explicit, this will not be populated for contract create and contract call transactions, right?

HIP/hip-844.md
```diff
@@ -0,0 +1,86 @@
...
+```
+2. Populate the new protobuf field when externalising the transaction.
+
+### Mirror node update
+
+When importing transactions from the record stream, update the value of the signer account's nonce that's persisted on the mirror node side with the value externalised in the new protobuf field.
+
+## Backward Compatibility
+
+When importing legacy transactions that don't have this field, the Mirror node should consider whether the transaction result is SUCCESS or CONTRACT_REVERT_EXECUTED and increment the signer nonce in these cases. Some statuses such as INSUFFICIENT_GAS can be returned both before and after entering the EVM, there is no way for the Mirror node to know whether the signer nonce should be updated. This was the root cause of multiple issues and the reason for the new protobuf field.
```
steven-sheehy, line 70: Currently the mirror node is only considering if it's successful or not and not handling the `CONTRACT_REVERT_EXECUTED`.

steven-sheehy, line 70: Will there be any consideration for a sidecar based nonce migration? Or are customs okay with it only being accurate going forward?

stoqnkpL: I don't think there's a way for the mirror node to differentiate between failures thrown prior to entering the EVM or from inside the EVM execution so the current behavior is as good as it can be IMO. 
When services starts populating the new protobuf field the Mirror node will always have the correct new nonce.

stoqnkpL: There is no migration planned at this stage.

steven-sheehy: Perhaps add a note to the doc somewhere about no migration and will only provide accurate nonces for new accounts.

stoqnkpL: added note

steven-sheehy (approved):
LGTM

HIP/hip-844.md
```diff
@@ -0,0 +1,88 @@
...
+
+The Mirror node will continue to serve outdated nonce information for some accounts and will provide accurate nonces for accounts that have signed valid EthereumTransactions post this HIP implementation.
```
Nana-EC, line 72: You can add a note to show users how to unblock themselves to submit a transaction when they have a wrong nonce.
HAPI account call or the error message from the relay on a submitted transaction.

HIP/hip-844.md
```diff
@@ -0,0 +1,88 @@
...
+     * If not null this field specifies what the value of the signer account nonce is post transaction execution. 
+     * For transactions that don't update the signer nonce (like HAPI ContractCall and ContractCreate transactions) this field should be null.
+     */
+    Int64Value signer_nonce = 15;
+}
+```
+
+### Services update
+
+1. Move the signer nonce update to be right before this code block that starts the transaction execution in the EVM:
```
Nana-EC, line 56: Instead of requiring a single line prior to this that could be refactored accidentally should we extract it to a single method that increments and initiates process e.g. `nonceIncrementAndEvmProcess()`

HIP/hip-844.md
```diff
@@ -0,0 +1,88 @@
...
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/845
+---
+
+## Abstract
+
+Introduce rules for when to update an Ethereum transaction signer's nonce and a new field for externalising the nonce update in the record stream. This will help resolve issues where the Consensus nodes and the Mirror nodes are out of sync in regards to what the current value of an account's nonce is.
+
+## Motivation
+
+Unclear rules for when the nonce of an EthereumTransaction signer account should be updated result in discrepancies between the Consensus nodes' and Mirror nodes' states. This leads to users experiencing issues when reading the account nonce from the Mirror node and trying to use it in a transaction sent to the Consensus node.
```
Nana-EC, line 20: Q: Should we consider listing the error case from the HAPI side and relay that are symptomatic of the problem.
In this way some users might link the issue back to the resolution from this HIP.

Nana-EC (commented):
Looking good, some suggestions.

HIP/hip-844.md
stoqnkpL: Added.

HIP/hip-844.md
stoqnkpL: Sounds like a good idea. But actually, I haven't analysed the modularised codebase implementation of the transaction processor and can't provide exact steps for the implementation fix. Should we leave this for the design doc/work package definition?

HIP/hip-844.md
stoqnkpL: Added the error in the Backward Compatibility section.

Nana-EC (approved):

mgarbs (approved):
LGTM

PR #846, opened 2023-12-12
HIP-820 Changes approved at 2023-12-12 WG Meeting

bugbytesinc:
**Description**:

Additional Changes approved at the 2023-12-12 working group meeting.


hgraphql (commented):
looks great to me.

one minor change from working group meeting is the name of the parameter for `signAndExecuteTransaction` and `executeTransaction` 

this PR suggests `transactionList` to be consistent with the object sent vs. `transaction` which is less specific about the payload

let's see if there are any objections on the param name, though functionality-wise this is consistent with the group decision

thanks so much!

hgraphql (approved):
lgtm

hgraphql:
just noting here, there are some questions regarding implementation with `signMessage` and SignatureMap. In the javascript SDK, SignatureMap seems to be specifically used for transactions, where the `signer` interface returns a `SignerSignature` type.

I'll follow up on these outside of github for now

mgarbs (approved):

PR #850, opened 2024-01-02
Create hip-850.md

ty-swirldslabs:
**Description**:
This PR introduces the implementation of the proposed HIP for enhancing the Supply Key functionality in NFT updates within the treasury wallet. The goal is to enable dynamic, secure updates to NFT data, ensuring data immutability post-distribution. This change is essential for supporting evolving use cases in the Hedera ecosystem.

- Implement Supply Key functionality for NFT data updates in the treasury wallet.
- Ensure data immutability of NFTs post-distribution.
- Add necessary code for controlling NFT updates via the Supply Key.

Discussions Link:
[Discussion 660](https://github.com/hashgraph/hedera-improvement-proposal/discussions/660)


mgarbs (dismissed):

mgarbs (dismissed):

HIP/hip-850.md
```diff
@@ -0,0 +1,69 @@
...
+status: Review
+created: 2024-01-02
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/660
+requested-by: HGraphPunks
+updated: 2024-01-24
+---
+
+## Abstract
+
+This HIP proposes enhancing the Supply Key functionality within the Hedera ecosystem, specifically enabling the Supply Key to update an NFT metadata field while the NFT is held in the treasury wallet. This ensures secure and controlled adaptability of NFTs, meeting dynamic needs while maintaining their integrity post-distribution.
```
Nana-EC, line 17: Q: why does the treasury account require this authority instead of the admin key?

HIP/hip-850.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Motivation
+
+Currently, there is a limitation in dynamically updating NFTs in a controlled manner. The proposed enhancement to the Supply Key functionality addresses this gap, allowing for secure and selective updates of NFTs within the treasury wallet, a feature particularly relevant for various evolving use cases in the Hedera ecosystem.
+
+## Rationale
+
+The enhancement focuses on granting the Supply Key the capability to update the data of an NFT serial number exclusively while it is in the treasury wallet. This approach ensures security and control, preventing unauthorized modifications once the NFT is distributed.
```
Nana-EC, line 25: Q: does this imply the modification may only be done prior to first distribution?
What if the treasury is sent the NFT after it previously distributed it, can it modify it then?
What about an NFT that was minted straight to the user and didn't go to the treasury first, if it is sent to the treasury can the treasury modify it?

HIP/hip-850.md
```diff
@@ -0,0 +1,69 @@
...
+
+## User Stories
+
+- As a game developer, I want to update game piece NFTs with player stats during gameplay, ensuring the data becomes immutable once the player regains ownership.
```
Nana-EC, line 29: These are great illustration of applications but maybe se should state it as a more general user story in terms of treasury account capability.
The user story will effectively be used as an E2E test cases to confirm the HIP.

HIP/hip-850.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Backwards Compatibility
+
+This proposal introduces no backward incompatibilities within the current Hedera ecosystem.
+
+## Security Implications
+
+The primary security consideration is the prevention of unauthorized data modifications to NFTs post-distribution. The treasury wallet requirement for updates addresses this concern effectively. 
+
+There is no added security concern with the ability of Supply Key to adapt the metadata when in the treasury wallet as the Supply Key can also burn and mint new tokens. The security implications of sending a NFT to a treasury wallet that has a Supply Key enabled does not become worse or better with this HIP.
```
Nana-EC, line 44: I think there's an application level security concern here.
For NFTs that are used as authentication for access to experiences or other items the ability to modify the data means there's impacts on the number of NFTs that comply or don't comply with the NFT collection intent.
For example in the game user story example  if a collection of 100 promised that only 5 NFTs existed that had special traits, then the ability of the treasury to change that would change the impression and understanding of the collection overall and the ecosystems built on it

Nana-EC (changes requested):
Some questions

HIP/hip-850.md
jascks: I think this HIP needs to be reconciled with [HIP-657: Mutable metadata fields for dynamic NFTs](https://hips.hedera.com/hip/hip-657), which introduces the collection level `metadata_key`, which can be used to modify the currently immutable individual NFT metadata via the new `TokenUpdateNftsTransactionBody`.

I could be mistaken, but my read is that as long as the `metadata_key` is defined for the token/collection, NFT metadata can be manipulated at any time, regardless of the of the current NFT owner. There are no notions of a treasury wallet, post-distribution or metadata key interactions with the supply key.

HIP/hip-850.md
ty-swirldslabs: This is already how the supply key works. We're following the same flow of how users burn and re-mint tokens with the supply key, which requires treasury wallet to hold the token.

We're adding the update metadata function to the supply key, but this is not a new flow, only 1 additional function to what you can do with the supply key flow as is.

HIP/hip-850.md
ty-swirldslabs: @Nana-EC 
_does this imply the modification may only be done prior to first distribution?_
No, this be done at any time when the token is in the treasury wallet, just like how burning tokens with the supply key currently works

_What if the treasury is sent the NFT after it previously distributed it, can it modify it then?_
Yes, anytime the token is in the treasury wallet, the supply key can modify the metadata field. 

What about an NFT that was minted straight to the user and didn't go to the treasury first, if it is sent to the treasury can the treasury modify it?
Yes, matching parity with how token burning works currently, this supply key adapting metadata of token will work the same exact way.

ty-swirldslabs: @jascks 
For the history of this HIP, it was pulled out of 657 into it's own to make the HIPs smaller and easier to get implemented. The functionality here is separate. 

This would create a dynamic that does not exisit on any network that I know of, where a NFT would become semi-fungible. The metadata could be updated ONLY when in the treasury. 

So if you had a game piece, and you need it updated with new stats earned in a game, if the NFT is sent to the treasury, it can be updated without destroying the provenance. 

When it's sent back to the user, the creator can't manipulate it which provides a layer of security and opens the door to some innovative dynamics that don't exist anywhere else.

HIP/hip-850.md
Nana-EC: Okay, so it's not that the treasury must be the supply key owner, it's rather the supply key owner can modify metadata when the tokens are in the treasury's balance, since it has yet to be disbursed?

HIP/hip-850.md
Nana-EC: Interesting. The concept that a disbursed NFT could have it's metadata changed upon return to the treasury has me thinking.
I guess the serial id doesn't change so it remain identifiable but i'm thinking of real world scenarios where the side effect may not be clear.
Thanks for clarifying

HIP/hip-850.md
Nana-EC: The burn case is fine as that NFT serial is gone a different one may be issued same or different.
The issue is that the same NFT serial can morph unexpectedly across the collection and change assumptions and inferences on the collection.
Not against it just exploring conceptual management by users.

HIP/hip-850.md
```diff
@@ -0,0 +1,69 @@
+---
+hip: 850
+title: Enhancing Supply Key Functionality for NFT Updates in Treasury Wallet
```
rbair23, line 3: Can we use the word "account" instead of "wallet"? Wallet is ambiguous, it sometimes is used to mean account, and sometimes is used to represent the wallet software (i.e. HashPack, Blade, MetaMask, etc)

mgarbs: Change made

mgarbs (dismissed):

HIP/hip-850.md
```diff
@@ -0,0 +1,70 @@
...
+updated: 2024-03-20
+---
+
+## Abstract
+
+This HIP proposes enhancing the Supply Key functionality within the Hedera ecosystem, specifically enabling the Supply Key to update an NFT metadata field while the NFT is held in the treasury account. This ensures secure and controlled adaptability of NFTs, meeting dynamic needs while maintaining their integrity post-distribution.
+
+## Motivation
+
+Currently, there is a limitation in dynamically updating NFTs in a controlled manner. The proposed enhancement to the Supply Key functionality addresses this gap, allowing for secure and selective updates of NFTs within the treasury account, a feature particularly relevant for various evolving use cases in the Hedera ecosystem.
```
Nana-EC, line 22: Can you expand on what the limitation is with some use cases where community might be blocked or find it challenging?

HIP/hip-850.md
Nana-EC: Same for example below

HIP/hip-850.md
```diff
@@ -0,0 +1,70 @@
...
+The enhancement focuses on granting the Supply Key the capability to update the data of an NFT serial number exclusively while it is in the treasury account. This approach ensures security and control, preventing unauthorized modifications once the NFT is distributed.
+
+## User Stories
+
+- As a game developer, I want to update game piece NFTs with player stats during gameplay, ensuring the data becomes immutable once the player regains ownership.
+- As an event organizer, I need to update event ticket NFTs with attendee information while they are in the treasury account, guaranteeing data immutability post-transfer.
+
+## Specification
+
+This HIP requires an update to the Supply Key mechanism, enabling it to modify the metadata of NFTs while they are held in the treasury account. The NFT data should remain immutable once distributed, ensuring integrity and trust in the NFT ecosystem.
```
Nana-EC, line 35: Can we expand on this section?
Which existing transaction are we tapping into. Is it just TokenUpdate or some other flow. We just want to be explicit here.

HIP/hip-850.md
```diff
@@ -0,0 +1,70 @@
...
+
+There is no added security concern with the ability of Supply Key to adapt the metadata when in the treasury account as the Supply Key can also burn and mint new tokens. The security implications of sending a NFT to a treasury account that has a Supply Key enabled does not become worse or better with this HIP.
+
+### Modification of Previously Immutable Data
+This HIP introduces a change to the previous guarantee of immutability for NFT metadata. It is important to note that, under the new functionality, the metadata of an NFT can be updated while it resides in the treasury account. This modification capability, while restricted to the treasury account, is a departure from the guarantee of immutability that was previously assured for NFT metadata once minted. Stakeholders should be aware that this change allows for a controlled flexibility in specific circumstances, while still ensuring that the metadata becomes immutable once the NFT is distributed to end-users.
+
+## How to Teach This
+
+Educational initiatives will be developed to guide users in leveraging the new Supply Key functionality, particularly focusing on use cases like gaming and event management.
```
Nana-EC, line 52: Protobuf and docs.hedera.com should explicitly be updated to note the presence of a supply key may imply metadata modification for NFTs should they be sent back to the treasury

HIP/hip-850.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Specification
+
+This HIP requires an update to the Supply Key mechanism, enabling it to modify the metadata of NFTs while they are held in the treasury account. The NFT data should remain immutable once distributed, ensuring integrity and trust in the NFT ecosystem.
+
+## Backwards Compatibility
+
+This proposal introduces no backward incompatibilities within the current Hedera ecosystem.
+
+## Security Implications
```
Nana-EC, line 41: Is there an implication on the mirror node here?
I imagine a modification of metadata needs to be captured so that I can see the history of an NFT metadata.

Nana-EC (changes requested):
Q: A few more suggestions and waiting on some clarifications.

HIP/hip-850.md
ty-swirldslabs: Currently it's impossible to update the metadata of any NFT. After HIP-657 goes into production, this will be more accessible but still doesn't lend to this specific use case of immutable outside the treasury wallet, and mutable in the treasury wallet.

Example use cases
- A game piece that can be updated when you stake the piece and play the game, but cannot be adapted when it's outside the treasury wallet, guaranteeing that the metadata isn't adapted to increase or decrease the secondary market value remotely

This cannot be achieved while maintaining provenance of history of changes on one serial number of a token id.

HIP/hip-850.md
ty-swirldslabs: That is correct

HIP/hip-850.md
ty-swirldslabs: I don't think so. The field will change from mutable from immutable in the update and when the record stream comes through it'll update the metadata record which will still be accessible through pulling that serial number from the mirror node API for that NFTid.

HIP/hip-850.md
jascks: And, how this interacts with the HIP-657 metadata key.

HIP/hip-850.md
jascks: Mirror node already (0.101+) tracks NFT metadata updated via the new (HIP-657) `TokenUpdateNftsTransaction`. This is facilitated by signing the transaction with the metadata key defined for the token. This can happen any time, and thus the NFT is not immutable. My understanding is that this can be done regardless of whether or not the NFT is held in the treasury account. More testing will be done as 0.49 and the client SDKs move forward.

HIP/hip-850.md
Nana-EC: @ty-swirldslabs this user story could be updated. Game dev is an application of the usage but a User story should be more general to the persona and activity

HIP/hip-850.md
Nana-EC: @ty-swirldslabs this part could be expanded on

HIP/hip-850.md
Nana-EC: @ty-swirldslabs can you clarify this.
If HIP 657 allows for metadata modification why is this HIP needed.

If this HIP was implemented the result would be
1. The metadata key can update anytime
2. The supply key can update metadata only when the token serial is in the treasury balance?

Is this correct? If so why is 2 needed when 1 already works?
If this is needed is the real goal creating immutability for the metadata, in this case when the token is in the treasuries balance?

Also would a supply key be able to update the metadata on a token if there was no metadata key (implying immutability?). 

These considerations should be clarified explicitly in this HIP for clarity

HIP/hip-850.md
Nana-EC: > After HIP-657 goes into production, this will be more accessible but still doesn't lend to this specific use case of immutable outside the treasury wallet, and mutable in the treasury wallet.

After making this comment I came back to the above. It seems like the above is the main aim of the HIP right and the mechanism being used is the supply key? HIP 657 create mutability everywhere when a metadata_key is present but for some scenarios you'd like to have pseudo-dynamic NFTs (mutable in treasury balance and immutable outside)

HIP/hip-850.md
Nana-EC: @ty-swirldslabs can you address these comments in the HIP

HIP/hip-850.md
Nana-EC, line 22: I still thinks it's a bit confusing that the Supply Key is used for this extra functionality.
Supply Key implies control over the supply quantity. This change the role description to include the attributes of the supply also i.e. metadata.

Have we considered building upon HIP 657 instead and utilizing the metadata key and a separate property for immutability when in the treasury?

HIP/hip-850.md
ty-swirldslabs: @Nana-EC 

These HIPs used to be the same HIP, but I was advised to separate them to help with implementation and understanding. They bring two very different functionalities to Hedera Token Service. 

HIP-657:
- Enables remote updating of the metadata field of NFTs in circulation
- All NFTs in collection at all times are update-able when metadata key is present

HIP-850
- Enables treasury updates of NFTs only when held by treasury
- Only NFTs in the treasury wallet are mutable
- All NFTs in the collection in circulation are immutable

657 gets us closer to Smart contract expecations of being able to remotely adaapt metadata at any time, which hedera currently can't do in a decentralized way (centralized server urls would work, but isn't true web3 NFT arch)

850 brings a new paradigm that does not exist anywhere else in Web3 to my knowledge and has vast business use cases. It allows token distributors to gaurentee immutability outside of the treasury, while still being able to update and maintain provenance when the asset is sent back to the distributor. 

Take RWAs being sold on an open market.

If the metadata key was on the token, that RWA could be adapted remotely after it's already been put on a market, or sold to a buyer. This introduces trust that the distributor won't change the assets underlying metadata that dictates it's value. 

However, if the asset didn't have a metadata key, but maintained it's supply key, the distributor would be able to adapt the asset when the owner sends it back to the treasury (house sale, lease update, agreement adaptations of a RWA, etc) but when the owner puts it onto a market, it's guaranteed to be immutable and not adaptable.

I believe this is an extremely powerful dynamic that will help enterprise create functionality for legal needs in the future of tokenization of assets. 

**If HIP 657 allows for metadata modification why is this HIP needed.**

- 657 does not guarantee immutability outside the treasury, leaving need for trust of no remote changes to tokenized asset. 850 provides this unique functionality not see

 **If so why is 2 needed when 1 already works?**
 They provide fundamentally different abilities for token distributors. 657 allows a metadata key to adapt any NFT at any time. 850 only allows NFTs to be adapted in the treasury wallet, guaranteeing immutability when on the market, unlocking functionality that does not  currently exisit on any network I'm aware of. 

**If this is needed is the real goal creating immutability for the metadata, in this case when the token is in the treasuries balance?**
No. It is creating mutability only in the treasury wallet, while guaranteeing immutability outside the treasury wallet. 

**Also would a supply key be able to update the metadata on a token if there was no metadata key (implying immutability?).**
Yes. Metadata key is a completely separate way of adapting data and is not required for the supply key to be able to adapt the metadata of NFts in the treasury wallet.

HIP/hip-850.md
ty-swirldslabs: Yes, I had a conversation about this nuance earlier last week. The issue is that we have about 3 HIPs talking about different instances of metadata and it's complicating the discussions as there's shared terminology that are distinctly different. 

HIP-765 is not referencing the metadata on a NFT id. It is only referencing the metadata field being added to the token level of a token. An admin key will not be able to adapt the metadata of NFT id.

The admin key will be able to adapt the metadata field on the token level, like any other field on the token level.

Neeharika-Sompalli: @ty-swirldslabs So this is specifically when the token doesn't have an admin key or metadata key. We will be able to update metadata using supply key _only_ when in treasury account.

HIP/hip-850.md
Neeharika-Sompalli: @ty-swirldslabs So as I understand this, this is specifically when the token doesn't have an admin key or metadata key. We will be able to update metadata using supply key _only_ when in treasury account?
Because with HIP-657, we will be able to update metadata of NFT when in treasury account using `metadataKey`

ty-swirldslabs:
After some more internal discussion, it's quite hard to get consensus on the best strategy to enable this feature of only being able to adapt metadata of a NFT that's in the treasury wallet.

There's been three different solutions proposed and the next step is to ask the community about the feature and see what they'd prefer out of the paths forward. 

**The options for this feature to be implemented are:**
1. The supply key gains a new functionality as proposed in this HIP. 
2. A new key is created that is on the token level and that key has the functionality described in this HIP
3. The metadata key is updated to have a boolean value. If false, the key can only adapt NFTs metadata if the NFT is in the treasury wallet. If true, the metadata key functionality is as it works today.

Burstall:
So happy to see this functionality coming back for discussion. I can see opinion is divided. 

It seems natural to me that it is added to the Supply Key as:
**Pros**
1) The Supply Key can burn/remint the NFT anyway, so power lies within bounds already (granted token serial would change).
2) This makes it fully retroactive opening up the design space for existing collections without forcing them to conduct a token swap.
3) No owner is impacted as they have a choice not to send NFTs to the treasury
4) Going forward, there is a clear delineation between collections that can only have the metadata adjusted with users' explicit permission by sending the token to the treasury [neat implementation would be a Smart Contract as the Treasury to allow users to interact and 'upgrade' the metadata in an atomic manner] vs. those tokens where the user must trust the team as 'remote' metadata shifts can be conducted without their consent.
5) Less complexity on the metadata key that may be harder for users to understand.

**Cons**
1) I am stuck on the cons.... the metadata key can change the metadata anyway.

HIP/hip-850.md
ty-swirldslabs: That is correct!

Nana-EC:
> So happy to see this functionality coming back for discussion. I can see opinion is divided.
> 
> It seems natural to me that it is added to the Supply Key to add 'controlled mutability': **Pros**
> 
> 1. The Supply Key can burn/remint the NFT anyway, so power lies within bounds already (granted token serial would change).
> 2. This makes it fully retroactive, opening up the design space for existing collections without forcing them to conduct a token swap.
> 3. No owner is impacted, as they have a choice not to send NFTs to the treasury
> 4. Going forward, there is a clear delineation between collections that can only have the metadata adjusted with users' explicit permission by sending the token to the treasury [neat implementation would be a Smart Contract as the Treasury to allow users to interact and 'upgrade' the metadata in an atomic manner] vs. those tokens where the user must trust the team as 'remote' metadata shifts can be conducted without their consent.
> 5. Less complexity on the metadata key that may be harder for users to understand.
> 6. Less 'deleted' serials for a token as mistake / bad mints can be updated rather than deleted and reminded on a new serial.
> 
> **Cons**
> 
> 1. I am stuck on the cons.... the metadata key can change the metadata anyway.

Very useful perspective, thanks @Burstall.
I like the idea and the final functionality seems to really open up what NFT collections can do especially in gaming and RWA applications as @ty-swirldslabs has highlighted.

Some considerations regarding your pros that I'm curious about

1. Agreed, but currently the supply key controls [create and delete only](https://github.com/hashgraph/hedera-protobufs/blob/main/services/token_create.proto#L137) and therefore supply. Granted when minting the metadata of an NFT is specified. Mutability of an existing entity could be seen as a different and new power.
2. Indeed and therefore has a backwards compatibility concern. The supply key would now do more than initially stated. Might this have business impacts on a dev or company that concluded supply key can only affect supply quantity by via mint or burn? Is there a heavy desire to use the supply key over metadata key to enable the functionality on older tokens and avoid a token collection owner/community having to migrate to a new token?
3. Agreed
4. Hmm, what about when both keys are present on a token? The flow you layout is pretty interesting. In the SC case a user would view this functionality as submitting their NFT back for some kind of an upgrade or touch up upon some agreed logic. NFT collection owners would support this by utilizing the treasury account and supply key. 🤔 Sounds fun
5. I think the opposite could also be argued. One could says it's less complexity because you know only a metadata key could ever modify metadata in any circumstance. If we add the supply key modification case the community needs education to know that for a token collection with a supply key individual NFT serial metadata can be modified by the metadata key at any time if present and/or by the supply key if the NFT is in the treasury balance.
6. Agreed this is a pro. Note, this is the case for either solution - metadata or supply key. 

All in all we just want to ensure the feature has a consistent feel that scales for devs and users when interacting with HTS

Burstall:
> > So happy to see this functionality coming back for discussion. I can see opinion is divided.
> > It seems natural to me that it is added to the Supply Key to add 'controlled mutability': **Pros**
> > 
> > 1. The Supply Key can burn/remint the NFT anyway, so power lies within bounds already (granted token serial would change).
> > 2. This makes it fully retroactive, opening up the design space for existing collections without forcing them to conduct a token swap.
> > 3. No owner is impacted, as they have a choice not to send NFTs to the treasury
> > 4. Going forward, there is a clear delineation between collections that can only have the metadata adjusted with users' explicit permission by sending the token to the treasury [neat implementation would be a Smart Contract as the Treasury to allow users to interact and 'upgrade' the metadata in an atomic manner] vs. those tokens where the user must trust the team as 'remote' metadata shifts can be conducted without their consent.
> > 5. Less complexity on the metadata key that may be harder for users to understand.
> > 6. Less 'deleted' serials for a token as mistake / bad mints can be updated rather than deleted and reminded on a new serial.
> > 
> > **Cons**
> > 
> > 1. I am stuck on the cons.... the metadata key can change the metadata anyway.
> 
> Very useful perspective, thanks @Burstall. I like the idea and the final functionality seems to really open up what NFT collections can do especially in gaming and RWA applications as @ty-swirldslabs has highlighted.
> 
> Some considerations regarding your pros that I'm curious about
> 
> 1. Agreed, but currently the supply key controls [create and delete only](https://github.com/hashgraph/hedera-protobufs/blob/main/services/token_create.proto#L137) and therefore supply. Granted when minting the metadata of an NFT is specified. Mutability of an existing entity could be seen as a different and new power.
> 2. Indeed and therefore has a backwards compatibility concern. The supply key would now do more than initially stated. Might this have business impacts on a dev or company that concluded supply key can only affect supply quantity by via mint or burn? Is there a heavy desire to use the supply key over metadata key to enable the functionality on older tokens and avoid a token collection owner/community having to migrate to a new token?
> 3. Agreed
> 4. Hmm, what about when both keys are present on a token? The flow you layout is pretty interesting. In the SC case a user would view this functionality as submitting their NFT back for some kind of an upgrade or touch up upon some agreed logic. NFT collection owners would support this by utilizing the treasury account and supply key. 🤔 Sounds fun
> 5. I think the opposite could also be argued. One could says it's less complexity because you know only a metadata key could ever modify metadata in any circumstance. If we add the supply key modification case the community needs education to know that for a token collection with a supply key individual NFT serial metadata can be modified by the metadata key at any time if present and/or by the supply key if the NFT is in the treasury balance.
> 6. Agreed this is a pro. Note, this is the case for either solution - metadata or supply key.
> 
> All in all we just want to ensure the feature has a consistent feel that scales for devs and users when interacting with HTS

Good questions. 

1. I think some of this is a philosophical difference. The supply key can burn and remint a token, which is an opportunity to adjust the metadata. The distinction is that the serial will be updated, and the original will be marked as deleted. This process can only happen with user consent; they must send it back to the treasury. In effect, we can already do this now, with the downside of the serial adjusting. We have established that users prefer to keep serials sometimes, so it would be great to let them. The change proposed 'does no harm' as the owner of an NFT has the choice not to engage. As such, I would argue expanding the design space without forcing a choice on a user and keeping the NFT immutable if they choose not to engage is a great choice. 

2. As in 1) I would make the case that the adjustment of metadata is a subset of the power the supply key already has. Forcing a project to have a metadata key forces the collectors to trust the team not to change the metadata. I agree this new metadata key has some solid use cases, but it involves the trust of the team and their key management - I don't think there should be an option where the design space is open without the **need** for that trust as the user retains control.

3. 👍

4. 
a) I'm glad you like my use case - I would love to roll that Smart Contract NFT minter/treasury out to add such functionality (right now, all I expose is a burn - this would be a fun enhancement). 
b) Back to the feedback, as a collector - I dare not mention how many I have collected other than to say I was relieved when you took off the 1k association cap so I could consolidate wallets - being able to collect tokens and knowing they are immutable unless I explicitly give permission [the supply key route] is far superior than being forced to trust the team if they convince me they should have metadata control [metadata key only]. Rugs are perhaps a moot point as the token may hold no value over and above historical significance; however, a rugging team can potentially upgrade your metadata as a 'poison pill' on the way out. Given it is hard for a user to get rid of a token (especially if there are fallback fees!), then having an NFT with a toxic image stuck in your wallet due to a griefer would always make me think twice before buying a token with a metadata key. Similarly, there have been many instances of teams losing control of keys which could see similar impacts; sidenote, hopefully hardware walle support for native HAPI calls and a solid multisig implementation (we use scripts that are functional if not pretty to lock down team wallets) might ease some of that if project creators realise how to use such keys for their tokens.

5. Fair. I was not clear. I meant the metadata key with an enumeration to have a subtle but important difference is tough to expect the 'average user' to grok. I say this from the number of times I have tried to explain keys or, indeed, the Hedera native royalties to people. Our 'average' user base has not read the API docs. 

6. 👍

Nana-EC:
> > > So happy to see this functionality coming back for discussion. I can see opinion is divided.
> > > It seems natural to me that it is added to the Supply Key to add 'controlled mutability': **Pros**
> > > 
> > > 1. The Supply Key can burn/remint the NFT anyway, so power lies within bounds already (granted token serial would change).
> > > 2. This makes it fully retroactive, opening up the design space for existing collections without forcing them to conduct a token swap.
> > > 3. No owner is impacted, as they have a choice not to send NFTs to the treasury
> > > 4. Going forward, there is a clear delineation between collections that can only have the metadata adjusted with users' explicit permission by sending the token to the treasury [neat implementation would be a Smart Contract as the Treasury to allow users to interact and 'upgrade' the metadata in an atomic manner] vs. those tokens where the user must trust the team as 'remote' metadata shifts can be conducted without their consent.
> > > 5. Less complexity on the metadata key that may be harder for users to understand.
> > > 6. Less 'deleted' serials for a token as mistake / bad mints can be updated rather than deleted and reminded on a new serial.
> > > 
> > > **Cons**
> > > 
> > > 1. I am stuck on the cons.... the metadata key can change the metadata anyway.
> > 
> > 
> > Very useful perspective, thanks @Burstall. I like the idea and the final functionality seems to really open up what NFT collections can do especially in gaming and RWA applications as @ty-swirldslabs has highlighted.
> > Some considerations regarding your pros that I'm curious about
> > 
> > 1. Agreed, but currently the supply key controls [create and delete only](https://github.com/hashgraph/hedera-protobufs/blob/main/services/token_create.proto#L137) and therefore supply. Granted when minting the metadata of an NFT is specified. Mutability of an existing entity could be seen as a different and new power.
> > 2. Indeed and therefore has a backwards compatibility concern. The supply key would now do more than initially stated. Might this have business impacts on a dev or company that concluded supply key can only affect supply quantity by via mint or burn? Is there a heavy desire to use the supply key over metadata key to enable the functionality on older tokens and avoid a token collection owner/community having to migrate to a new token?
> > 3. Agreed
> > 4. Hmm, what about when both keys are present on a token? The flow you layout is pretty interesting. In the SC case a user would view this functionality as submitting their NFT back for some kind of an upgrade or touch up upon some agreed logic. NFT collection owners would support this by utilizing the treasury account and supply key. 🤔 Sounds fun
> > 5. I think the opposite could also be argued. One could says it's less complexity because you know only a metadata key could ever modify metadata in any circumstance. If we add the supply key modification case the community needs education to know that for a token collection with a supply key individual NFT serial metadata can be modified by the metadata key at any time if present and/or by the supply key if the NFT is in the treasury balance.
> > 6. Agreed this is a pro. Note, this is the case for either solution - metadata or supply key.
> > 
> > All in all we just want to ensure the feature has a consistent feel that scales for devs and users when interacting with HTS
> 
> Good questions.
> 
> 1. I think some of this is a philosophical difference. The supply key can burn and remint a token, which is an opportunity to adjust the metadata. The distinction is that the serial will be updated, and the original will be marked as deleted. This process can only happen with user consent; they must send it back to the treasury. In effect, we can already do this now, with the downside of the serial adjusting. We have established that users prefer to keep serials sometimes, so it would be great to let them. The change proposed 'does no harm' as the owner of an NFT has the choice not to engage. As such, I would argue expanding the design space without forcing a choice on a user and keeping the NFT immutable if they choose not to engage is a great choice.
> 2. As in 1) I would make the case that the adjustment of metadata is a subset of the power the supply key already has. Forcing a project to have a metadata key forces the collectors to trust the team not to change the metadata. I agree this new metadata key has some solid use cases, but it involves the trust of the team and their key management - I don't think there should be an option where the design space is open without the **need** for that trust as the user retains control.
> 3. 👍
> 
> a) I'm glad you like my use case - I would love to roll that Smart Contract NFT minter/treasury out to add such functionality (right now, all I expose is a burn - this would be a fun enhancement). b) Back to the feedback, as a collector - I dare not mention how many I have collected other than to say I was relieved when you took off the 1k association cap so I could consolidate wallets - being able to collect tokens and knowing they are immutable unless I explicitly give permission [the supply key route] is far superior than being forced to trust the team if they convince me they should have metadata control [metadata key only]. Rugs are perhaps a moot point as the token may hold no value over and above historical significance; however, a rugging team can potentially upgrade your metadata as a 'poison pill' on the way out. Given it is hard for a user to get rid of a token (especially if there are fallback fees!), then having an NFT with a toxic image stuck in your wallet due to a griefer would always make me think twice before buying a token with a metadata key. Similarly, there have been many instances of teams losing control of keys which could see similar impacts; sidenote, hopefully hardware walle support for native HAPI calls and a solid multisig implementation (we use scripts that are functional if not pretty to lock down team wallets) might ease some of that if project creators realise how to use such keys for their tokens.
> 
> 5. Fair. I was not clear. I meant the metadata key with an enumeration to have a subtle but important difference is tough to expect the 'average user' to grok. I say this from the number of times I have tried to explain keys or, indeed, the Hedera native royalties to people. Our 'average' user base has not read the API docs.
> 6. 👍

Thanks, loving these details :)

1. Agreed it's a philosophical nuance. I think we'll have to agree to disagree, which is totally fine 😄. I think the similarity is only present when you aggregate the changes to the system and view it from the end users view which is at a higher level than the transaction level that a node and the network must consider. The impact is mainly around documentation and education, the dev exp is minimal. I'm okay deferring to community spec here. 
2. The trust factor here is the swaying consideration for me. I like the clarity that a supply key only requires trust when you take the explicit step of sending a token back to the treasury. This can also be gated and specified through smart contract as you said so a user knows the logic ahead of time. The metadata_key on the other hands requires trust at all times when present. Thanks for this clarity.
3. 👍🏾
4. Looking forward to seeing this 🙌🏾
5. 😭 , I can appreciate this. For future readers please check out [Hedera Docs](https://docs.hedera.com/hedera) and for a more technical deep dive see [HAPI protobuf specs](https://github.com/hashgraph/hedera-protobufs)
6. 👍🏾 

I love HIPs with community feedback and input, I'm excited to see the views here as it's a good example for others.

I'm gonna unblock my review here as the community has made the context more clear. 
Thanks @ty-swirldslabs and @Burstall

Nana-EC (commented):
@ty-swirldslabs please update the rejected ideas section with the options other than using the `supply_key`

HIP/hip-850.md
Nana-EC: Comments further down expanded on details here.

HIP/hip-850.md
Nana-EC: @ty-swirldslabs I think the HIP needs to explicitly call out which transaction the supply key owner will utilize.
Sounds like its the `TokenUpdateNftsTransaction` . You should note this under the Specification section to note it is reusing an existing HAPI transaction.
It may actually be worth specifying the comments to update that transactions comments to better describe how it can be used

HIP/hip-850.md
Nana-EC: I think this should be called out explicitly in the Backwards compatibility section. For some there is a change in define so we should call it out as a conceptual difference for those that don't expect supply_key to be able to to update metadata of a single NFT after mint

Nana-EC (approved):
LG
My concerns are addressed.

mgarbs (approved):

PR #863, opened 2024-01-23
Update metadata specifications

netopyr:
**Description**:
This PR updates HIP-657, HIP-646, and HIP-765. These HIPs are closely related and should be viewed together.

- Unify the specifications
- Clarify some confusing items
- Remove a few inconsistencies
- Some general cleanup

### DIFF
--- a/HIP/hip-657.md
+++ b/HIP/hip-657.md
@@ -10,7 +10,7 @@ status: Accepted
 last-call-date-time: 2023-02-28T07:00:00Z
 created: 2023-01-04
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/607
-updated: 2023-03-07
+updated: 2024-01-23
 ---
 
 
--- a/HIP/hip-657.md
+++ b/HIP/hip-657.md
@@ -58,9 +58,9 @@ The METADATA key will also allow updating the metadata introduced by these HIPs.
 ## Specification
 
 
-The METADATA key will be a new key that allows for updating the metadata of an NFT in a new UpdateNft transaction.
+The METADATA key will be a new key that allows for updating the metadata of an NFT in a new UpdateNfts transaction.
 
-The UpdateNft transaction should accept the following parameters:
+The UpdateNfts transaction should accept the following parameters:
 
 - `nft_id`: the identifier (`NftID`) of the NFT to be updated
 - `metadata`: the new metadata for the NFT
@@ -68,7 +68,7 @@ The UpdateNft transaction should accept the following parameters:
 
 Example transaction:
 ```js
-const transaction = await TokenUpdateNft()
+const transaction = await TokenUpdateNfts()
      .setTokenId(tokenId)
      .setSerialNumbers([nft1, nft2, nft3])
      .setMetadata(newMetadata)
@@ -79,7 +79,7 @@ const transaction = await TokenUpdateNft()
 
 Protobuf
 ```
-message TokenUpdateNftTransactionBody {
+message TokenUpdateNftsTransactionBody {
     /**
      * The token for which to update NFTs.
      */
--- a/HIP/hip-657.md
+++ b/HIP/hip-657.md
@@ -38,7 +38,7 @@ The use of the METADATA key to update NFT metadata will not affect the functiona
 ## User stories
 
 
-- As a collector of NFTs, I want to be able to update the metadata of an NFT in my collection so that I can add new information or make changes to the existing metadata.
+- As a creator of NFTs, I want to be able to update the metadata of an NFT in my collection so that I can add new information or make changes to the existing metadata.
 
 
 - As a creator of NFTs, I want to be able to update the metadata of an NFT while leaving an auditable on-chain history of when the update happened and which key was used to sign for the update action so that I can retain as much transparency as possible. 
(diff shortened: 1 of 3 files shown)


HIP/hip-657.md
```diff
@@ -53,45 +53,38 @@ The use of the METADATA key to update NFT metadata will not affect the functiona
...
      */
-    TokenID token = 1;
-
-    /**
-     * The metadata key of the Token. If Token is immutable, transaction will resolve to
-     * TOKEN_IS_IMMUTABlE.
-     */
-     
-     Key metadataKey = 2;
+    NftID nft_id = 1;
```
steven-sheehy, line 82: Believe this is the first time that `NftId` is being used by HAPI, so probably we should also define it in this HIP. I think it is only used by services state currently.

HIP/hip-765.md
```diff
@@ -10,41 +10,16 @@ status: Accepted
...
-
-As a wallet or explorer, I want to query a mirror node to obtain the metadata for any NFT Collection token, similar to current individual non-fungible tokens. Mirror nodes currently return metadata for individual NFTs; but for an NFT Collection token genrally, we want to get **Token-level** (ie Collection-level) metadata.
-
-As a developer, I want to store the metadata uri in a dedicated field, to avoid ambiguity of using an overloaded field such as 'memo'.
-  
-## Specification
-
-## HAPI Changes
-=======
 This HIP proposes the addition of the metadata field to Non-Fungible Token Class, taking after the individual Non-Fungible Token (NFT) metadata field, which was added in [HIP-17](https://hips.hedera.com/hip/hip-17). While HIP-17 metadata is scoped to a single NFT, here we propose adding metadata scoped to an entire token class.
```
steven-sheehy, line 18: We should list the related HIPs 646 and 657 and note that 646 is adding the same fields.

HIP/hip-657.md
ty-swirldslabs: Would it be easier for implementation / less of an anti-pattern if the format matched burning tokens where the tokenId and the serial numbers are set in the transaction instead?
https://docs.hedera.com/hedera/sdks-and-apis/sdks/token-service/burn-a-token

HIP/hip-657.md
```diff
@@ -53,45 +53,38 @@ The use of the METADATA key to update NFT metadata will not affect the functiona
...
 
-The METADATA key will be a new key that allows for updating the metadata of an NFT in a new NftMetadataUpdate transaction.
+The METADATA key will be a new key that allows for updating the metadata of an NFT in a new UpdateNft transaction.
 
-The NftMetadataUpdate transaction should accept the following parameters:
-	- identifier (string): the 0.0.<token-type>.<serial-no>
-- the new NFT metadata: the new metadata for the NFT
+[HIP-646](https://hips.hedera.com/hip/hip-646) and [HIP-765](https://hips.hedera.com/hip/hip-765) will add metadata on the token level. The METADATA key will also allow the metadata to be updated on the token level.
 
+The UpdateNft transaction should accept the following parameters:
```
ty-swirldslabs, line 60: HIP-646 is for Fungible Tokens

HIP/hip-657.md
```diff
@@ -53,45 +53,38 @@ The use of the METADATA key to update NFT metadata will not affect the functiona
...
+- `nft_id`: the identifier (`NftID`) of the NFT to be updated
+- `metadata`: the new metadata for the NFT
 
 
 Example transaction:
 ```js
-NftMetadataUpdate()
-     .setTokenId(tokenId)
-     .setMetaData(newMetaData)
+const transaction = await TokenUpdateNft()
```
ty-swirldslabs, line 68: @netopyr 
Should this just be TokenUpdate? or TokenMetadataUpdate?

When converging with all three HIPs, one is referncing the token level metadata for a FT.

I don't see a different transaction type for the FT and if we remove NFT from this defintion it should work for both token types I think. thoughts?

HIP/hip-657.md
ty-swirldslabs: This path would also allow multiple NFTs that need to be updated to the same metadata to be executed in one transaction submition, better developer experience.

netopyr: I think this is an excellent idea @ty-swirldslabs. Will discuss it with the team.

netopyr: I have updated the parameter list to match `TokenBurn`.

HIP/hip-765.md
netopyr: Done

HIP/hip-657.md
netopyr: Clarified the document

HIP/hip-657.md
netopyr: I need some help understanding this comment. `TokenUpdate` allows metadata to be modified on the token level. This applies to fungible and non-fungible tokens. `TokenUpdateNfts` modifies individual NFTs. There is no equivalent functionality needed for fungible tokens.

steven-sheehy (approved):
LGTM

HIP/hip-657.md
ty-swirldslabs: My thought process here is that if the partition HIP comes through with partitions on fungible tokens, and those partitions have metadata ability, we will have to make a new call to update each of those partitions if the metadata key is set. 

Conceptually, if partitions are treated as new tokens (they'll have individual entities with parent entities assigned I believe) and that is treated the same way as a token level update, then all my concerns are gone and I agree with this path.

mgarbs (approved):

HIP/hip-657.md
```diff
@@ -102,24 +104,36 @@ message NftMetadataUpdateTransactionBody {
...
+     * The key which can change the metadata of a token definition or individual NFT.
      */
-    Key metadataKey = 11;
+    Key metadata_key = 17;
 }
 ```
 
 
-Requirements: The NFT must exist on the Hedera network and should only be allowed to execute if it is signed by the new METADATAKEY. The new metadata must be a valid byte array limited to 100 bytes. The update must be made atomically and be recorded on the Hedera network.
+Requirements: The NFT must exist on the Hedera network and should only be allowed to execute if it is signed by the new METADATA key. The new metadata must be a valid byte array limited to 100 bytes. The update must be made atomically and be recorded on the Hedera network.
```
jascks, line 122: Should not the current metadata key be signing this rather than the new key?

netopyr: When we change keys, we always require both keys to sign the signature.

PR #868, opened 2024-01-24
Create HIP: Adopt Cancun EVM SELFDESTRUCT semantics in Hedera

shemnon:
**Description**:

HIP to adop the SELFDESTRUCT semantics added to the EVM as part of the Cancun fork.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

This is one of four related HIPs covering Cancun support.

**Checklist**

### DIFF
--- a/HIP/hip-868.md
+++ b/HIP/hip-868.md
@@ -1,5 +1,5 @@
 ---
-hip: <HIP number (this is determined by the HIP editor)>
+hip: 868
 title: Support Cancun Self-Destruct Semantics in Smart Contract Services
 author: Danno Ferrin (@shemnon)
 working-group: Nana Essilfie-Conduah <@nana-ec>, Jasper Potts <@jasperpotts>, Richard Bair <@rbair23>, Stoyan Panayotov <stoyan.panayotov@limechain.tech>
@@ -7,12 +7,8 @@ type: Standards Track
 category: Core
 needs-council-approval: Yes
 status: Draft
-created: // TBD
-discussions-to: // TBD
-updated: <comma separated list of dates>
-requires: <HIP number(s)>
-replaces: <HIP number(s)>
-superseded-by: <HIP number(s)>
+created: 2024-01-24
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/872
 ---
 
 ## Abstract
--- a/HIP/hip-868.md
+++ b/HIP/hip-868.md
@@ -2,7 +2,7 @@
 hip: 868
 title: Support Cancun Self-Destruct Semantics in Smart Contract Services
 author: Danno Ferrin (@shemnon)
-working-group: Nana Essilfie-Conduah <@nana-ec>, Jasper Potts <@jasperpotts>, Richard Bair <@rbair23>, Stoyan Panayotov <stoyan.panayotov@limechain.tech>
+working-group: Nana Essilfie-Conduah <@nana-ec>, Jasper Potts <@jasperpotts>, Richard Bair <@rbair23>, Stoyan Panayotov <stoyan.panayotov@limechain.tech>, David Bakin <@david-bakin-sl>
 type: Standards Track
 category: Core
 needs-council-approval: Yes
--- a/HIP/hip-868.md
+++ b/HIP/hip-868.md
@@ -95,10 +95,14 @@ contract will not be marked for removal.
 
 ### Hedera Services
 
-Hedera services may need to be changed to ensure that the transaction processing
-does not leave a "shadow" account number for transient self-destructs. Such
-transient self-destructs will also need to be tested to ensure all HTS tokens
-are properly swept during the sweep as well as the transient contract modes.
+This HIP addresses only EVM changes with respect to self-destruct, and thus,
+only involves changes to the case where a SELFDESTRUCT is done as a sweep
+(the contract being destructed existed prior to the transaction in which it is
+destructed).  When it is a transient self-destruct no changes to existing
+Hedera behavior will be made.
+
+* For a transient self-destruct the record file will record the EVM address
+  of the contract and not any Hedera contract id.
 
 ## Backwards Compatibility
 
@@ -131,7 +135,7 @@ term stable.
 
 ## Reference Implementation
 
-//TODO
+A reference implementation will be provided later.
 
 ## Rejected Ideas
 
@@ -142,7 +146,15 @@ out-of-date features within Hedera.
 
 ## Open Issues
 
-// NONE
+There will be EVM non-equivalences where self-destructed contracts hold HTS tokens
+and/or use Hedera system contracts, especially in the transient case (where the HTS
+tokens are transferred to the contract in the same transaction as it has been
+self-destructed), where the Hedera security model interacts with the sweeping of
+tokens and balances.  For example, if, the beneficiary account doesn't have an
+allowance for a token held by the contract.  In this case (and similar situations)
+the SELFDESTRUCT will fail, where in the EVM it never fails.  These non-equivalences
+already exist and will not be resolved by this HIP (but can be reconsidered in
+the future.)
 
 ## References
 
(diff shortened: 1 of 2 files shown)


HIP/hip-868.md
```diff
@@ -0,0 +1,155 @@
...
+sweep mode changes will be added. All hBar and HTS tokens will be swept as part
+of the self-destruct to the beneficiary address, what will change is that the
+contract will not be marked for removal.
+
+### Hedera Services
+
+Hedera services may need to be changed to ensure that the transaction processing
+does not leave a "shadow" account number for transient self-destructs. Such
+transient self-destructs will also need to be tested to ensure all HTS tokens
+are properly swept during the sweep as well as the transient contract modes.
```
Nana-EC, line 101: @Q: Does Ethereum have support for cases where the SELFDESTRUCT attempt fail for some reason and funds aren't swept?
We may want to support cases where the beneficiary account isn't associated with the HTS token

HIP/hip-868.md
```diff
@@ -0,0 +1,155 @@
...
+similar to how the Besu EVM's implementation was altered.
+
+The existing transient self-destruct behaviors will be retained, and the new
+sweep mode changes will be added. All hBar and HTS tokens will be swept as part
+of the self-destruct to the beneficiary address, what will change is that the
+contract will not be marked for removal.
+
+### Hedera Services
+
+Hedera services may need to be changed to ensure that the transaction processing
```
Nana-EC, line 98: Q: Does the EVM assign an address to the contract that did not exist prior to transaction processing or does it go unaddressed?

Nana-EC: Another question is should that contract call another contract or send cryptocurrency.
What address does the EVM/Ethereum use to express the from address in that account to account interaction?
This may inform how we identify the transient contract

Nana-EC (changes requested):
This is the more intricate of the Cancun features.
Left a few questions to inform some final consideration

HIP/hip-868.md
shemnon: No.  Selfdestruct always happens.  The only way it doesn't stick is if the outer calling contract reverts.  Then all sub-calls side effects are fully reverted.

HIP/hip-868.md
shemnon: You know the address of the contract during the initcode, so yes.  The case it matters for 6780 would be CREATE2 initcode, where the address is a function within the EVM of the salt and initcode.  We are safe not allocating a hedera address until after the full transaction completes (until the hedera account service precompile launches...)

The other case is contract creation transaction.  That already has a high overhead cost to users.

HIP/hip-868.md
david-bakin-sl: This HIP should specify what happens, then, in this case where the beneficiary account doesn't have an association to the HTS token?  _Allowing_ the sweep to happen in that case breaks our rules for tokens (and seems to allow for exploits), but if not doing it then what?

HIP/hip-868.md
```diff
@@ -0,0 +1,155 @@
...
+does not leave a "shadow" account number for transient self-destructs. Such
+transient self-destructs will also need to be tested to ensure all HTS tokens
+are properly swept during the sweep as well as the transient contract modes.
+
+## Backwards Compatibility
+
+This change could break contracts that depend on selfdestruct destroying old
+contracts. Relying on such behaviors has been discouraged for years and formally
+deprecated for at least a year on Ethereum Mainnet, so it is expected that such
+use cases have never found their way onto Hedera's Mainnet.
```
david-bakin-sl, line 108: Is it worth looking at mirror node data to see if in fact it _has_ happened on mainnet? (Or confirm it hasn't, as expected.)

HIP/hip-868.md
```diff
@@ -0,0 +1,155 @@
...
+In the event such a pattern is discovered, the ability to use HAPI calls to
+delete the contract can be used as a shim to restore proper functionality
+between transactions.
+
+## Security Implications
+
+The substantial changes of this HIP are the removal of some functionality in
+some circumstances. There are no known security impacts of this removal and
+furthermore there are no known security mitigations that require the presence of
+the removed functionality. It is expected the security implications will be
```
david-bakin-sl, line 119: "minimal to none" -> why not just "none" then?

(Although I see that [EIP-6780](https://eips.ethereum.org/EIPS/eip-6780#security-considerations) says something more specific anyway - does it need to be repeated here?

image


)

david-bakin-sl (commented):
Left some questions.

HIP/hip-868.md
```diff
@@ -0,0 +1,155 @@
...
+swept to the beneficiary. Contracts within the same transaction will be
+destroyed as per prior behavior.
+
+However, since Hedera places "ownership" of token balances with the account
+holding the balance we will need to extend the behavior of the sweep mode to
+also transfer all HTS tokens to the beneficiary. Notably, for a successful
+transaction beneficiary accounts must support the receiving of HTS tokens via
+association configuration. This is consistent with current self-destruct
+behavior and consistent with the spirit of the Ethereum change in that code and
+the account are no longer deleted.
```
stoqnkpL, line 66: Currently, if the contract holds HTS token balance the self-destruct fails with `TRANSACTION_REQUIRES_ZERO_TOKEN_BALANCES`. This specification is changing the current behavior.

HIP/hip-868.md
Nana-EC: What about for

Sweep (choose 1)
1. Revert the overall transaction if the transfer account is not associated with the token. This stays in line with Hedera architecture since HTS tokens is an outside EVM
2. Best effort to transfer tokens to transfer account. If it's not associated the transaction does fail since the HBAR is swept. I'm leaning to this option since Sweep on the EVM doesn't factor in token balances anyway


Delete (intention with fallback logic)
1. transfer tokens to the transfer account noted in the selfdestruct
2. If transfer account is not associated transfer tokens to the treasurer as the last option.

Wdyt?

HIP/hip-868.md
Nana-EC: The problem is if we don't create an account then the record file will have an ambiguous contractId as the recipient of any transfers as part of the sweep.

If we can get a unique address that will never be reused then we should issue an contractId with the record files showing ContractCreate, Contract Delete child transactions

HIP/hip-868.md
Nana-EC: Yes, let's check

Nana-EC (commented):
In follow up seems we'll
- keep it simple and focus on EVM changes only here
- utilize pre-existing flow to govern HTS balances and transient address/account considerations

shemnon:
David's changes LGTM

HIP/hip-868.md
```diff
@@ -0,0 +1,167 @@
+---
+hip: 868
+title: Support Cancun Self-Destruct Semantics in Smart Contract Services
+author: Danno Ferrin (@shemnon)
+working-group: Nana Essilfie-Conduah <@nana-ec>, Jasper Potts <@jasperpotts>, Richard Bair <@rbair23>, Stoyan Panayotov <stoyan.panayotov@limechain.tech>
```
Nana-EC, line 5:
```suggestion
working-group: Nana Essilfie-Conduah <@nana-ec>, Jasper Potts <@jasperpotts>, Richard Bair <@rbair23>, Stoyan Panayotov <stoyan.panayotov@limechain.tech>, David Bakin <@david-bakin-sl>
```

Nana-EC (dismissed):

stoqnkpL (approved):

IvanKavaldzhiev (approved):
LGTM

HIP/hip-868.md
```diff
@@ -0,0 +1,167 @@
...
+type: Standards Track
```
mgarbs, line 6: We added a new header for council members to see who requested hips. Could you add the 'requested-by' header here and add whoever requested this hip?

Nana-EC: This is part of EVM equivalence and ensuring the network continues to support the latest Ethereum Hardfork features.
There's no one or few requestors. Just the network staying up to date to ensure support

mgarbs (approved):

PR #886, opened 2024-02-14
Remove 851

mgarbs:
Council is not opposed, just needs more discussion


RaphaelMessian (approved):

mgarbs:
> Council is not opposed, just needs more discussion

I removed this but need to bring it back  which is what this PR is https://github.com/hiero-ledger/hiero-improvement-proposals/pull/1182

PR #920, opened 2024-03-15
Create hip-0000-nft-music-metadata.md

itsbrandondev:
**Description**:
This PR adds a new draft HIP titled "Extending HIP-412 to Establish a Music Metadata Standard". This new HIP was created by The Hashgraph NFT Working Group in collaboration with the community over a period of 8 months. It enables music applications and musicians to fully leverage NFTs on Hedera and for the music industry to easily integrate their existing catalogs.

**Checklist**


hip-0000-nft-music-metadata.md
```diff
@@ -0,0 +1,252 @@
...
+|                                         | pro                     |          | PRO (Rights)            | Array                 | Performance Rights Organization(s) or society(s) associated with mechanical rights                                 |
+|                                         | release_date            |          | Release Date            | String                | The release date of the audio track.String (date format: "UTC timestamp")                                          |
+|                                         | isrc                    |          | ISRC                    | String                | The International Standard Recording Code (ISRC) associated with the audio track.                                  |
+|                                         | lyrics                  |          | Lyrics                  | String                | The time-stamped lyrics for the audio track (symbols in UNICODE), ex: [00:00:00] lyric here [00:00:05] lyric there |
```
bguiz, line 99: Instead of having lyrics inline
```json
"lyrics": "[00:00:00] Echoes in the circuitry... [00:00:30] A digital awakening...",
```
it might be better to simply reference a file instead, using `String (URI)` and perhaps also `String (MIME type)`

The contents of this file could conform to any one of a set of a whitelisted set of file formats that are already in use (and therefore this HIP would not need to define, and for which existing software libraries could be used to parse). e.g. [lrc](https://en.wikipedia.org/wiki/LRC_(file_format)), or even [srt](https://en.wikipedia.org/wiki/SubRip#Format).

hip-0000-nft-music-metadata.md
```diff
@@ -0,0 +1,252 @@
...
+        }
+      },
+      {
+        "uri": "ipfs://track2-uri",
+        "type": "audio/mpeg",
+        "metadata": {
+          "name": "Circuit Breaker",
+          "description": "A powerful track about breaking free from digital constraints.",
+          "properties": {
+            // Similar metadata as for Track 1, with relevant changes for track number, ISRC, etc.
```
bguiz, line 201: I think that there is a music industry/standards based specific modification to this that I'd suggest:
1. Have the 1st entry in the `files` array represent the **album**.
2. This album would have `metadata.properties` that would specify all the common properties of the album plus all of its tracks.
3. Each of the subsequent entries in the `files` array represent **tracks**.
4. Each track would have `metadata.properties` that would only specify its different properties, in other words "with relevant changes for track number, ISRC, etc"
5. The client side applications who ingest/process/display the NFT metadata would be able to identify this album/ track distinction and "copy" the album properties that aren't specified across to each of the tracks (for all properties unspecified by the track)

However, how to make this backward compatible with the existing HIP-412 standard is a bit of a question mark - needs more thought.

hip-0000-nft-music-metadata.md
itsbrandondev: i actually really like this. it seems like a pretty simple change and cleans up the metadata a lot. what do you think of this @ty-swirldslabs

hip-0000-nft-music-metadata.md
itsbrandondev: i really like this approach, but this standard is designed for backwards compatibility to HIP-412. essentially, we think of it as a sub-standard of HIP-412 if that makes sense...

mgarbs (approved):

mgarbs (approved):

hip-0000-nft-music-metadata.md
bguiz: I've thought about this again, and I think there's a fairly straightforward way to implement this in a manner that is backwards compatible with HIP-412:

1. Add a new object at `token.properties.tracks`, which will have all the properties that are common to all tracks
2. In `token.files[].properties` only add the properties that are different between the tracks
3. Specify, as part of the HIP-920 standard (this HIP) the process through which client apps are expected to "merge" properties found in `token.properties.tracks` into `token.files[].properties`.

This way, (1) redundancy is much reduced, and (2) HIP-920 is still backwards compatible with HIP-412

itsbrandondev: ok, i really dig this approach. cleans up the metadata a lot. in music you deal with so many redundancies.

interested to see what @ty-swirldslabs thinks as well

HIP/hip-920.md
```diff
@@ -0,0 +1,252 @@
...
+
+The standard does not directly impact security but emphasizes the importance of accurate and immutable metadata for NFTs, which is crucial for rights and royalty management in music assets.
+
+## How to Teach This
+
+Educational materials will be developed to assist artists, record labels, and NFT creators in understanding and implementing the new Music Metadata Standard in their digital assets.
+
+## Reference Implementation
+
+Multiple projects and artists have already incorporated this working music metadata standard.
```
bguiz, line 236: Since this is largely based on how the structure is defined/ validated for the metadata, I think that there is low hanging fruit here in terms of a reference implementation - specifically I'm thinking of adding a schema. E.g. [JSON schema](https://json-schema.org/)

itsbrandondev: i don't know if this is what you mean, but you can see these working standards live on an NFT project i did as an example: https://dd.style/hellofuture

bguiz: Yes, I think including a link to this project (and others implementing this standard) would be a great idea.
in addition to the website, perhaps also link to it on hashscan (e.g. https://hashscan.io/mainnet/token/0.0.2997600 )
> you can see these working standards live on an NFT project i did as an example: https://dd.style/hellofuture

My suggestion here was something different though - it was to include a schema validator as part of the HIP.  Essentially, take the tables describing each of the fields of the metadata, and turn that into a schema. So that would mean that you can take the schema plus the metadata, and pass those as inputs into a validator, and it would output whether it conforms.
> i don't know if this is what you mean,

itsbrandondev: VERY valuable feedback thanks so much. im going to take your 3 pieces of feedback so far and make sure the working group is aware (we have a call tomorrow). i am a big fan, but this feels the most urgent to update the HIP with proper examples.

i was also realizing i need to fill out the "requested by" field because there are a few more projects/individuals to add.

hip-0000-nft-music-metadata.md
itsbrandondev: hey @bguiz just following up here. the group has determined that inline lyrics would be the ideal approach to minimize requests and simplify the composition of the metadata (in the music world, metadata complexity is expected, for some thousands of lines).

an option to explore would be "if MIME look for URI" but that would add complexity. for the scope of this, i think inline lyrics suit the purpose.

hip-0000-nft-music-metadata.md
itsbrandondev: hey @bguiz following up here as well. the group outlined that what you're describing is largely handled by the "album" object.

HIP/hip-920.md
itsbrandondev: just updating you here @bguiz the group greatly appreciated this feedback, a top priority now.

i have been slammed, but will get the HIP updated with this stuff as soon as i can : )

itsbrandondev:
this HIP has been updated:
* add entities to `requested-by` field
* add examples to `reference implimentation`

mgarbs (approved):

PR #924, opened 2024-03-23
Add filtering for hips

mgarbs:
**Description**:
This PR introduces a dual-filtering system for hips, allowing users to efficiently sort and view proposals based on their type (such as Core, Service, Application etc.) and current status (like Draft, Review, Accepted etc.). It streamlines the process of navigating HIPs. This enhancement is geared towards improving accessibility and manageability of proposal data.

image


sergmetelin (approved):
LGTM

PR #938, opened 2024-04-09
Updates to HIP-904

MiroslavGatsanoga:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-904.md
+++ b/HIP/hip-904.md
@@ -849,7 +849,9 @@ message TokenRejectTransactionBody {
     /**
      * A list of one or more token rejections.<br/>
      * On success each rejected token serial number or balance SHALL be transferred from
-     * the requesting account to the treasury account for that token type.
+     * the requesting account to the treasury account for that token type.<br/>
+     * After rejection the requesting account will still be associated with the token and
+     * if token dissociation is needed then a separate TokenDissociate transaction can be used to achieve it.
      */
     repeated TokenReference rejections = 1;
 }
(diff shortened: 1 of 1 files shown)


HIP/hip-904.md
```diff
@@ -1131,16 +1128,18 @@ existing accounts. During the integration of this HIP we propose the following:
 - If an account has a one or more `maxAutomaticAssociations` slots, then we can infer that the
   account holder is willing/interested in receiving some unsolicited tokens in their account.
   Therefore, the network will leave this setting to be as it is.
-- If an account has an account with `maxAutomaticAssociations` set to max integer value, then we can
+- If an account has `maxAutomaticAssociations` set to max integer value, then we can
```
Nana-EC, line 1131: Honestly I think we should just scrap this bullet point. Since we're not doing a migration this is captured by the previous point.
We can add a sentence to be explicit and say no migration to modify `maxAutomaticAssociations` will be carried out.

HIP/hip-904.md
```diff
@@ -848,9 +848,6 @@ was transferred.
 message TokenRejectTransactionBody {
     /**
      * A list of one or more token rejections.<br/>
-     * Each token rejection SHALL dissociate the requesting account from a single
```
Nana-EC, line 851: You can be explicit and not that after rejection a user is still associated with the token and can take the measure to submit a TokenDissociate as needed.

Nana-EC (commented):
Thanks.
Added 2 suggestions

HIP/hip-904.md
MiroslavGatsanoga: Updated.

mgarbs (approved):

PR #950, opened 2024-04-26
fix: clarify that "invalid" means "unusable" in HIP-540

tinker-michaelj:
**Description**:
 - Clarify that HIP-540's `TokenKeyValidation.NONE` mode does **not** allow the user to set structurally invalid keys in state; but only to omit new keys from the `TokenUpdate` signing requirements.

### DIFF
--- a/HIP/hip-540.md
+++ b/HIP/hip-540.md
@@ -16,7 +16,7 @@ updated: 2024-03-22
 
 ## Abstract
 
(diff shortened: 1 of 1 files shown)


Neeharika-Sompalli (approved):
Makes sense to me. Thanks @tinker-michaelj

netopyr (approved):
LGTM - thanks @tinker-michaelj

ty-swirldslabs (approved):
lgtm

HIP/hip-540.md
```diff
@@ -73,39 +73,39 @@ Neither of these approaches is ideal and could easily be solved by allowing the
 
 ## Language
 
-First, let's address important language to set a clear distinction between "removed" and "invalid".
+First, let's address important language to set a clear distinction between "removed" and "unusable".
```
thenswan, line 76: Should it be '..between "absent" and "unusable"'?

PR #968, opened 2024-05-14
Enhance mirror node rest API to query by token name

AliNik4n:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


HIP/hip-xxx.md
```diff
@@ -0,0 +1,140 @@
...
+
+## User stories
+
+- As a user of mirror node, I want to be able to query with a token name (e.g Sauce) and get the corresponding token ID and information.
+- As a user of mirror node, I want to be able to see full match and partial match to my search. (e.g when querying for Sauce, I want to see result for $sauce and $xsauce)
+- As a user of a Hedera explorer, I want to look up a token by searching for token name.
+
+## Specification
+
+The proposed enhancement involves adding  query parameters `name` and `sortby` to the existing route `/api/v1/tokens` to return information about a token.
```
steven-sheehy, line 34: Please remove all mentions of the `sortby` parameter and its operators. We cannot support such a parameter due to performance reasons. This was already discussed and agreed offline.

HIP/hip-xxx.md
```diff
@@ -0,0 +1,140 @@
...
+**Multiple Query Results**
+
+Names assigned to tokens when tokens are created on the network are not unique. Therefore, many tokens with the same name can exist. Querying a token by a given name will return all the tokens that have the specified name defined in the token name.
+
+**Name Query Parameter**
+
+It enables to specify value that token names should match. The following form can be specified:
+
+- name={value}
+- name={op}:{value} where op is one of the usual operators (`eq`,`ne`,`lt`,`lte`,`gt`,`gte`)
```
steven-sheehy, line 45: Remove all operators. We cannot support any of those except equals and since that's the default we don't need any `op`. Implicitly `name=foo` will search for it anywhere in the string (starts with, ends with, contains) case insensitively.

HIP/hip-xxx.md
```diff
@@ -0,0 +1,140 @@
...
+
+## **Backwards Compatibility**
+
+The change is backwards compatible as it is simply adding two query parameters to an existing route. The data required to filter by name is already stored in the mirror node.
+
+## Security Implications
+
+Full text search indices can slow down ingest performance if the same lexeme shows up in many token names.
+
+https://mainnet-public.mirrornode.hedera.com/api/v1/docs/#/tokens/listTokens
```
steven-sheehy, line 128: Remove random unrelated link.

```suggestion
```

HIP/hip-xxx.md
```diff
@@ -0,0 +1,140 @@
...
+GET /api/v1/tokens?name=sauce
+```
+**Response** 
+[no change to the data provided in the response]
+```json
+{
+"admin_key": {
+"_type": "ED25519",
+"key": "0636cfc5037af7d2a7d4f17228a36bb6946ff011b12c1bd970963c6b4266b1ef"
+},
```
steven-sheehy, line 81: Add proper indentation everywhere

steven-sheehy (changes requested):

HIP/hip-xxx.md
AliNik4n: Updated

HIP/hip-xxx.md
AliNik4n: Updated, let me know if it's accurate now

HIP/hip-968.md
```diff
@@ -0,0 +1,131 @@
+---
+hip: 968
+title: Enhance mirror node rest API to query by token name
+author: Simi Hunjan <SimiHunjan>, Ali Nikan <@alinik4n>
+working-group: Steven Sheehy <@steven-sheehy> Eric <ericleponner> Simon <svienot>
```
Nana-EC, line 5:
```suggestion
working-group: Steven Sheehy <@steven-sheehy>, Eric <ericleponner>, Simon <svienot>
```

HIP/hip-968.md
```diff
@@ -0,0 +1,131 @@
...
+---
+
+## Abstract
+
+This HIP provides the capability of querying tokens by token name in mirror node REST API and getting the corresponding token ID information.
+
+## Motivation
+
+To enhance user experience and allow blockchain explorers such as Hashscan to enable the capability of searching by token name, we want to add this functionality to the mirror node to support querying tokens by token name.
+This is a very common use case across other explorers in the larger ecosystem, and will allow retail users and developers to be able to search for a token by token name instead of token ID.
```
Nana-EC, line 22: Q: On other chains the API is on the explorer I believe, at least the EVM ones. 
In this case we're adding it to the mirror node which has impacts on requests and db perf since token name isn't a unique identifier. 
Was their a design choice here for mirror node vs mirror node explorer?

If there was it could be added to rejected ideas as something previously explored

HIP/hip-968.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Rationale
+
+Many times a user may not know the exact token ID when searching for a token, but they know the name. This capability allow users to search by the name and will return results when there is a partial match or exact match. Community members from bridges, Dexes, etc. have been requesting for this for a while.
+
+## User stories
+
+- As a user of mirror node, I want to be able to query with a token name (e.g Sauce) and get the corresponding token ID and information.
+- As a user of mirror node, I want to be able to see full match and partial match to my search. (e.g when querying for Sauce, I want to see result for $sauce and $xsauce)
+- As a user of a Hedera explorer, I want to look up a token by searching for token name.
```
Nana-EC, line 32: From a web3 view many users will not know whether a token is native or a smart contract.
Is the purpose of this HIP for HTS only?
If so we should call this out.

If not then I think there's missing consideration for searching for smart ERC 20/721 tokens at the least.
That might be better positioned for the explorer than mirror node as it would require some preprocessing.
It may also share requirements with an open ticket on the mirror node to support listing of ERC20/721's 

It's fine to say out of scope which is my guess and advise, but we should be explicit of what's inclusive for this HIP.

Nana-EC (commented):
Looking good, will definitely see usage from the community.
I shared 2 main suggestions about the product choice and the scope of the HIP for tokens.
Both are likely resolve by just updating the HIP with more details.

HIP/hip-968.md
AliNik4n: That's a great point Nana, 
I didn't know there would be a significant difference in processing queries when it comes to EVM tokens on Hedera, 
Considering HTS tokens are more important and all requests we are getting for this feature are for HTS tokens, we can only target HTS tokens here and consider ERC20/721 out of scope. But I would love to understand what is stopping us from including all token types in this HIP. 

@steven-sheehy , @SimiHunjan , would love to get your feedback on this

steven-sheehy: I'm not sure how to go about searching for ERC 20/721 tokens. Potentially if we knew how to determine that during processing of contract creates we could also store those in the `token` table with a special hts=false or erc=true flag or some other table. But think that's outside the scope for now. If we want to add that later and return it in the same API call that would not require a new HIP.

HIP/hip-968.md
```diff
@@ -0,0 +1,131 @@
...
+
+```
+GET /api/v1/tokens?name=sauce
+```
+**Response** 
+[no change to the data provided in the response]
+
+
+```json
+{
```
steven-sheehy, line 72: This output isn't accurate. For `/api/v1/tokens` it actually returns a list of tokens with a next link. Also, it would be useful to show a real example with a proper next link as it's important to point out that the next link will still page based upon `token.id` and not the sort order of the `name`. Here's what I suggest:

`GET /api/v1/tokens?limit=2&name=sauce`

> Note: Most fields omitted for brevity

```json
{
  "tokens": [{
      "name": "SAUCE",
      "symbol": "SAUCE",
      "token_id": "0.0.731861"
    }, {
      "name": "xSAUCE",
      "symbol": "XSAUCE",
      "token_id": "0.0.1460200"
    }
  ],
  "links": {
    "next": "/api/v1/tokens?limit=2&name=sauce&token.id=gt:0.0.1460200"
  }
}
```

HIP/hip-968.md
```diff
@@ -0,0 +1,131 @@
...
+
+
+**Request**
+```
+/api/v1/tokens?name=exampleToken
+```
+**Example Request** 
+
+```
+GET /api/v1/tokens?name=sauce
```
steven-sheehy, line 65: Remove this in favor of my other suggestion

HIP/hip-968.md
```diff
@@ -0,0 +1,131 @@
...
+- name={value}
+- name=`lk`:{value} where `lk` is a new operator
+
+`name`=`lk`:{value} returns all tokens whose name:
+
+- starts with {value}
+- ends with {value}
+- contains {value}
+
+Comparison performed by `lk` operator should be case insensitive.
```
steven-sheehy, line 55: We're not supporting a new `lk` operator. Instead, only `name={value}` will be supported and this allows case insensitive sub-string matches.

```suggestion
A new `name` query parameter will be added to search for tokens by their name. This query parameter will perform a case insensitive sub-string match. That is, it will search for token names that start with, contain, or end with the given value. The usual operators like `eq`, `gt`, `gte`, `lt, `lte`, and `ne` will not be supported.

`/api/v1/tokens?name={value}`
```

steven-sheehy (changes requested):

HIP/hip-968.md
SimiHunjan: The mirror node explorer does not store any data or have its own queries for which it returns information for. It was one of the design choices made early on when the explorer was being developed from what I've been told and since then this has not changed. As @ericleponner would say mirror node explorer is the client part of mirror node.  So when we search to return something in the explorer it is mapped to a specific mirror node query. We can consider things to be in the explorer in a future state when we have support to do these kinds of things.

HIP/hip-968.md
AliNik4n: @steven-sheehy , @Nana-EC added a line to specification to call this out: 

> This new capability only supports HTS tokens at this point and will not include ERC 20/721 tokens, further improvements will be introduced later to support ERC 20/721. 

Let me know me know if any further changes needed here

steven-sheehy (approved):
LGTM

mgarbs (approved):
LGTM

mgarbs (approved):

PR #980, opened 2024-06-05
HIP-980: Extend mirror node APIs with timestamp query

IvanKavaldzhiev:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-980.md
+++ b/HIP/hip-980.md
@@ -6,7 +6,7 @@ working-group: Steven Sheehy <@steven-sheehy>, Nana Essilfie-Conduah <nana@swirl
 type: Standards Track
 category: Mirror
 needs-council-approval: Yes
-status: Accepted
+status: In review
 created: 2024-06-03
 requested-by: Solo Labs
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/981
@@ -15,11 +15,11 @@ updated: 2024-06-25
 
 ## Abstract
 
-This HIP enhances part of the mirror node REST APIs to support querying by timestamp. This will allow users to query for tokens, transactions, and accounts that were created at a specific timestamp.
+This HIP enhances part of the mirror node REST APIs to support querying by timestamp. This will allow users to query for the historical state of an entity on the network.
 
 ## Motivation
 
-To enhance user experience and allow users or developers to query data from a specific point in time we will add a timestamp query parameter to some of the REST APIs, which are missing this support.
+To enhance user experience and allow users or developers to the state of an entity a specific point in time we will add a timestamp query parameter to all of the entity-based REST APIs that currently don't have one.
 This change will be especially valuable for building the foundation of the so-called shadow fork functionality, where a user or developer can use a local mirror node instance but run it against a production environment state.
 Having this support, users or developers will have the ability to use public state data to simulate and troubleshoot existing issues or replay historical transactions using local mirror node instance against production state.
 
@@ -30,9 +30,9 @@ for the simulations.
 
 ## User stories
 
-- As a user of mirror node, I want to be able to query historical data for accounts, contracts, tokens, and transactions that were created at a specific timestamp.
+- As a user of mirror node, I want to be able to query for the historical state of accounts, tokens, and other network entities.
 - As a user of mirror node, I want to be able to run locally an `eth_call`, `eth_estimateGas` or `eth_debugTraceTransaction` call that fails on a production environment against the corresponding public state.
-- As a developer of mirror node, I want to be able to test new feature locally against a real life data from a public environment, so that I have better understanding of the feature behavior.
+- As a mirror node developer, I want to be able to test new features locally against production state, so that I have better understanding of the feature behavior.
 
 ## Specification
 
@@ -47,14 +47,19 @@ The proposed enhancement involves adding a query parameter `timestamp` to the ex
 
 **Timestamp Query Parameter**
 
-A new `timestamp` query parameter will be added to search for different resources. This query parameter will represent a consensus timestamp in a Unix format in `seconds.nanoseconds` format and will have an optional comparison operator. That is, it will search for results that `equal`, `not equal`, are `greater` or `less` than the timestamp itself or its seconds component.
+A new `timestamp` query parameter will be added to search for different resources. This query parameter will represent a consensus timestamp in a Unix format in `seconds.nanoseconds` format and will have an optional comparison operator. That is, it will search for results that `equal`, `greater` or `less` than the timestamp itself or its seconds component.
 
-`/api/v1/accounts/{idOrAliasOrEvmAddress}/tokens?timestamp={value}`
-`/api/v1/accounts/{idOrAliasOrEvmAddress}/nfts?timestamp={value}`
-`/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/crypto?timestamp={value}`
-`/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/nfts?timestamp={value}`
-`/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/tokens?timestamp={value}`
-`/api/v1/tokens/{tokenId}/nfts/{serialNumber}?timestamp={value}`
+The exact names of the operators will be `eq`, `gt` and `lt`. The user shouldn't know the exact timestamp value of the time when an entity was created or updated, so that the parameter to work correctly. The API will return the results that match the timestamp or the closest one to it.
+For example, if we have an `asc` ordering of the results, then `?timestamp=X`, `timestamp=eq:X`, or `timestamp=lte:X` will have the same behaviour and vice versa, where `X` is the desired timestamp.
+
+```
+/api/v1/accounts/{idOrAliasOrEvmAddress}/tokens?timestamp={value}
+/api/v1/accounts/{idOrAliasOrEvmAddress}/nfts?timestamp={value}
+/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/crypto?timestamp={value}
+/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/nfts?timestamp={value}
+/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/tokens?timestamp={value}
+/api/v1/tokens/{tokenId}/nfts/{serialNumber}?timestamp={value}
+```
 
 **Request**
 ```
@@ -63,9 +68,13 @@ A new `timestamp` query parameter will be added to search for different resource
 **Example Request**
 
 ```
-GET api/v1/accounts/4408244/tokens?order=desc?timestamp=1705334449.803393003
+GET api/v1/accounts/4408244/tokens?order=asc&timestamp=lte:1725334849.603493001
 ```
-Note: if ```limit``` is not specified in the request, response will be limited to 25 by default
+Note that the following two requests will return the same response as the query above:
+```
+GET api/v1/accounts/4408244/tokens?order=asc&timestamp=1705334449.803393003
+GET api/v1/accounts/4408244/tokens?order=asc&timestamp=eq:1705334449.803393003
+```
 
 **Response**
 
@@ -92,6 +101,8 @@ Note: if ```limit``` is not specified in the request, response will be limited t
 
 The change is backwards compatible as it is simply adding а query parameter to an existing route. The data required to filter by timestamp is already stored in the mirror node.
 
+If a timestamp query parameter is not provided, the API will use the latest state of the entity as it does currently.
+
 ## Security Implications
 
 Historical timestamp filtering can slow down some of the queries, especially if the timestamp is not indexed or the data set is very large.
@@ -102,12 +113,7 @@ None
 
 ## References
 
-https://mainnet-public.mirrornode.hedera.com/api/v1/docs/#/accounts/idOrAliasOrEvmAddress/tokens
-https://mainnet-public.mirrornode.hedera.com/api/v1/docs/#/accounts/idOrAliasOrEvmAddress/nfts
-https://mainnet-public.mirrornode.hedera.com/api/v1/docs/#/accounts/idOrAliasOrEvmAddress/allowances/crypto
-https://mainnet-public.mirrornode.hedera.com/api/v1/docs/#/accounts/idOrAliasOrEvmAddress/allowances/nfts
-https://mainnet-public.mirrornode.hedera.com/api/v1/docs/#/accounts/idOrAliasOrEvmAddress/allowances/tokens
-https://mainnet-public.mirrornode.hedera.com/api/v1/docs/#/tokens/tokenId/nfts/serialNumber
+https://mainnet-public.mirrornode.hedera.com/api/v1/docs/
 
 ## Copyright/license
 
--- a/HIP/hip-980.md
+++ b/HIP/hip-980.md
@@ -6,11 +6,12 @@ working-group: Steven Sheehy <@steven-sheehy>, Nana Essilfie-Conduah <nana@swirl
 type: Standards Track
 category: Mirror
 needs-council-approval: Yes
-status: In review
+status: Last Call
+last-call-date-time: 2024-07-23T07:00:00Z
 created: 2024-06-03
 requested-by: Solo Labs
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/981
-updated: 2024-06-25
+updated: 2024-07-09
 ---
 
 ## Abstract
--- a/HIP/hip-980.md
+++ b/HIP/hip-980.md
@@ -6,10 +6,11 @@ working-group: Steven Sheehy <@steven-sheehy>, Nana Essilfie-Conduah <nana@swirl
 type: Standards Track
 category: Mirror
 needs-council-approval: Yes
-status: Draft
+status: Accepted
 created: 2024-06-03
 requested-by: Solo Labs
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/981
+updated: 2024-06-25
 ---
 
 ## Abstract
@@ -37,7 +38,8 @@ for the simulations.
 
 The proposed enhancement involves adding a query parameter `timestamp` to the existing routes:
 
-- `/api/v1/accounts/{idOrAliasOrEvmAddress}/tokens` to return correct historical information about an account and their associated tokens for a specific timestamp
+- `/api/v1/accounts/{idOrAliasOrEvmAddress}/tokens` to return correct historical information about an account and their associated fungible tokens for a specific timestamp
+- `/api/v1/accounts/{idOrAliasOrEvmAddress}/nfts` to return correct historical information about an account and their associated nfts for a specific timestamp
 - `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/crypto` to return correct historical information about an account's crypto allowances for a specific timestamp
 - `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/nfts` to return correct historical information about an account's NFT allowances for a specific timestamp
 - `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/tokens` to return correct historical information about an account's token allowances for a specific timestamp
@@ -48,6 +50,7 @@ The proposed enhancement involves adding a query parameter `timestamp` to the ex
 A new `timestamp` query parameter will be added to search for different resources. This query parameter will represent a consensus timestamp in a Unix format in `seconds.nanoseconds` format and will have an optional comparison operator. That is, it will search for results that `equal`, `not equal`, are `greater` or `less` than the timestamp itself or its seconds component.
 
 `/api/v1/accounts/{idOrAliasOrEvmAddress}/tokens?timestamp={value}`
+`/api/v1/accounts/{idOrAliasOrEvmAddress}/nfts?timestamp={value}`
 `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/crypto?timestamp={value}`
 `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/nfts?timestamp={value}`
 `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/tokens?timestamp={value}`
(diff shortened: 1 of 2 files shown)


HIP/hip-980.md
```diff
@@ -0,0 +1,111 @@
...
+- `/api/v1/accounts/{idOrAliasOrEvmAddress}/tokens` to return correct historical information about an account and their associated tokens for a specific timestamp
+- `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/crypto` to return correct historical information about an account's crypto allowances for a specific timestamp
+- `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/nfts` to return correct historical information about an account's NFT allowances for a specific timestamp
+- `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/tokens` to return correct historical information about an account's token allowances for a specific timestamp
+- `/api/v1/tokens/{tokenId}/nfts/{serialNumber}` to return correct historical information about an NFT for a specific timestamp
+
+**Timestamp Query Parameter**
+
+A new `timestamp` query parameter will be added to search for different resources. This query parameter will represent a consensus timestamp in a Unix format in `seconds.nanoseconds` format and will have an optional comparison operator. That is, it will search for results that `equal`, `not equal`, are `greater` or `less` than the timestamp itself or its seconds component.
```
jnels124, line 49: Do we want to support ne case? Semantics are always a bit strange with it.

HIP/hip-980.md
```diff
@@ -0,0 +1,111 @@
...
+
+- As a user of mirror node, I want to be able to query historical data for accounts, contracts, tokens, and transactions that were created at a specific timestamp.
+- As a user of mirror node, I want to be able to run locally an `eth_call`, `eth_estimateGas` or `eth_debugTraceTransaction` call that fails on a production environment against the corresponding public state.
+- As a developer of mirror node, I want to be able to test new feature locally against a real life data from a public environment, so that I have better understanding of the feature behavior.
+
+## Specification
+
+The proposed enhancement involves adding a query parameter `timestamp` to the existing routes:
+
+- `/api/v1/accounts/{idOrAliasOrEvmAddress}/tokens` to return correct historical information about an account and their associated tokens for a specific timestamp
```
mgoelswirlds, line 41: Would we need this for `/api/v1/accounts/{idOrAliasOrEvmAddress}/nfts` as well to get the nft information?

HIP/hip-980.md
jnels124: Do we want to force bounded timestamp queries? (i.e. inidcate a note saying all unbounded timestamp queries will be automatically bounded to +- 30 days give < or >.

jnels124: Nevermind these tables are not partitioned by date

HIP/hip-980.md
IvanKavaldzhiev: I think you're right, I will add this endpoint also to the HIP.

HIP/hip-980.md
IvanKavaldzhiev: The reason to add the not equal semantics is to keep the same logic and behaviour as the existing timestamp filter for some of the other endpoints, e.g. `api/v1/network/fees`.

HIP/hip-980.md
```diff
@@ -0,0 +1,114 @@
...
+## Rationale
+
+Users may want to run historical transactions on a local mirror node instance using public state data. To make this possible, some of the existing REST APIs should be enhanced with a timestamp query parameter, so that historical public data can be fetched and used
+for the simulations.
+
+## User stories
+
+- As a user of mirror node, I want to be able to query historical data for accounts, contracts, tokens, and transactions that were created at a specific timestamp.
+- As a user of mirror node, I want to be able to run locally an `eth_call`, `eth_estimateGas` or `eth_debugTraceTransaction` call that fails on a production environment against the corresponding public state.
+- As a developer of mirror node, I want to be able to test new feature locally against a real life data from a public environment, so that I have better understanding of the feature behavior.
```
steven-sheehy, line 35:
```suggestion
- As a mirror node developer, I want to be able to test new features locally against production state so that I have better understanding of the feature behavior.
```

HIP/hip-980.md
steven-sheehy: Please remove not equal. It's not performant as it can scan too many rows.

HIP/hip-980.md
```diff
@@ -0,0 +1,114 @@
...
+- `/api/v1/accounts/{idOrAliasOrEvmAddress}/nfts` to return correct historical information about an account and their associated nfts for a specific timestamp
+- `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/crypto` to return correct historical information about an account's crypto allowances for a specific timestamp
+- `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/nfts` to return correct historical information about an account's NFT allowances for a specific timestamp
+- `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/tokens` to return correct historical information about an account's token allowances for a specific timestamp
+- `/api/v1/tokens/{tokenId}/nfts/{serialNumber}` to return correct historical information about an NFT for a specific timestamp
+
+**Timestamp Query Parameter**
+
+A new `timestamp` query parameter will be added to search for different resources. This query parameter will represent a consensus timestamp in a Unix format in `seconds.nanoseconds` format and will have an optional comparison operator. That is, it will search for results that `equal`, `not equal`, are `greater` or `less` than the timestamp itself or its seconds component.
```
steven-sheehy, line 50: Though it spells out the operators, it doesn't actually list the operator names like `eq`, `lt`, etc. We should also add an example below with operators

HIP/hip-980.md
```diff
@@ -0,0 +1,114 @@
...
+**Request**
+```
+/api/v1/accounts/{idOrAliasOrEvmAddress}/tokens?timestamp={value}
+```
+**Example Request**
+
+```
+GET api/v1/accounts/4408244/tokens?order=desc?timestamp=1705334449.803393003
+```
+Note: if ```limit``` is not specified in the request, response will be limited to 25 by default
```
steven-sheehy, line 68: This information is not really relevant and should be omitted.

HIP/hip-980.md
```diff
@@ -0,0 +1,114 @@
...
+`/api/v1/tokens/{tokenId}/nfts/{serialNumber}?timestamp={value}`
+
+**Request**
+```
+/api/v1/accounts/{idOrAliasOrEvmAddress}/tokens?timestamp={value}
+```
+**Example Request**
+
+```
+GET api/v1/accounts/4408244/tokens?order=desc?timestamp=1705334449.803393003
```
steven-sheehy, line 66:
```suggestion
GET api/v1/accounts/4408244/tokens?order=desc&timestamp=1705334449.803393003
```

HIP/hip-980.md
```diff
@@ -0,0 +1,114 @@
...
+  ],
+  "links": {
+    "next": null
+  }
+}
+```
+
+## **Backwards Compatibility**
+
+The change is backwards compatible as it is simply adding а query parameter to an existing route. The data required to filter by timestamp is already stored in the mirror node.
```
steven-sheehy, line 93: Add that if the timestamp parameter is omitted it will revert to the previous behavior of querying for the latest state.

HIP/hip-980.md
```diff
@@ -0,0 +1,114 @@
...
+None
+
+## References
+
+https://mainnet-public.mirrornode.hedera.com/api/v1/docs/#/accounts/idOrAliasOrEvmAddress/tokens
+https://mainnet-public.mirrornode.hedera.com/api/v1/docs/#/accounts/idOrAliasOrEvmAddress/nfts
+https://mainnet-public.mirrornode.hedera.com/api/v1/docs/#/accounts/idOrAliasOrEvmAddress/allowances/crypto
+https://mainnet-public.mirrornode.hedera.com/api/v1/docs/#/accounts/idOrAliasOrEvmAddress/allowances/nfts
+https://mainnet-public.mirrornode.hedera.com/api/v1/docs/#/accounts/idOrAliasOrEvmAddress/allowances/tokens
+https://mainnet-public.mirrornode.hedera.com/api/v1/docs/#/tokens/tokenId/nfts/serialNumber
```
steven-sheehy, line 110: `s/mainnet-public/mainnet/g`

HIP/hip-980.md
```diff
@@ -0,0 +1,114 @@
...
+
+**Response**
+
+```json
+{
+  "tokens": [
+    {
+      "automatic_association": true,
+      "balance": 1,
+      "created_timestamp": "1705334449.803393003",
```
steven-sheehy, line 78: This makes it seem like the user has to know the created_timestamp to pass via the timestamp parameter. We should note somewhere the behavior of the timestamp parameter and how it is linked to the implicit or explicit order. That is we return the state of that entity on or before the given timestamp. It doesn't have to be an exact timestamp match for when it was created or updated. So if order is ascending and `timestamp=X`, `timestamp=eq:X`, or `timestamp=lte:X` the behavior is the same. And vice versa.

steven-sheehy (changes requested):

HIP/hip-980.md
IvanKavaldzhiev: Removed

HIP/hip-980.md
```diff
@@ -0,0 +1,114 @@
...
+status: Accepted
+created: 2024-06-03
+requested-by: Solo Labs
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/981
+updated: 2024-06-25
+---
+
+## Abstract
+
+This HIP enhances part of the mirror node REST APIs to support querying by timestamp. This will allow users to query for tokens, transactions, and accounts that were created at a specific timestamp.
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Fixed

HIP/hip-980.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Motivation
+
+To enhance user experience and allow users or developers to query data from a specific point in time we will add a timestamp query parameter to some of the REST APIs, which are missing this support.
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Fixed

HIP/hip-980.md
```diff
@@ -0,0 +1,114 @@
...
+Having this support, users or developers will have the ability to use public state data to simulate and troubleshoot existing issues or replay historical transactions using local mirror node instance against production state.
+
+## Rationale
+
+Users may want to run historical transactions on a local mirror node instance using public state data. To make this possible, some of the existing REST APIs should be enhanced with a timestamp query parameter, so that historical public data can be fetched and used
+for the simulations.
+
+## User stories
+
+- As a user of mirror node, I want to be able to query historical data for accounts, contracts, tokens, and transactions that were created at a specific timestamp.
```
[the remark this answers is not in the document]
IvanKavaldzhiev: Fixed

HIP/hip-980.md
IvanKavaldzhiev: Fixed

HIP/hip-980.md
IvanKavaldzhiev: Fixed

HIP/hip-980.md
IvanKavaldzhiev: Removed

HIP/hip-980.md
IvanKavaldzhiev: Fixed

HIP/hip-980.md
IvanKavaldzhiev: Added

HIP/hip-980.md
IvanKavaldzhiev: Fixed

HIP/hip-980.md
IvanKavaldzhiev: Added more explanation

HIP/hip-980.md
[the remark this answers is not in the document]
steven-sheehy: Whoops, I didn't mean to suggest to remove `query`.

```
To enhance user experience and allow users or developers to query the state of an entity
```

HIP/hip-980.md
```diff
@@ -0,0 +1,120 @@
...
+- `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/nfts` to return correct historical information about an account's NFT allowances for a specific timestamp
+- `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/tokens` to return correct historical information about an account's token allowances for a specific timestamp
+- `/api/v1/tokens/{tokenId}/nfts/{serialNumber}` to return correct historical information about an NFT for a specific timestamp
+
+**Timestamp Query Parameter**
+
+A new `timestamp` query parameter will be added to search for different resources. This query parameter will represent a consensus timestamp in a Unix format in `seconds.nanoseconds` format and will have an optional comparison operator. That is, it will search for results that `equal`, `greater` or `less` than the timestamp itself or its seconds component.
+
+The exact names of the operators will be `eq`, `gt` and `lt`. The user shouldn't know the exact timestamp value of the time when an entity was created or updated, so that the parameter to work correctly. The API will return the results that match the timestamp or the closest one to it.
```
steven-sheehy, line 52:
```suggestion
The exact names of the operators will be `eq`, `gt`, `gte`, `lt`, and `lte`. The user doesn't need to know the exact timestamp at which an entity was created or updated for it to function. The API will return the results that match the timestamp or the closest one to it.
```

HIP/hip-980.md
```diff
@@ -0,0 +1,120 @@
...
+status: In review
+created: 2024-06-03
+requested-by: Solo Labs
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/981
+updated: 2024-06-25
+---
+
+## Abstract
+
+This HIP enhances part of the mirror node REST APIs to support querying by timestamp. This will allow users to query for the historical state of an entity on the network.
```
Nana-EC, line 18:
```suggestion
This HIP enhances part of the mirror node REST APIs to support querying by timestamp. This will allow users to query for the historical state of any entity on the network.
```

HIP/hip-980.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Motivation
+
+To enhance user experience and allow users or developers to query the state of an entity a specific point in time we will add a timestamp query parameter to all of the entity-based REST APIs that currently don't have one.
```
Nana-EC, line 22:
```suggestion
To enhance user experience and allow users or developers to query the state of an entity at a specific point in time we will add a timestamp query parameter to all of the entity-based REST APIs that currently don't have one.
Currently file, topic and schedule entities do not have this capability whiles account, contracts and tokens do.
```

HIP/hip-980.md
```diff
@@ -0,0 +1,120 @@
...
+Having this support, users or developers will have the ability to use public state data to simulate and troubleshoot existing issues or replay historical transactions using local mirror node instance against production state.
+
+## Rationale
+
+Users may want to run historical transactions on a local mirror node instance using public state data. To make this possible, some of the existing REST APIs should be enhanced with a timestamp query parameter, so that historical public data can be fetched and used
+for the simulations.
+
+## User stories
+
+- As a user of mirror node, I want to be able to query for the historical state of accounts, tokens, and other network entities.
```
Nana-EC, line 33: Break this out into a user story per entity being enabled.
This will allow for easy mapping to E2E tests scenarios for other products and clearer understanding of capabilities developers can now utilize.

HIP/hip-980.md
```diff
@@ -0,0 +1,120 @@
...
+- As a user of mirror node, I want to be able to run locally an `eth_call`, `eth_estimateGas` or `eth_debugTraceTransaction` call that fails on a production environment against the corresponding public state.
```
Nana-EC, line 34: Similar to above I would break this out into a user story per API noted

HIP/hip-980.md
```diff
@@ -0,0 +1,120 @@
...
+The exact names of the operators will be `eq`, `gt`, `gte`, `lt` and `lte`. The user doesn't need to know the exact timestamp at which an entity was created or updated for it to function. The API will return the results that match the timestamp or the closest one to it.
+For example, if we have an `asc` ordering of the results, then `?timestamp=X`, `timestamp=eq:X`, or `timestamp=lte:X` will have the same behaviour and vice versa, where `X` is the desired timestamp.
+
+```
+/api/v1/accounts/{idOrAliasOrEvmAddress}/tokens?timestamp={value}
+/api/v1/accounts/{idOrAliasOrEvmAddress}/nfts?timestamp={value}
+/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/crypto?timestamp={value}
+/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/nfts?timestamp={value}
+/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/tokens?timestamp={value}
+/api/v1/tokens/{tokenId}/nfts/{serialNumber}?timestamp={value}
```
Nana-EC, line 61: Missing contracts 
```suggestion
/api/v1/contracts/{idOrAliasOrEvmAddress}?timestamp={value}
/api/v1/tokens/{tokenId}/nfts/{serialNumber}?timestamp={value}
```

HIP/hip-980.md
```diff
@@ -0,0 +1,120 @@
+---
+hip: 980
+title: Enhance mirror node rest APIs with timestamp query parameter
```
Nana-EC, line 3:
```suggestion
title: Enhance mirror node entity rest APIs with timestamp query parameter
```

Nana-EC (commented):
Left some comments

HIP/hip-980.md
IvanKavaldzhiev: The contracts API already utilizes a timestamp filter.

HIP/hip-980.md
IvanKavaldzhiev: Fixed

HIP/hip-980.md
IvanKavaldzhiev: Fixed

HIP/hip-980.md
IvanKavaldzhiev: Good idea, I broke the user stories in separate cases

HIP/hip-980.md
IvanKavaldzhiev: Done

HIP/hip-980.md
IvanKavaldzhiev: Fixed

HIP/hip-980.md
Nana-EC: Gotcha

steven-sheehy (dismissed):
LGTM except some failing checks

Nana-EC (dismissed):

Neurone:
Hi guys, the HIP sounds good, but I suggest evaluating and clarifying some details about the formats and the results of the queries.

When using the timestamp parameter, other endpoints support `array[string]` as format; it's `timestampQueryParam`.

This feature allows developers to use multiple timestamp values to restrict the result, for example, getting all blocks between two timestamps like this:

```
GET https://mainnet-public.mirrornode.hedera.com/api/v1/blocks?timestamp=gt:1720493368.000474097&timestamp=lt:1720493376
```
 
So the question is, **are multiple timestamps accepted for the endpoints mentioned in the HIP-980, like all the other endpoints**?

- In case the HIP supports them, we need to clarify the behavior in those cases:
  - MN returns an array of all the multiple distinct intermediate states between two timestamps;
  - MN returns the merge of the results, i.e., all the tokens owned during that time frame. So, if I get a TokenA, sell it, get a TokenB, sell it, get TokenC, and now I query the MN, it would return a single state with TokenA, TokenB, and TokenC.
- If the HIP doesn't support multiple timestamps, **it is confusing using the same syntax or parameter**. I see three options in this case, and only one seems acceptable to me:
  - We use different parameter names and type like `fixed.timestamp`, `timestamp.fixed`, `timestamp.single`, etc. (**I suggest this option**)
  - We can use the syntax we currently use for the only two endpoints with single timestamp ( /api/v1/topics/messages/{timestamp} and  /api/v1/contracts/{contractIdOrAddress}/results/{timestamp}) where that parameter is in the path (type timestampPathParam), but this way we would lose the query capabilities.
  - We keep `timestamp` for the new parameter and rename the old one (i.e., `timeframe`, `timestamps`, etc.), but we introduce a breaking change.

HIP/hip-980.md
```diff
@@ -0,0 +1,124 @@
...
+  ],
+  "links": {
+    "next": null
+  }
+}
+```
+
+## **Backwards Compatibility**
+
+The change is backwards compatible as it is simply adding а query parameter to an existing route. The data required to filter by timestamp is already stored in the mirror node.
```
Neurone, line 106: The `а` in the phrase 

> ... simply adding а query parameter ...

is a Cyrillic and not a Latin `a`.

You can search the two _As_ above in this web page to verify it ;)

IvanKavaldzhiev: Nice catch :)

IvanKavaldzhiev:
> Hi guys, the HIP sounds good, but I suggest evaluating and clarifying some details about the formats and the results of the queries.
> 
> When using the timestamp parameter, other endpoints support `array[string]` as format; it's `timestampQueryParam`.
> 
> This feature allows developers to use multiple timestamp values to restrict the result, for example, getting all blocks between two timestamps like this:
> 
> ```
> GET https://mainnet-public.mirrornode.hedera.com/api/v1/blocks?timestamp=gt:1720493368.000474097&timestamp=lt:1720493376
> ```
> 
>   So the question is, **are multiple timestamps accepted for the endpoints mentioned in the HIP-980, like all the other endpoints**?
> 
> * In case the HIP supports them, we need to clarify the behavior in those cases:
>   
>   * MN returns an array of all the multiple distinct intermediate states between two timestamps;
>   * MN returns the merge of the results, i.e., all the tokens owned during that time frame. So, if I get a TokenA, sell it, get a TokenB, sell it, get TokenC, and now I query the MN, it would return a single state with TokenA, TokenB, and TokenC.
> * If the HIP doesn't support multiple timestamps, **it is confusing using the same syntax or parameter**. I see three options in this case, and only one seems acceptable to me:
>   
>   * We use different parameter names and type like `fixed.timestamp`, `timestamp.fixed`, `timestamp.single`, etc. (**I suggest this option**)
>   * We can use the syntax we currently use for the only two endpoints with single timestamp ( `/api/v1/topics/messages/{timestamp}` and  `/api/v1/contracts/{contractIdOrAddress}/results/{timestamp}`) where that parameter is in the path (type `timestampPathParam`), but this way we would lose the query capabilities.
>   * We keep `timestamp` for the new parameter and rename the old one (i.e., `timeframe`, `timestamps`, etc.), but we introduce a breaking change.

I think we should support the array[string] format of the timestamp query param. Then we should return a list of all of the distinct intermediate states valid for this time range.

@steven-sheehy what are your thoughts?

mgarbs (approved):

Neurone:
I confirm I have no other topics to discuss on this. HIP-980 is fine now for me.

PR #983, opened 2024-06-07
HIP 904 minor updates

MiroslavGatsanoga:
**Description**:

Updating HIP according to latest updates in design docs and Slack discussions.

**Checklist**


HIP/hip-904.md
```diff
@@ -723,7 +820,7 @@ message TransactionRecord {
      * A list of token transfers.<br/>
      * Each transfer completed SHALL emit an entry in this list.
      */
-    repeated TokenTransferList token_transfer_lists = 11;
+    repeated TokenTransferList tokenTransferLists = 11;
```
jsync-swirlds, line 823: The snake_case version is more correct.  The compiler will generate the camelCase version for languages like Java.

HIP/hip-904.md
```diff
@@ -768,9 +865,7 @@ Scenario 1: If Alice airdrops with the following transfer breakout
   association slots
 - NFT serials 3 & 4 of T4 to Carol, where Carol is not associated with T4 but has available
   association slots
-- 10 FTs of token T5 to Dave, where Dave is not associated with T5 and has no available
-  association slots
-- 5 FTs of token T5 to Dave, where Dave is not associated with T5 and has no available
+- 15 FTs of token T5 to Dave, where Dave is not associated with T5 and has no available
```
jsync-swirlds, line 868: I believe the split was deliberate to show the consolidation of fungible pending air drops.

HIP/hip-904.md
```diff
@@ -800,9 +895,9 @@ The record file emitted from a consensus node will contain
 - Alice intent to transfer 20 T7 tokens to Eric
 - Alice intent to transfer NFT serials 7 & 8 of token T8 to Eric
 
-Note: In the transaction body Alice had split her intended transfers to Dave in two but in the
-result exposed by the record they were combined. TokenAirdrops may be combined in an additive
-fashion. It is not possible to subtract. For this a sender may cancel and reissue their airdrop
+Note: TokenAirdrops for fungible tokens may be combined in an additive
+fashion. It is not possible to subtract. For this a sender may cancel and reissue their airdrop.
+Multiple TokenAirdrops for same NFT serial are not possible.
```
jsync-swirlds, line 900: The former text is better, as it uses an example.  This just makes an assertion that's covered elsewhere.

HIP/hip-904.md
```diff
@@ -811,10 +906,7 @@ Scenario 2: If Dave submits a TokenClaim for token T5 and a Token claim of seria
 - Alice debit of 15 T5 tokens → Dave credit of 15 T5 tokens
 - Alice debit of NFT serial 5 of token T6 → Dave credit of NFT serial 5 of token T6
 
-`automatic_token_associations` containing
-
-- Dave's automatic association with token T5
-- Dave's automatic association with token T6
+`automatic_token_associations` will be empty, but Dave will be associated to token T5 and token T6, meaning that successful TokenClaim acts like explicit token association if needed
```
jsync-swirlds, line 909: Line should wrap at column 80 for readability.

HIP/hip-904.md
```diff
@@ -844,34 +936,104 @@ was transferred.
...
+   * If not set, the payer for this transaction SHALL be the account rejecting tokens.
+   */
+  AccountID owner = 1;
+
+  /**
+   * A list of one or more token rejections.<br/>
+   * On success each rejected token serial number or balance SHALL be transferred from
+   * the requesting account to the treasury account for that token type.<br/>
+   * After rejection the requesting account SHALL continue to be associated with the token.<br/>
+   * if dissociation is desired then a separate TokenDissociate transaction MUST be submitted to remove the association.
```
jsync-swirlds, line 956: Lines should wrap at column 80.

jsync-swirlds, line 956: Nitpick:
if ==> If

HIP/hip-904.md
```diff
@@ -844,34 +936,104 @@ was transferred.
...
-    oneof token_identifier {
-        /**
-         * A fungible/common token type.
-         */
-        TokenID fungible_token = 1;
+  oneof token_identifier {
+    /**
+     * A fungible/common token type.
+     */
+    TokenID fungible_token = 1;
```
jsync-swirlds, line 971: Indentation should be 4 spaces, not 2.

HIP/hip-904.md
MiroslavGatsanoga: Here I am referring to how the field is [named in `main` protobufs branch](https://github.com/hashgraph/hedera-protobufs/blob/main/services/transaction_record.proto#L97).

jsync-swirlds: hmmm...  So this is a pre-existing name?  In that case it will have to match existing.

HIP/hip-904.md
MiroslavGatsanoga: I removed the reference to the example because it was not valid one (see https://github.com/hashgraph/hedera-improvement-proposal/pull/983#discussion_r1633442790).

HIP/hip-904.md
jsync-swirlds: Ahh, I see.  We still need an example to refer to in other text for the intended scenario.  Might be worth ensuring that is present.

HIP/hip-904.md
jsync-swirlds: Note that we may have to include split amounts in two or more "airdrop" transactions in the examples to refer to that setup here.

PR #991, opened 2024-06-14
feat: add permissionless revenue generating topics

kantorcodes:
**Description**:
Introduces a fixed fee system for submitting messages to topics on the Hedera network, aimed at enhancing economic scalability and revenue predictability for topic operators. This change facilitates easier monetization and resource management while supporting token payments for fees.

* Implement fixed fee system for topic message submissions
* Introduce `FixedFee` and `CustomFee` Protobuf messages
* Update `TopicCreateTransactionBody` and `TopicInfo` messages to support custom fees
* Enhance `TransactionRecord` with fee details

### Use Cases Enabled by HIP 991

1. **Monetization for Topic Operators**:
   - Operators can monetize their topics by setting fixed fees for message submissions, creating predictable revenue streams.

2. **Enhanced Economic Scalability**:
   - Developers and businesses can integrate fee structures to monetize access to premium content or services within their applications.

3. **Management of Spam and Quality Control**:
   - Fees deter spam and low-quality submissions, improving the overall quality and relevance of interactions within topics.

4. **Integration with Token Economies**:
   - Accepting fees in tokens other than HBAR expands the utility of tokens within the Hedera ecosystem, fostering broader token adoption.

5. **Predictable Revenue Streams**:
   - Entities managing topics can establish reliable revenue models, enabling better planning and investment in operations.

6. **Support for Decentralized Applications (dApps)**:
   - dApp developers can implement fee structures for posting on premium topics or accessing exclusive content, enabling new business models.

7. **Flexible Economic Models**:
   - Various sectors can adopt custom economic models using topic fees, facilitating innovation in digital asset management and exchange on Hedera.

These use cases collectively enhance the economic viability and utility of topics on the Hedera Consensus Service (HCS), fostering a more robust and diverse ecosystem of decentralized applications and services.

**Notes for reviewer**:
No additional logs or screenshots needed.

**Checklist**

---

### DIFF
--- a/HIP/hip-991.md
+++ b/HIP/hip-991.md
@@ -51,11 +51,11 @@ This document outlines the development of a fixed fee system for the submission 
 
 We propose adding a fixed fee mechanism to the Hedera Consensus Service (HCS) for topic messages, similar to the custom fee structures in the Hedera Token Service (HTS). This will involve creating new Protobuf messages and modifying existing ones to accommodate fixed fees for topics.
 
-### **New Protobuf Messages**
+### **Protobuf Messages Needed**
 
 ### **FixedFee**
 
-The `FixedFee` message represents a fixed fee charged for each message submission to a topic.
+The `FixedFee` message represents a fixed fee charged for each message submission to a topic. This protobuf message already exists so it can be re-used.
 
 ```
 message FixedFee {
@@ -66,16 +66,15 @@ message FixedFee {
 }
 ```
 
-### **CustomFee**
+### **CustomFixedFee**
 
-The `CustomFee` message defines the type of fee and the account receiving the fee assessed during a message submission to the associated topic. A custom fee may only be fixed and must specify a fee collector account to receive the assessed fees.
+The `CustomFixedFee` message defines the type of fee and the account receiving the fee assessed during a message submission to the associated topic. A custom fee may only be fixed and must specify a fee collector account to receive the assessed fees.
 
 ```
-message CustomFee {
-  oneof fee {
-    // Fixed fee to be charged
-    FixedFee fixed_fee = 1;
-  }
+message CustomFixedFee {
+  // Fixed fee to be charged
+  FixedFee fixed_fee = 1;
+
   // The account to receive the custom fee
   AccountID fee_collector_account_id = 3;
 }
@@ -111,7 +110,7 @@ message TopicInfo {
 
 ### **TransactionRecord**
 
-The `TransactionRecord` message is updated to include the list of custom fees assessed as part of executing the transaction represented by this record.
+The `TransactionRecord` message is updated to include the list of custom fees assessed as part of executing the transaction represented by this record. This protobuf message already exists so it can be re-used.
 
 ```
 message TransactionRecord {
--- a/HIP/hip-991.md
+++ b/HIP/hip-991.md
@@ -1,5 +1,5 @@
 ---
-hip: 969
+hip: 991
 title: Permissionless revenue-generating Topic Ids for Topic Operators
 author: Michael Kantor (@kantorcodes), Ty Smith (@ty-swirldslabs)
 type: Standards Track
--- a/HIP/hip-991.md
+++ b/HIP/hip-991.md
@@ -6,7 +6,8 @@ requested-by: TierBot
 type: Standards Track
 category: Core
 needs-council-approval: Yes
-status: Review
+status: Last Call
+last-call-date-time: 2023-07-24T07:00:00Z
 created: 2024-06-14
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/pull/991
 updated: 2024-07-10
--- a/HIP/hip-991.md
+++ b/HIP/hip-991.md
@@ -2,10 +2,14 @@
 hip: 991
 title: Permissionless revenue-generating Topic Ids for Topic Operators
 author: Michael Kantor (@kantorcodes), Ty Smith (@ty-swirldslabs)
+requested-by: TierBot
 type: Standards Track
 category: Core
+needs-council-approval: Yes
+status: Review
 created: 2024-06-14
-requested-by: TierBot
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/pull/991
+updated: 2024-07-10
 ---
 
 ## **Abstract**
--- a/HIP/hip-969.md
+++ b/HIP/hip-969.md
@@ -1,0 +1,221 @@
+---
+hip: 969
+title: Permissionless revenue-generating Topic Ids for Topic Operators
+author: Michael Kantor (@kantorcodes), Ty Smith (@ty-swirldslabs)
+type: Standards Track
+category: Core
+created: 2024-06-14
+requested-by: TierBot
+---
+
+## **Abstract**
+
+This document outlines the development of a fixed fee system for the submission of topic messages on the Hedera network. It addresses the need for improved economic scalability and aims to enhance resource allocation and revenue distribution mechanisms for topic operators. It also seeks to simplify engineering complexity for dApp developers on the Hedera network.
+
+## **Background and Motivation**
+
+**Problems to Solve:**
+
+* **Dynamic Restriction of Access to Topic Writing:** Users currently face difficulties when creating new topic IDs because they must choose between making the topic publicly writable for a very low cost or implementing a complex web architecture to manage access through the submit key. This complexity can hinder user experience and development efficiency.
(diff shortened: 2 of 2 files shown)


HIP/hip-991.md
```diff
@@ -0,0 +1,221 @@
...
+### **FixedFee**
+
+The `FixedFee` message represents a fixed fee charged for each message submission to a topic.
+
+```
+message FixedFee {
+  // The number of units to assess as a fee
+  int64 amount = 1;
+  // The denomination of the fee; taken as hbar if left unset
+  TokenID denominating_token_id = 2;
```
justin-atwell, line 61: I wonder if we should call this "currency" or just "denomination" since the "token_id" at the end is a bit redundant.

HIP/hip-991.md
```diff
@@ -0,0 +1,221 @@
...
+The `CustomFee` message defines the type of fee and the account receiving the fee assessed during a message submission to the associated topic. A custom fee may only be fixed and must specify a fee collector account to receive the assessed fees.
+
+```
+message CustomFee {
+  oneof fee {
+    // Fixed fee to be charged
+    FixedFee fixed_fee = 1;
+  }
+  // The account to receive the custom fee
+  AccountID fee_collector_account_id = 3;
```
justin-atwell, line 76: should we remove the "account_id" from the end?

HIP/hip-991.md
```diff
@@ -0,0 +1,221 @@
...
+
+**Related Requests:** This enhancement is in response to community feedback (Tier Bot) requesting more versatile and predictable economic models within topic IDs on the Hedera network.
+
+## **Specification**
+
+### **Conceptual Overview**
+
+We propose adding a fixed fee mechanism to the Hedera Consensus Service (HCS) for topic messages, similar to the custom fee structures in the Hedera Token Service (HTS). This will involve creating new Protobuf messages and modifying existing ones to accommodate fixed fees for topics.
+
+### **New Protobuf Messages**
```
Neeharika-Sompalli, line 50: These protobuf messages already exist and can be re-used

HIP/hip-991.md
```diff
@@ -0,0 +1,221 @@
...
+```
+
+### **TransactionRecord**
+
+The `TransactionRecord` message is updated to include the list of custom fees assessed as part of executing the transaction represented by this record.
+
+```
+message TransactionRecord {
+  // All custom fees that were assessed during a message submission
+  repeated AssessedCustomFee assessed_custom_fees = 13;
```
Neeharika-Sompalli, line 115: This message also exist. We should just update the documentation in protobufs.

HIP/hip-991.md
```diff
@@ -0,0 +1,221 @@
...
+### **CustomFee**
+
+The `CustomFee` message defines the type of fee and the account receiving the fee assessed during a message submission to the associated topic. A custom fee may only be fixed and must specify a fee collector account to receive the assessed fees.
+
+```
+message CustomFee {
+  oneof fee {
+    // Fixed fee to be charged
+    FixedFee fixed_fee = 1;
+  }
```
Neeharika-Sompalli, line 74: There is no need for a oneOf, since it is just a fixed fee

HIP/hip-991.md
```diff
@@ -0,0 +1,221 @@
...
+  TokenID denominating_token_id = 2;
+}
+```
+
+### **CustomFee**
+
+The `CustomFee` message defines the type of fee and the account receiving the fee assessed during a message submission to the associated topic. A custom fee may only be fixed and must specify a fee collector account to receive the assessed fees.
+
+```
+message CustomFee {
```
Neeharika-Sompalli, line 70: We should probably rename this to `CustomFixedFee` since `CustomFee` message already exist in protobufs

HIP/hip-991.md
```diff
@@ -0,0 +1,221 @@
...
+We propose adding a fixed fee mechanism to the Hedera Consensus Service (HCS) for topic messages, similar to the custom fee structures in the Hedera Token Service (HTS). This will involve creating new Protobuf messages and modifying existing ones to accommodate fixed fees for topics.
+
+### **New Protobuf Messages**
+
+### **FixedFee**
+
+The `FixedFee` message represents a fixed fee charged for each message submission to a topic.
+
+```
+message FixedFee {
```
Neeharika-Sompalli, line 57: This protobuf message already exist and can be re-used

HIP/hip-991.md
mgarbs: Added the note for this

HIP/hip-991.md
mgarbs: Added the note for this

Neeharika-Sompalli (approved):
LGTM!

mgarbs (approved):

Neurone:
Hi guys, the HIP seems very cool to me. However, we still need to define clear guardrails to prevent potential exploits of the new features.

In particular, I suggest the following changes:

1) Comply with the current security requirements regarding allowance for moving user's funds. In particular, before being able to send paid HCS messages to a topic, the users should previously set an allowance for the recipient topic, the same way they do in the case of HTS, to allow paying for fixed or custom fees.

2) Set a new flag in the `ConsensusSubmitMessageTransactionBody` proto message to allow paying for a message (default is false). If the flag si not set or is not `true`, a tx error will be fired.

3) Update the SDK to include a flag  (i.e. `payForAdditionalFees`) to `true` if the message can pay additional fees. Default: false.

4) Explain that fees can be set only once during the topic creation and cannot be changed in the future.

5) Fee type in the HIP definitions should be `CustomFixedFee`, and not `CustomFee`, for example:

```proto
message TopicCreateTransactionBody {
  // The ID of the topic to be created
  TopicID topic_id = 1;
  // The custom fees to be assessed during a message submission to this topic
  repeated CustomFixedFee custom_fees = 2;
}
```
or

```proto
message TopicInfo {
  // The ID of the topic
  TopicID topic_id = 1;
  // The custom fees to be assessed during a message submission to this topic
  repeated CustomFixedFee custom_fees = 2;
}
```

6) Security implications. Many clients use blind signing or third-party apps to sign transactions, so the workflow that shows the custom fees to the user can not be applicable. For this reason, it's very important that the backend layer explicitly allows for payment of customs fees to show that is exactly what the backend wants to do. This security level can be achieved with a generic allowance to the topic ID and an explicit flag set for each message.

ty-swirldslabs:
> Hi guys, the HIP seems very cool to me. However, we still need to define clear guardrails to prevent potential exploits of the new features.

Hey @Neurone!

Thanks for adding these ideas for guardrails on topics and expenditure of HBAR from fees of topics. I do think these are good ideas that can extend the base functionality, but I don't believe they are imperative enough to be added to this HIP. 

**My main rationale:**

- There is no allowance for Topics for spending from Accounts and that's big enough to be a HIP on it's own after this is implemented. Great idea but I don't believe it is required for this.
- We added this callout on the HIP: when programmatically connecting to topics that have a custom fee key enabled, set max fee on transactions so that if the fee is changed to be higher per message submission, the user isn’t charged more.
- I believe the set max transaction fee takes care of the concern addressed above

**Specific Callouts:**

4. I believe for a topic id fee to be effective, and have an ability to update with inflation and possible costs changes without impacting provenance to a topic, adding the custom fee key makes the most sense as outlined in the HIP. 

6. I agree, and when writing this HIP we decided to rely on the already present set max transaction fee function to enforce backends not overpaying if custom fees change, since there is no prompt. We also want to make sure highlight this approach and inform users as the HIP rolls out.

Neurone:
Hi Ty, for me the HIP cannot be complete if it does not address all security concerns, regardless of worries about the final size of the task.

Currently, the maxTransactionFee parameter that the user can set on an HTS transaction has no effect on the royalties or fixed fees applied to the tokens. And that makes sense, because technically these are not fees you pay to the network, they are payments made between users.

The same applies to the HCS topics, because we are not talking about network or node fees, we are talking about payments between users, and so a security mechanism is mandatory.

Allowance is the low hanging fruit already used with HTS, and it is easy to understand in the EVM world, so I suggest sticking with that instead of inventing a novel security procedure.

4. Ok, I don't like it :D and as a user I wouldn't use a paid topic with fees that can be changed arbitrarily, but I understand that it makes sense to give the option and let the users decide. But I think this makes the pre-emptive permission to the topic and the explicit flag true on the HCS messages even more essential features to include in the HIP.

6. As described above, I don't suggest mixing behaviours. The maxTransactionFees does not apply to fixedFees or royalties for HTS, so I discourage its use in this case. Also, the maxTransactionFee and a general total fee serve different purposes and use cases, with the former being more technical and low-level, often configured on the application or IT environment, and the latter being more manageable at a high-level, without changing the application's configuration, but via Hedera transactions by a dedicated admin/manager.

Neurone:
Hi @ty-swirldslabs, [here](https://github.com/hashgraph/hedera-improvement-proposal/pull/1018) is the final PR to address the topics we discussed. Let me know if it makes sense for you.

PR #1011, opened 2024-07-16
docs(HIP-632): Decouple from virtual addresses; referring only to aliases

david-bakin-sl:
**Description**:

- Removing virtual addresses from HIP-632; referring only to aliases.
- Changed some method signatures to return a response code as well as "actual" result
   - but not `isAuthorizedRaw` unfortunately as it has already been released

### DIFF
--- a/HIP/hip-632.md
+++ b/HIP/hip-632.md
@@ -10,8 +10,8 @@ status: Accepted
 last-call-date-time: 2023-01-10T07:00:00Z
 created: 2022-11-28
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/638
-updated: 2024-05-22
-requires: 631
+updated: 2024-07-15
+requires: 583
 ---
 
 ## Abstract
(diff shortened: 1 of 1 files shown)


mgarbs (approved):

HIP/hip-632.md
```diff
@@ -44,26 +44,46 @@ Additionally, to support standard implicit checks the network should support sys
...
 
-A new system contract with functions to support address (Hedera Account ID and virtual) mapping reconciliation and authorization is proposed.
+A new system contract with functions to support address (Hedera Account ID and EVM address alias) mapping and authorization is proposed.
 
-This will aid developers to support smart contract migration logic for cases where Hedera Account ID or virtual address values are encountered. Additionally, EVM developers who were limited to `ECRECOVER`  authorization flows will be able to expand authorization checks to other key types. 
+This will aid developers to support smart contract migration logic for cases where aliases are encountered.
+Additionally, EVM developers, who were limited to `ECRECOVER` authorization flows, will be able to expand
+authorization checks to other key types. 
+
+HAS methods here can use EVM 20-byte addresses which are either Hedera account num aliases or EVM address
```
Nana-EC, line 53:
```suggestion
HAS methods here can use EVM styled 20-byte addresses which are either Hedera account num aliases or EVM address 
```

HIP/hip-632.md
```diff
@@ -44,26 +44,46 @@ Additionally, to support standard implicit checks the network should support sys
...
-|  | getHederaAddress(address) | address | returns the top level Hedera Account ID if applicable |
-|  | isVirtualAddress(address) | bool | true if valid virtual address, false if long-zero or non existing account |
-|  | isAuthorized(address, messageHash, signatureBlob) | bool | true if account is authorized to carry out transaction execution on account. Accepts protobuf key signature blobs. May be used for ECDSA, ED25519 and complex key flows |
-|  | isAuthorizedRaw(address, messageHash, signatureBlob) | bool | true if account is authorized to carry out transaction execution on account. Accepts single key raw signature blobs (ECDSA and ED25519). This provides similar logic to ECRECOVER. |
+| `0xdea3d081` | getEvmAddressAlias(address) | address | given a Hedera Account ID (account num alias) returns the EVM address alias, if present |
+| `0xbbf12d2e` | getHederaAccountNumAlias(address) | address | given an alias returns the top level Hedera Account ID (as account num alias), if applicable |
+| `0xcd82a665` | isAlias(address) | bool | true if an account num alias or evm address alias), false if non existing account |
```
Nana-EC, line 65: I think this should be isEvmAddressAlias()
```suggestion
| `0xcd82a665` | isEvmAddressAlias(address) | bool | true if a valid evm address alias on the network, false if an account num alias or non existing account |
```

Nana-EC (commented):
Looking good 2 suggestions

HIP/hip-632.md
david-bakin-sl: What we had was `isVirtualAddress` defined as `true if valid virtual address, false if long-zero or non existing account`.  

But we want to do two things at once:  Change the wording `virtual address` to `evm address alias` **and** remove the limitation "can't be a long-zero" which we felt was not a restriction we needed (or that I could find any reason for among us).  

So what we're left with, if we want to go that route, is "is evm address alias or long-zero _== account num alias_ and it must be for an existing account", i.e., is an alias and the account it is an alias for exists.  (Even hollow accounts.). Thus the name.

Do we still want to go in that direction (i.e., _also_ eliminate the restriction that it can't be a long-zero)?

Perhaps call it `isValidAlias` to make it explicit that it must be for a valid (== existing) account.

HIP/hip-632.md
```diff
@@ -132,6 +170,11 @@ Give example of using each function in context to show how they can be used to d
 ## Reference Implementation
 
 
+## (New) Response Codes That Can Be Returned
+
+* `INVALID_ARGUMENT_TO_SYSTEM_CONTRACT_METHOD` - an argument to a system contract method is invalid (e.g., byte array of wrong length)
```
lukelee-sl, line 175: Have we checked that there are not existing response codes that could be used for these purposes.  Just wondering as adding response codes I believe at entails updating the HederaResponseCodes.sol file at the very least.

lukelee-sl (approved):
Generally looks good.  Left a small comment for possible further research.

HIP/hip-632.md
david-bakin-sl: I did check, didn't find anything that looked appropriate.

HIP/hip-632.md
```diff
@@ -10,65 +10,118 @@ status: Accepted
...
-updated: 2022-12-13, 2022-12-20, 2023-01-03, 2023-01-18
-requires: 631
+updated: 2024-07-15
+requires: 583
 ---
 
 ## Abstract
 
-Virtual Addresses introduced a list of ECDSA public addresses that are stored in the account entity and serve as the account identifier on the EVM for transactions. This is in addition to the `Hedera Account ID` that is used by default when no virtual addresses exist.
+Hedera accounts have the capability to have a linked _EVM address alias_, an EVM address.  In this way smart contracts can deal
```
lukelee-sl, line 19: perhaps it'll be helpful to link/refer to the table in HIP 583 that defines the various address formats.

HIP/hip-632.md
```diff
@@ -10,65 +10,118 @@ status: Accepted
...
 
-A new system contract with functions to support address (Hedera Account ID and virtual) mapping reconciliation and authorization is proposed.
+A new system contract with functions to support address (Hedera Account ID and EVM address alias) mapping and authorization is proposed.
+
+This will aid developers to support smart contract migration logic for cases where aliases are encountered.
+Additionally, EVM developers, who were limited to `ECRECOVER` authorization flows, will be able to expand
+authorization checks to other key types. 
 
-This will aid developers to support smart contract migration logic for cases where Hedera Account ID or virtual address values are encountered. Additionally, EVM developers who were limited to `ECRECOVER`  authorization flows will be able to expand authorization checks to other key types. 
+HAS methods here can use EVM-styled 20-byte addresses which are either Hedera account num aliases or EVM address
```
lukelee-sl, line 73: Perhaps define `HAS` before it is used here?
```suggestion
HAS methods can use EVM-styled 20-byte addresses which are either Hedera account num aliases or EVM address 
```

lukelee-sl (approved):
LGTM

HIP/hip-632.md
```diff
@@ -10,65 +10,118 @@ status: Accepted
...
+are not supported by these methods.)
 
 | hash | signature | return | description |
 | --- | --- | --- | --- |
-|  | getVirtualAddresses(address) | address[] | returns an array of virtual addresses for a given Hedera Account ID  |
-|  | getHederaAddress(address) | address | returns the top level Hedera Account ID if applicable |
-|  | isVirtualAddress(address) | bool | true if valid virtual address, false if long-zero or non existing account |
-|  | isAuthorized(address, messageHash, signatureBlob) | bool | true if account is authorized to carry out transaction execution on account. Accepts protobuf key signature blobs. May be used for ECDSA, ED25519 and complex key flows |
-|  | isAuthorizedRaw(address, messageHash, signatureBlob) | bool | true if account is authorized to carry out transaction execution on account. Accepts single key raw signature blobs (ECDSA and ED25519). This provides similar logic to ECRECOVER. |
+| `0xdea3d081` | getEvmAddressAlias(address) | address | given a Hedera Account ID (account num alias) returns the EVM address alias, if present |
```
lukelee-sl, line 83: started implementation and noticed that the return type should probably be `(int64 responseCode, address alias)` to be consistent with the other view functions.  The same for `getHederaAccountNumAlias` below

lukelee-sl (changes requested):
Probably should fix up the return types for `getHederaAccountNumAlias` and `getEvmAddressAlias`

lukelee-sl (dismissed):
LGTM

HIP/hip-632.md
david-bakin-sl: done

HIP/hip-632.md
```diff
@@ -105,7 +160,7 @@ message SignaturePair {
          * ECDSA p-384 signature
          */
         bytes ECDSA_384 = 5;
-
+gi tpush
```
stoyanov-st, line 163: Probably an unintentional addition?

david-bakin-sl: oops!

HIP/hip-632.md
```diff
@@ -10,65 +10,120 @@ status: Accepted
...
 
 ## Motivation
 
-Prior to HIP 631, deployed smart contracts may have stored an account with a Hedera Account ID or the extracted public address from an ECDSA public key.  Subsequent to HIP 631, when a virtual address is added to an account it will now present itself as the virtual address and no longer use the Hedera Account ID.  To avoid issues of lossed balance and permissions it is important that developers can distinguish between Hedera Account IDs and virtual addresses and translate easily between the them.
+Since the same account may be referred to with two different 20-byte identifiers (EVM address alias or Hedera account num alias),
+to avoid issues of loss of balance and permissions it is important that developers can distinguish between Hedera Account IDs 
+and EVM addresses and translate easily between them.
+
+Many smart contracts utilize `ECRECOVER` as part of their authorization process in smart contracts.
+However, this is limited to `ECDSA` based signatures, which leaves accounts that have `E25519` and complex key types, supported by Hedera,
```
stoyanov-st, line 34:
```suggestion
However, this is limited to `ECDSA` based signatures, which leaves accounts that have `ED25519` and complex key types, supported by Hedera,
```

lukelee-sl (approved):
LGTM

stoyanov-st (commented):
Overall LGTM with a small correction

stoyanov-st (dismissed):

lukelee-sl (dismissed):
LGTM

HIP/hip-632.md
```diff
@@ -10,65 +10,121 @@ status: Accepted
...
-requires: 631
+updated: 2024-09-13
+requested-by: David Bakin <@david-bakin-sl>
+requires: 583
 ---
 
 ## Abstract
 
-Virtual Addresses introduced a list of ECDSA public addresses that are stored in the account entity and serve as the account identifier on the EVM for transactions. This is in addition to the `Hedera Account ID` that is used by default when no virtual addresses exist.
+Hedera accounts have the capability to have a linked _EVM address alias_, an EVM address.  In this way smart contracts can deal
```
Nana-EC, line 20: Maybe?
```suggestion
Hedera accounts have the capability to have a linked _EVM address alias_, an EVM address.  In this way the EVM can represent these accounts in an industry standard format
```

Nana-EC (approved):

mgarbs (approved):

PR #1018, opened 2024-07-29
Improvements to HIP-991

Neurone:
**Description**:

- Add: Users need to set allowance to pay for the fees in HBAR or Tokens
- Add: Fee Schedule Key specifications;
- Add: Allowance specifications;
- Add: Details about single features;
- Add: new and updated protobuf messages;
- Update: Updated CustomFixedFee (it already exists in the SDKs as a specification of the CustomFees) to ConsensusCustomFee
- Update: Paying with tokens is fully implemented with the current HIP
- Update: Users don't need to send HBAR while submitting the message
- Update: Restructured the HIP sections to comply with HIP-000;
- Update: While the gRPC service must be updated to support the new behaviours, the service interface changes accordingly to the new protobuf messages
- Fix: This HIP does not change the TransactionRecord message definition
- Fix: Names and references of som protobuf messages;
- Fix: Source code samples;
- Fix: Make linter happy;
- Fix: Fee Schedule Key can now send free messages, in addition to Submit Key;
- Fix: Reference implementation pseudo source code;

**Related issue(s)**:

Fixes # N/A

**Notes for reviewer**: 

N/A

**Checklist**

### DIFF
--- a/HIP/hip-991.md
+++ b/HIP/hip-991.md
@@ -10,7 +10,7 @@ status: Council Review
 last-call-date-time: 2024-07-24T07:00:00Z
 created: 2024-06-14
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/pull/991
-updated: 2024-07-30
+updated: 2024-08-05
 ---
 
 ## Abstract
@@ -71,7 +71,9 @@ We propose adding a fixed fee mechanism to the Hedera Consensus Service (HCS) fo
 
 * Topics can have a Fee Schedule Key set at creation time.
 * Fee Schedule Key can manage fee updates and send free messages to the topic.
-* The Fee Schedule Key can be updated with an update topic transaction signed by the Admin Key and the new Fee Schedule Key.
+* Fee Schedule Key follows the same "Lower privilege key" rules introduced by [HIP-540](https://hips.hedera.com/hip/hip-540), that means:
+  * Only the admin key should be able to remove the Fee Schedule Key.
+  * Fee Schedule Key can change itself to another valid or unusable key.
 * If the topic was created without a Fee Schedule Key, the key cannot be added later.
 * Topics can have fees for submitting messages.
 * Fees can be set at creation time.
--- a/HIP/hip-991.md
+++ b/HIP/hip-991.md
@@ -13,11 +13,11 @@ discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/pull/99
 updated: 2024-07-24
 ---
 
-## **Abstract**
+## Abstract
 
 This document outlines the development of a fixed fee system for the submission of topic messages on the Hedera network. It addresses the need for improved economic scalability and aims to enhance resource allocation and revenue distribution mechanisms for topic operators. It also seeks to simplify engineering complexity for dApp developers on the Hedera network.
 
-## **Background and Motivation**
+## Background and Motivation
 
 **Problems to Solve:**
 
@@ -25,8 +25,6 @@ This document outlines the development of a fixed fee system for the submission 
 * **Ease of Monetization:** Developers and businesses find it challenging to charge users for submitting information to a topic due to the lack of a native fee structure. A built-in payment system would streamline monetization, allowing developers to easily charge for topic submissions, which can enhance the financial viability of their applications.
 * **Resource Allocation:** The current low fee of $0.0001 per message makes it difficult for topic operators to manage and predict the economic impact of high-traffic topics. Users can suffer from spam and unexpected messages that operators have to handle, potentially degrading the quality of service.
 * **Token and HBAR Utilization:** There is an opportunity to expand the utility of various tokens within the Hedera ecosystem. By allowing fees to be paid in tokens other than HBAR, such as those required by DAOs, users can leverage their token holdings for topic submissions, enhancing the ecosystem's flexibility and user engagement.
-
----
 
 **Feature Summary:**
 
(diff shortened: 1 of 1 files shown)


kantorcodes (approved):

HIP/hip-991.md
```diff
@@ -44,182 +44,310 @@ This document outlines the development of a fixed fee system for the submission
...
-  FixedFee fixed_fee = 1;
+### Requirements and Scope
 
-  // The account to receive the custom fee
-  AccountID fee_collector_account_id = 3;
-}
-```
+* Topics can have a Fee Schedule Key set at creation time.
+* Fee Schedule Key can manage fee updates and send free messages to the topic.
+* The Fee Schedule Key can be updated with an update topic transaction signed by the Admin Key and the new Fee Schedule Key.
```
ty-swirldslabs, line 74: Should we add the ability for the fee schedule key to update itself? I think a HIP went through recently that gave lower level privilege keys that ability

HIP/hip-991.md
```diff
@@ -44,182 +44,310 @@ This document outlines the development of a fixed fee system for the submission
...
+* If the topic was created without a Fee Schedule Key, the key cannot be added later.
+* Topics can have fees for submitting messages.
+* Fees can be set at creation time.
+* Fees can be changed (updated or removed) from a topic with an update topic transaction signed by the Fee Schedule Key.
+* Fees are defined as a list of custom fee.
+* A custom fee can be set in HBAR or HTS fungible tokens, and must have an accountID as a receiver.
+* Fee payer must set an allowance in HBAR or HTS tokens to be used to pay fees for message submissions, on an HCS topic basis.
```
ty-swirldslabs, line 81: These fees are per message allowance and not total allowance. Don't know if we need to be overly specific here but maybe that's an important callout as it's different than allowances today.

HIP/hip-991.md
```diff
@@ -44,182 +44,310 @@ This document outlines the development of a fixed fee system for the submission
...
+* Allowance granted to topic ID does not allow the topic's Admin Key (or any of the topic's keys) to spend that allowance.
+* Fee payer must set can set a maximum fee per message, on an HCS topic basis.
```
ty-swirldslabs, line 83: How does this interact with allowance / is it a max per message allowance, or a total allowance, and a max fee per message?

HIP/hip-991.md
```diff
@@ -44,182 +44,310 @@ This document outlines the development of a fixed fee system for the submission
...
+* Messages signed by the Fee Schedule Key or by a Submit Key don't incur fees.
```
ty-swirldslabs, line 84: Should the fee schedule key be signing transactions of messages? I think only the submit key, and if the account is the collector of the fee / or the operator of the topic should they be exempt.

HIP/hip-991.md
```diff
@@ -44,182 +44,310 @@ This document outlines the development of a fixed fee system for the submission
...
+* Applications must set the `allowCustomFeesPayment` flag to true.
```
ty-swirldslabs, line 85: Applications or the account that's submitting the messages?

HIP/hip-991.md
```diff
@@ -44,182 +44,310 @@ This document outlines the development of a fixed fee system for the submission
...
+* Default value of the `allowCustomFeesPayment` flag is false.
+* If the `allowCustomFeesPayment` flag is not set, or is not true, a transaction error will be fired.
 
-### **Modified Protobuf Messages**
+### User Flows and Interaction
 
-### **TopicCreateTransactionBody**
+* Users will specify the fee settings during the topic creation process through a simple interface in their Hedera client (refer to the creation of token custom fees/fixed fee for reference).
+* Before submitting a message to a topic through an application or wallet interface, users must set an allowance and a maximum fee per message.
```
ty-swirldslabs, line 92: I'm not sure if allowance should be required. I think per message fee makes sense, but in most cases where a user is sending 1 message through a wallet, requiring a max allowance, and a per message fee seems like high friction. 

I like having alowance to cap teh total amount for backend systems, but I'm not sure it should be required to be set to send messages to the topic.

HIP/hip-991.md
```diff
@@ -44,182 +44,310 @@ This document outlines the development of a fixed fee system for the submission
...
-The introduction of custom fees adds an additional layer of economic control but also introduces potential vectors for abuse, such as fee manipulation.
+## How to Teach This
 
+TBD
 
-### **Backwards Compatibility**
+## Reference Implementation
 
-There are no known backwards compatibility issues. Existing topics without custom fees will continue to function as they currently do. New topics with custom fees will require appropriate updates to the Hedera SDKs, mirror nodes, and applications to support the new fee structures.
+Below is a pseudo-code example of creating a topic with a fixed fee, setting the allowance, and sending a message.
```
ty-swirldslabs, line 297: love the example code!

ty-swirldslabs (changes requested):
I have a couple questions and comments but in all this looks very good!

HIP/hip-991.md
Neurone: See above, the suggestion is to have both options available.

HIP/hip-991.md
Neurone: Side note: in case you don't care about an overall max, you are still forced to use a very high max total allowance like users do now when they don't care about a real limit. The other option would have been to introduce something like a negative number for infinite, but it would have been an additional complexity to implement and explain, and a different behaviour vs the common practice.

HIP/hip-991.md
Neurone: For applications I mean the app that uses the SDK and write the code to prepare the transaction. It needs to set the allowance clearly in the SDK, so you "declare" to the SDK you know what you are doing.

In the example is this last part:

```JavaScript
[...]
// Send message to the topic
transactionResponse = await new TopicMessageSubmitTransaction({ topicId: topicId, message: "Hello, HCS!" })
  .setAllowCustomFeesPayment(true) // Make the agreement on payment explicit at the application level. Useful for any backend / non-front-user-facing application.
  .execute(client);
```

HIP/hip-991.md
Neurone: The idea is, because you need to send a tx to set a max per message in any case, you can send at the same time a max overall, and the UX does not change (you don't need to send an additional transaction). If you don't care about a max overall, you can just put a very high max, but if you care, you can set it and leverage also that limit.

HIP/hip-991.md
Neurone: Yep, true, that makes sense! It was the [HIP-540](https://hips.hedera.com/hip/hip-540) that added the following rules to HTS:

> 1. Only the admin key should be able to remove itself or other keys.
> 2. All keys can change themselves to another valid or unusable key (such as all-zeros). Except for the admin key - we do not support setting an unusable admin key. This is because removing the admin key and making it unusable are functionally identical. It should simply be removed if no longer desired.

I'll send an update about this point and add a reference to the HIP-540.

Neurone: The text now reflects the change, let me know if it makes sense.

HIP/hip-991.md
Neurone: Yeah, that was my initial thought also, but it doesn't work well.

The point is that if a topic has a submit key, only the submit key can sign valid message transactions for that topic. Nobody else can send messages on the topic.

So, if you don't set a supply key, everyone can send messages to the topic, but now you no longer have a "free" messages key.

Because I assume the latter is the use case the HIP wants to address, I introduced the free messages for the Fee Schedule Key.

What do you think?

mgarbs (dismissed):

HIP/hip-991.md
ty-swirldslabs: Gotcha, that is a great point. For clarity, the operator who creates the topic and the fee collectors  on the custom fee will NOT be charged, so you would still have a free submission from the creator of the topic. 

But you would not have an ability to adhoc sign transactions from other accounts you may also want to remove fees from. I think there's enough of a use case for the need of the custom fee key to be used 

I agree, allowing the fee schedule key to sign to not charge fees is a good feature

HIP/hip-991.md
ty-swirldslabs: Knowing it can be executed in the same transaction is very helpful. 

If we make sure that default value is auto-injected in the SDKs if it's not defined then I'm ok with it. I think there's a lot of use cases where people will not want to set a max allowance.

Neurone: Yep, that can be achieved by the developers because they can set that default values as they prefer for their users.

But the SDK always starts with the safest default values for any parameters (so in this case 0 max allowance, 0 max allowance per message, no owner, no spender) so developers don't unintentionally harm their users.

HIP/hip-991.md
Neurone: Good! 👍

Another topic: the HIP does not currently include the creators and fee collectors in the list of signers for free messages. We need to specify this clearly if we want it to.

**Creator**

Technically, we don't have that concept of the token's creator, but a topic can have an admin key (if it's set at creation time), so maybe we can add the admin key to the list of valid signers for free messages.

What do you think?

**Fee collectors**

Regarding fee collectors, I suggest reconsidering having them as free message signers for some reason, but in particular because I assume fee collectors must remain free not to trust each other.

Scenario
- Fees are set as 1 TokenA to Alice and 2 TokenB to Bob.
- Bob and Alice do not necessarily like or trust each other. Or they do now, but not in the future.
- Bob still wants his part of the fees when Alice sends messages.
- Alice can still technically send free messages for her part (because her balance in TokenA will not change), but she still needs to pay Bob's cut of the pie with TokenB.
- The scenario seems reasonable and predictable.

If we introduce free messages for any fee collectors in general, Alice can fool Bob, for example, by trading his fees OTC to her users: Bob will never see payments collected.

Even in a non-fraud scenario, Alice might be able to afford to send X number of messages without collecting the fee (i.e., she wants to promote her application), but this may not be the case for Bob.

I think fee receivers are a different concept from free message senders, so if we still feel there's a business need for a free message key list, I suggest keeping them separated.

We can introduce one explicit list, which only the admin key can update. 

In this way, as an admin, you gain the ability to make messages free for some users, and you make the rules check simpler and clearer.

Because the list can be updated, the free messages for specific users can be set for a limited time, so it would be more flexible instead of being assigned by your role.

Finally, this approach keeps the trust assumptions the same because Alice and Bob need to constantly trust the admin in any case to set their fee correctly.

What do you think?

ty-swirldslabs:
Giuseppe and I had a great meeting and wanted to post back the takeaways from that to the PR. HEre's three points we discussed

1.  setAllowanceCustomFeePayment set on the message submission transaction:
- No changes to the submission of a message to a topic
    - EXCEPT in the SDK with the security flag
    - setAllowanceCustomFeePayment = true to allow transactions to go through as an extra safety mechanism

- setAllowanceCustomFeePayment
    - This would only be in the SDK
    - Additional security mechanism
        - Not something you see on the network level
    - There’s two transactions for users:
        - 1. Set the allowance to the topic id
        - 2. Send the message to the topic id

2. What should the default values of the SDK of the network
- Default of the SDK - usually set at the safest default  
- When you set the allowance, maximum is default the highest value possible
    - To set the maximum set it to a number
    - Default is maximum in SDK
        - Huge callout in documentation that says, set maximum if you’re concerned about drainage 
        - If you don’t set anything, it’s an infinite number

3. Accounts that can submit messages for free:
    - Add a account public key list on the topic id main level that can submit free messages
    - Decouple the fee collectors from the free submission
    - Admin keys are the only thing that can change these account ids after creation
    - Discuss with engineers on what the maximum should be for the free account id array

Neurone:
I sent the updates to the PR to include the above discussion. I'm going to merge as soon as the PR is reviewed.

mgarbs (approved):

PR #1021, opened 2024-08-01
feat: create hip-1021

kantorcodes:
## Description

This pull request introduces HIP-1021, titled "Improve Assignment of Auto-Renew Account ID for Topics." The proposal aims to enhance the Hedera Consensus Service (HCS) by automating the assignment of the `autoRenewAccountId` to the account that creates a `TopicCreateTransaction`. This feature is particularly useful when an Admin Key is not present during topic creation, ensuring that the auto-renewal mechanism is consistently applied. 

The key changes include:
- **Automatic Assignment:** The `autoRenewAccountId` will automatically be set to the `payer_account_id` of the transaction if an Admin Key is not provided during topic creation.
- **Retroactive Application:** For existing topics that have only a Submit Key and no `autoRenewAccountId`, the original `payer_account_id` will be retroactively set as the `autoRenewAccountId`.
- **Consistency and Error Reduction:** This change aims to prevent user errors and ensure consistent management of topic renewals.

## Improvements

1. **Automatic Setting of Renewal Account:**
   - The account that creates a topic will now automatically be used for renewals. Users don't need to manually set this anymore, making the process simpler and reducing the chance of mistakes.

2. **Uniform Handling of Old Topics:**
   - For older topics without a set renewal account, the system will automatically use the original account that created them for rent. This makes renewals consistent across all topics.

3. **Fewer Mistakes:**
   - By automating the setup of the renewal account, there’s a lower risk of topics accidentally expiring due to missing settings.

4. **Easier for Users:**
   - Users no longer need to worry about setting up renewal accounts; the system handles it automatically, making the overall process more user-friendly.

5. **Ready for Future Changes:**
   - This change also prepares the system for potential future updates, like introducing rent for topics, ensuring everything is correctly configured in advance.

### DIFF
--- a/HIP/hip-1021.md
+++ b/HIP/hip-1021.md
@@ -4,11 +4,12 @@ title: Improve Assignment of Auto-Renew Account ID for Topics
 author: Michael Kantor (@kantorcodes)
 type: Standards Track
 category: Service
-status: Draft
+status: Review
 created: 2024-08-01
 requested-by: TierBot
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/pull/1021
 needs-council-approval: Yes
+updated: 2024-12-04
 ---
 
 ## Abstract
--- a/HIP/hip-1021.md
+++ b/HIP/hip-1021.md
@@ -4,7 +4,9 @@ title: Improve Assignment of Auto-Renew Account ID for Topics
 author: Michael Kantor (@kantorcodes)
 type: Standards Track
 category: Service
-status: Review
+needs-tsc-approval: Yes
+status: Last Call
+last-call-date-time: 2024-12-18T07:00:00Z
 created: 2024-08-01
 requested-by: TierBot
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/pull/1021
--- a/HIP/hip-1021.md
+++ b/HIP/hip-1021.md
@@ -1,5 +1,5 @@
 ---
-hip: 1020
+hip: 1021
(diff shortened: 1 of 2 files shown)


HIP/hip-1021.md
```diff
@@ -0,0 +1,65 @@
...
+requested-by: TierBot
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/pull/1021
+needs-council-approval: Yes
+---
+
+## Abstract
+
+This HIP proposes a modification to the Hedera Consensus Service (HCS) to:
+ - Enable setting the `autoRenewAccountId` to the account creating a `TopicCreateTransaction` when an Admin Key is not present during Topic Creation.
+ - It ensures that for all existing topics with only a Submit Key and no `autoRenewAccountId`, the original `payer_account_id` of the topic is retroactively set as the `autoRenewAccountId`.
```
netopyr, line 18: We should not modify existing topics IMO:

- We usually do not modify existing entities. A user may have left autoRenewAccount empty intentionally and disagrees with this change.
- This would have to be initiated outside of the consensus node because the consensus node does not store the original `payer_account_id`.
- It is unclear who should pay for these transactions.

HIP/hip-1021.md
```diff
@@ -0,0 +1,65 @@
...
+ - Automatically set the `autoRenewAccountId` to the `payer_account_id` of the Topic Id to prevent unintended user error.
```
netopyr, line 19: How would I be able to leave this field empty intentionally?

mgarbs (approved):

PR #1028, opened 2024-08-16
HIP-1028: Metadata management via SmartContracts

mustafauzunn:
**Description**:

This PR describes the changes needed to support the metadata updates via smart contracts.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-1028.md
+++ b/HIP/hip-1028.md
@@ -6,8 +6,8 @@ working-group: Valentin Valkanov <valentin.valkanov@limechain.tech>, David Bakin
 type: Standards Track
 category: Service
 needs-council-approval: Yes
-status: Review
-last-call-date-time:
+status: Last Call
+last-call-date-time: 2024-09-10T07:00:00Z
 created: 2024-08-15
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/pull/1028
 requires: 646, 657, 765
--- a/HIP/hip-1028.md
+++ b/HIP/hip-1028.md
@@ -2,7 +2,7 @@
 hip: 1028
 title: Metadata management via SmartContracts
 author: Mustafa Uzun <mustafa.uzun@limechain.tech>, Luke Lee <luke.lee@swirldslabs.com>
-working-group: Valentin Valkanov <valentin.valkanov@limechain.tech>, David Bakin <david.bakin@swirldsbals.com>
+working-group: Valentin Valkanov <valentin.valkanov@limechain.tech>, David Bakin <david.bakin@swirldslabs.com>
 type: Standards Track
 category: Service
 needs-council-approval: Yes
@@ -17,9 +17,9 @@ requested-by: NA
 
 ## Abstract
 
-HIP-646/657/765 introduces a new metadata field for Fungible Tokens (FT), Non-Fungible Tokens (NFT) and METADATA key  
+HIP-646/657/765 introduced a new metadata field for Fungible Tokens (FT), Non-Fungible Tokens (NFT) and METADATA key  
 for updating metadata. However, these features are not supported by smart contracts.  
-This proposal aims to enhance Hedera Smart Contract Service (HSCS) by exposing HIP-646/657/765 features.
+This proposal aims to enhance the Hedera Smart Contract Service (HSCS) by exposing HIP-646/657/765 features.
 
 ## Motivation
 
@@ -63,7 +63,7 @@ The following structs have been added to support metadata and metadata key
 struct HederaTokenV2 {
     String name;
     String symbol;
-    // other fields are like in HederaToken
+    // other fields identical to the existing HederaToken struct
     ...
     // The new field for metadata
     bytes metadata;
@@ -74,7 +74,7 @@ struct HederaTokenV2 {
 struct TokenInfoV2 {
     // The new struct with metadata
     HederaTokenV2 token;
-    // other fields are like in TokenInfo
+    // other fields identical to the existing TokenInfo struct
     ...
 }
 ```
@@ -91,7 +91,7 @@ struct FungibleTokenInfoV2 {
 struct NonFungibleTokenInfoV2 {
     // The new struct with metadata
     TokenInfoV2 token;
-    // other fields are like in NonFungibleTokenInfo
+    // other fields identical to the existing NonFungibleTokenInfo struct
     ...
 }
 ```
--- a/HIP/hip-1028.md
+++ b/HIP/hip-1028.md
@@ -9,9 +9,10 @@ needs-council-approval: Yes
 status: Draft
 last-call-date-time:
 created: 2024-08-15
-discussions-to:
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/pull/1028
 requires: 646, 657, 765
-updated: 2024-08-15
+updated: 2024-08-19
+requested-by: NA
 ---
 
 ## Abstract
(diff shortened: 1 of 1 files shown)


HIP/hip-1028.md
```diff
@@ -0,0 +1,157 @@
...
+## Motivation
+
+A user may need to create, update and get info for Fungible Tokens (FT), Non-Fungible Tokens (NFT) and METADATA key through contract.  
+The current version of Hedera Smart Contract Service (HSCS) does not cover these use cases.
+
+## Rationale
+
+The HSCS already has the functionality to create and update tokens, so we would want to extend these.  
+
+However, we will add new functions for getting info for tokens to not break the older version of ``HederaTokenService.sol``
```
david-bakin-sl, line 33: These two lines don't "read" well together - first suggests extending the functionality but "functionality" could be confused with meaning the actual methods.  Second line does say new methods will be added.

Some kind of rewording might help.

Let's see ... just an idea ...

```
The HAPI calls for creating tokens and updating token information were modified 
to include the new metadata and metadata key fields in their existing protobuf 
structures - a backward compatible change allowed by protobufs.  But the 
Solidity language doesn't work that way: structs can't inherit from other structs, 
and changing a struct changes the method signature and thus the method selector.

Instead of replacing the existing methods this proposal adds _overloads_ for the
setting methods - methods with the same name but different signatures.  Because
the return type of a Solidity method does _not_ change the signature, the get info
methods have different names.
```

But this is just a suggestion of what could be said (and I'm not very happy with it myself).

HIP/hip-1028.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Abstract
+
+HIP-646/657/765 introduces a new metadata field for Fungible Tokens (FT), Non-Fungible Tokens (NFT) and METADATA key  
+for updating metadata. However, these features are not supported by smart contracts.  
+This proposal aims to enhance Hedera Smart Contract Service (HSCS) by exposing HIP-646/657/765 features.
+
+## Motivation
+
+A user may need to create, update and get info for Fungible Tokens (FT), Non-Fungible Tokens (NFT) and METADATA key through contract.
```
david-bakin-sl, line 26: Specifically mention "token metadata and the metadata key"?

HIP/hip-1028.md
```diff
@@ -0,0 +1,157 @@
...
+3. As an EOA I would like to initiate a smart contract transaction that updates a fungible token with metadata.
+4. As an EOA I would like to initiate a smart contract transaction that updates a non-fungible token collection with metadata.
+5. As an EOA I would like to initiate a smart contract transaction that creates a fungible token with a metadata key.
+6. As an EOA I would like to initiate a smart contract transaction that creates a non-fungible token collection with metadata key.
+7. As an EOA I would like to initiate a smart contract transaction that updates a fungible token with metadata key.
+8. As an EOA I would like to initiate a smart contract transaction that updates a non-fungible token collection with metadata key.
+9. As an EOA I would like to initiate a smart contract transaction that creates a fungible token with metadata and metadata key.
+10. As an EOA I would like to initiate a smart contract transaction that creates a non-fungible token collection with metadata and metadata key.
+11. As an EOA I would like to initiate a smart contract transaction that updates the metadata of NFT(s).
```
david-bakin-sl, line 49: "of NFT(s)" -> "of specific unique instances of an NFT" ??

HIP/hip-1028.md
```diff
@@ -0,0 +1,157 @@
...
+2. As an EOA I would like to initiate a smart contract transaction that creates a non-fungible token collection with metadata.
+3. As an EOA I would like to initiate a smart contract transaction that updates a fungible token with metadata.
+4. As an EOA I would like to initiate a smart contract transaction that updates a non-fungible token collection with metadata.
+5. As an EOA I would like to initiate a smart contract transaction that creates a fungible token with a metadata key.
+6. As an EOA I would like to initiate a smart contract transaction that creates a non-fungible token collection with metadata key.
+7. As an EOA I would like to initiate a smart contract transaction that updates a fungible token with metadata key.
+8. As an EOA I would like to initiate a smart contract transaction that updates a non-fungible token collection with metadata key.
+9. As an EOA I would like to initiate a smart contract transaction that creates a fungible token with metadata and metadata key.
+10. As an EOA I would like to initiate a smart contract transaction that creates a non-fungible token collection with metadata and metadata key.
```
david-bakin-sl, line 48: I know it is the practice in our existing HIPs to just spell out all these variations in a linear list like this, but especially in this particular case I think it could be done differently.  And better:  Right now this list 1..10 looks to me like "twisty passages all alike".  When it is really a 2x2x2 table with an extra 2 that don't quite fit (9&10).

Would it be too different to start with something like this table and then make it better?  Would that be easier to understand all the variations?  It could even be more compact by coalescing Create+Update columns (use `C` and `U` as the column markers) and Fungible+Non-Fungible columns (use `F` and `N` as the markers?).  Or some other kind of table?

|Create|Update|Fungible|Non-Fungible||Metadata?|Metadata Key?|
|----|----|----|----|----|----|----|
| ◉ || ◉ ||| ✅ ||
| ◉ || ◉ |||| ✅ |
| ◉ || ◉ ||| ✅ | ✅ |
| ◉ ||| ◉ || ✅ ||
| ◉ ||| ◉ ||| ✅ |
| ◉ ||| ◉ || ✅ | ✅ |
|| ◉ | ◉ ||| ✅ ||
|| ◉ | ◉ |||| ✅ |
|| ◉ || ◉ || ✅ ||
|| ◉ || ◉ ||| ✅ |

HIP/hip-1028.md
```diff
@@ -0,0 +1,157 @@
...
+11. As an EOA I would like to initiate a smart contract transaction that updates the metadata of NFT(s).
+
+## Specification
+
+We will utilize the existing Hedera Token Service (HTS) HAPI transactions.  
+To achieve this the Hedera Token Service (HTS) system contract must be updated to expose the necessary features.  
+Existing system contract functions must be extended and new functions must be added to the `IHederaTokenService` interface
```
david-bakin-sl, line 55: Above mention is made of `HederaTokenService.sol`, here `IHederaTokenService.sol`.  Perhaps here should _also_ be mentioned `HederaTokenService.sol`?  (Maybe as a footnote too a reminder to the reader of what each one is _for_.)

HIP/hip-1028.md
```diff
@@ -0,0 +1,157 @@
...
+## Specification
+
+We will utilize the existing Hedera Token Service (HTS) HAPI transactions.  
+To achieve this the Hedera Token Service (HTS) system contract must be updated to expose the necessary features.  
+Existing system contract functions must be extended and new functions must be added to the `IHederaTokenService` interface  
+to support the metadata field and key.
+
+### System contract extension
+
+The `IHederaTokenService.sol` is further extended with new types and function signatures that the system contract will respond to.
```
david-bakin-sl, line 60: This seems redundant with "specification" above.  Or if it _is_ the specification, needs more detail.  E.g., the actual Solidity code (as for "New Solidity Types" below)?  Or maybe it is redundant with the "Solidity Function Signatures" section below?

HIP/hip-1028.md
```diff
@@ -0,0 +1,157 @@
...
+}
+```
+
+### Solidity Function Signatures
+
+The new structs will change the signature of the functions. However the old versions of the functions will continue working.
+
+| Hash | Selector                                                                                                                                                                      | Return                                                                     |
+| --- |-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------|
+|0fcaca1f| `updateNFTsMetadata(address token, int64[] memory serialNumbers, bytes memory metadata)`                                                                                      | `int responsecode`                                                         |
```
david-bakin-sl, line 131: Move to end to match order of use cases?  It's kind of the odd man out.  Another approach is to split this table into three lists (separate section headers or maybe not, just three separate tables: getters, setters, and then the update uniques method.

HIP/hip-1028.md
```diff
@@ -0,0 +1,157 @@
...
+|54c832a5| `updateTokenInfo(address token, HederaTokenV2 memory tokenInfo)`                                                                                                              | `int responseCode`                                                         |
+|bc03816f| `getTokenInfoV2(address token)`                                                                                                                                               | `(int64 responseCode, TokenInfoV2 memory tokenInfo)`                       |   
+|fb29ac6e| `getNonFungibleTokenInfoV2(address token, int64 serialNumber)`                                                                                                                | `(int64 responseCode, NonFungibleTokenInfoV2 memory nonFungibleTokenInfo)` |
+|3f9dc353| `getFungibleTokenInfoV2(address token)`                                                                                                                                       | `(int64 responseCode, FungibleTokenInfoV2 memory fungibleTokenInfo)`       |
+
+## Backwards Compatibility
+Backward compatibility is ensured by adding new functions where needed.
```
david-bakin-sl, line 142: Backwards compatibilty is by the consensus node supporting both the existing and the new methods (they've got different selectors).

HIP/hip-1028.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Security Implications
+
+The HIP provides exposure to existing HTS functionality. Fundamentally security is still governed by the ledger security logic  
+in combination with the EVM. As such, there should be no security implications.
```
david-bakin-sl, line 147: "no security implications" -> "no new security implications".

david-bakin-sl (commented):
Quick look, thus just a "comment" at this point.

HIP/hip-1028.md
lukelee-sl: How about including both?  I like the table in the context of the linear list of user stories.

HIP/hip-1028.md
lukelee-sl: suggestion: through contract -> from a smart contract.

HIP/hip-1028.md
```diff
@@ -0,0 +1,155 @@
+---
+hip: 1028
+title: Metadata management via SmartContracts
+author: Mustafa Uzun <mustafa.uzun@limechain.tech>, Luke Lee <luke.lee@swirldslabs.com>
+working-group: Valentin Valkanov <valentin.valkanov@limechain.tech>, David Bakin <david.bakin@swirldsbals.com>
```
lukelee-sl, line 5: swirldslabs.com typo

HIP/hip-1028.md
```diff
@@ -0,0 +1,155 @@
...
+requires: 646, 657, 765
+updated: 2024-08-20
+requested-by: NA
+---
+
+## Abstract
+
+HIP-646/657/765 introduces a new metadata field for Fungible Tokens (FT), Non-Fungible Tokens (NFT) and METADATA key  
+for updating metadata. However, these features are not supported by smart contracts.  
+This proposal aims to enhance Hedera Smart Contract Service (HSCS) by exposing HIP-646/657/765 features.
```
lukelee-sl, line 22: enhance -> enhance the

HIP/hip-1028.md
```diff
@@ -0,0 +1,155 @@
...
+created: 2024-08-15
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/pull/1028
+requires: 646, 657, 765
+updated: 2024-08-20
+requested-by: NA
+---
+
+## Abstract
+
+HIP-646/657/765 introduces a new metadata field for Fungible Tokens (FT), Non-Fungible Tokens (NFT) and METADATA key
```
lukelee-sl, line 20: introduces -> introduced

HIP/hip-1028.md
```diff
@@ -0,0 +1,155 @@
...
+
+### New Solidity Types
+
+The following structs have been added to support metadata and metadata key
+
+```solidity
+struct HederaTokenV2 {
+    String name;
+    String symbol;
+    // other fields are like in HederaToken
```
lukelee-sl, line 66:
```suggestion
    // other fields are like in HederaToken
```
// other fields identical to the existing HederaToken struct

HIP/hip-1028.md
```diff
@@ -0,0 +1,155 @@
...
+```solidity
+struct NonFungibleTokenInfoV2 {
+    // The new struct with metadata
+    TokenInfoV2 token;
+    // other fields are like in NonFungibleTokenInfo
+    ...
+}
+```
+Adding a comment to the TokenKey struct to clarify that the 7th bit is designated for the metadata key.  
+However, existing methods that use `HederaToken` with the `TokenKey` struct will continue to ignore the metadataKey bit.
```
lukelee-sl, line 99: Thanks for adding this

mgarbs (approved):

PR #1037, opened 2024-08-31
Create HIP for Protocol Buffer Specification

jsync-swirlds:
This pull request introduces HIP-1035, titled "Protocol Buffer API Specification" The proposal aims to enhance the Hedera API declarations by adding detailed API specification text to every protocol buffer element. This API specification text is added in a manner very similar to the Java API specification by adding properly formatted comments above each API message and field. The standard protoc plugin protoc-gen-doc processes these comments, in combination with an included template, to produce markdown files that fully detail the API.


HIP/hip-1037.md
```diff
@@ -0,0 +1,95 @@
...
+  to Hedera-led software.
+
+## Specification
+This HIP makes sweeping changes to specification text for the existing API
+documents. These changes affect over 20,000 lines of text across almost 150
+files, which is excessive for a document such as this HIP. The full
+specification, therefore, is only linked here for brevity and clarity.<br/>
+The specification changes intended for this HIP are detailed in github as a
+[pull request](https://github.com/hashgraph/hedera-protobufs/pull/388)
+for the hedera-protobufs repository.
```
david-bakin-sl, line 65: Yes, for the actual improved specifications for each service/message/field in the protobufs - but I think it would be appropriate to put here in this HIP the _standards_ that have been used in those specifications in that PR which we intend to require all future protobuf additions/changes to use!

I'm not speaking of the standards we're going to use in _defining_ the protobufs themselves, (e.g., the section [_Content Guidelines_](https://github.com/hashgraph/hedera-protobufs/blob/02b1569675efa79db01c279d84b3295d23c0c810/documents/Specification-Format-Style-Guidelines.md?plain=1#L88) with, e.g., the preference of unsigned vs signed integer fields) but rather the stuff in [_Style Guidelines for "specification" comments_](https://github.com/hashgraph/hedera-protobufs/blob/02b1569675efa79db01c279d84b3295d23c0c810/documents/Specification-Format-Style-Guidelines.md?plain=1#L3) (where it says, e.g., "be prescriptive", all messages need a description but not all fields do, etc.).

* And the above with the associated examples given there in the PR.

In fact, I'll go so far as to suggest that the direction of this HIP be changed to _be_ the specification of how we'll both specify and document the specification of all protobufs in the future.  To contain _both_ of those sections I mentioned above.  And relegate the [PR itself](https://github.com/hashgraph/hedera-protobufs/pull/388) to not "be the HIP, just in a remote location" but instead to be the "reference implementation". That would seem to be the more appropriate use of the intent and process of HIPs (especially as the PR doesn't actually change anything in the protobufs themselves, only **greatly** improves the documentation).

And if that suggestion was taken then this HIP could also contain (eventually, not necessarily now) our _other_ requirements for protobufs - e.g., that we follow rules such that the encoding can be deterministic and that in fact the PBJ protobuf compiler must be used (if that's the case); this information is important for the user story of the "independent implementator" above and thus should be documented in this HIP.  (I don't think it is anywhere else.). Also how we _version_ protobufs for backward/forward compatibiility (I think we're just using standard protobuf "best practices" for that but even if so we could document them _here_ instead of relying on developers to just know them.)

Finally, mention should be made here in this "Specification" section that we're using protoc-gen-doc style comments.  It's mentioned above in the "Abstract" but the abstract is not the specification part of the HIP and in any event it is mentioned tentatively up there ("we should").

HIP/hip-1037.md
```diff
@@ -0,0 +1,95 @@
...
+The Hedera API is a critical component of the design and interaction for the
+Hedera ecosystem. Improvements to the formality and completeness of the API
+specification help everyone involved to develop distributed applications
+more quickly and with greater confidence.
+
+## Reference Implementation
+Not applicable
+
+## Rejected Ideas
+Not applicable
```
david-bakin-sl, line 85: Was there any consideration of some formal specification notation (e.g., Z, UML, SPIN) that was rejected for some reason (e.g., impractical at this time)?  Or of introducing any notation beyond markdown to the way Javadoc does (e.g., `@param`, `@return`)?  If so mentioning why they were rejected might be interesting to readers.

HIP/hip-1037.md
```diff
@@ -0,0 +1,95 @@
...
+API. This does not materially impact the security of the system.
+
+## How to Teach This
+The Hedera API is a critical component of the design and interaction for the
+Hedera ecosystem. Improvements to the formality and completeness of the API
+specification help everyone involved to develop distributed applications
+more quickly and with greater confidence.
+
+## Reference Implementation
+Not applicable
```
david-bakin-sl, line 82: This might be where you reference the proposed PR.

HIP/hip-1037.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Motivation
+The existing protocol buffer API documentation is brief, entirely descriptive,
+and often inaccurate. This produces two negative consequences. First, it is not
+possible to produce a compatible independent implementation of the Hedera
+consensus node software without extensive reference to the existing source
+code. Second, developers are often unclear how best to interact with the
+Hedera API, and independent SDK developers, in particular, may struggle to
+correctly implement these API calls.
+As part of expanding the Hedera ecosystem we MUST reduce the difficulty and
```
david-bakin-sl, line 33: Would read better if this was an actual new paragraph (markdown has thwarted you here, I suspect).  And at L42 as well.

HIP/hip-1037.md
```diff
@@ -0,0 +1,95 @@
...
+## Rationale
+This HIP improves the developer experience and enhances decentralization by
+making it easier to design distributed applications or alternative
+implementations of core network systems.
+The use of a markdown format that is automatically generated from the
+functional protocol buffer definition files encourages more frequent updates
+that are more closely tied to the functional updates to the API documents, and
+reduces the inevitable "drift" between functional API and API specification.
+
+## User stories
```
david-bakin-sl, line 47: We have an example now of a protobuf which is used to talk to a system contract method (for [HIP-632's `isAuthorized` method](https://hips.hedera.com/hip/hip-632#isauthorizedaddress-message-signatureblob-function-usage)) and so that is potentially another user story to mention here: smart contract developers who need to send a protobuf to a smart contract in a transaction.
* This is slightly different from the current "dAPP developer" which is speaking of "Hedera APIs" i.e., HAPI.  Though that user story could be expanded (just a bit) to take care of this case.

HIP/hip-1037.md
```diff
@@ -0,0 +1,95 @@
...
+## Reference Implementation
+Not applicable
+
+## Rejected Ideas
+Not applicable
+
+## Open Issues
+None.
+
+## References
```
david-bakin-sl, line 90: Can point to [`proto-gen-doc`'s](https://github.com/pseudomuto/protoc-gen-doc/blob/master/README.md) readme right here.

david-bakin-sl (commented):
You'll see below a suggestion that the thrust of this HIP be changed; hope you consider it!

HIP/hip-1037.md
jsync-swirlds: I'll add these to the rejected ideas section.

HIP/hip-1037.md
jsync-swirlds: I'll add these paragraph breaks (which are just blank lines in markdown).

HIP/hip-1037.md
jsync-swirlds: I am not clear on the use case.  Would you be willing to write a suggested text for this?

HIP/hip-1037.md
jsync-swirlds: Good idea, added using the main-page sublink.

HIP/hip-1037.md
jsync-swirlds: This is a very good suggestion.
I've implemented the suggested changes as I understand them.  Please reply with further clarification if I missed part of your intent here.

HIP/hip-1037.md
```diff
@@ -0,0 +1,113 @@
...
+functional protocol buffer definition files encourages more frequent updates
+that are more closely tied to the functional updates to the API documents, and
+reduces the inevitable "drift" between functional API and API specification.
+
+## User stories
+- As a dApp developer I want to have clear and accurate specification for all
+  Hedera APIs so that I can confidently design my application to call those
+  APIs.
+- As an independent implementor I want to have clear and accurate specification
+  for all Hedera APIs so that I can design an independent, correct, and
```
netopyr, line 54: Do we plan to rename them to "Hiero APIs" or something similar?

jsync-swirlds: Not sure.  Definitely something to consider.

HIP/hip-1037.md
Nana-EC: Does something like this work @david-bakin-sl and @jsync-swirlds ?
```suggestion
## User stories
- As a dApp developer that receives bytes or expected to parse bytes representing HAPI protobuf message I want to have a clear description of the structure and implications of the corresponding proto messages to understand how to correctly parse them. 
```

david-bakin-sl: Yes, that is the use case I was thinking of, TY

assets/hip-1037/Specification-Format-Style-Guidelines.md
```diff
@@ -0,0 +1,408 @@
...
+  rare, however; we prefer to avoid the need for such large blocks of
+  specification text.
+- Line oriented comments (i.e. lines prefixed with `//`) may be used for
+  any comments needed in a proto file that are not intended for API
+  specification.  Examples include explaining an `oneof` block, or describing
+  why a particular field number or name is reserved.
+
+## Content Guidelines
+### General `proto` File Guidelines
+- Protocol buffers are compiled for the `hedera-services` codebase using a
```
Nana-EC, line 135: Q: Is this true? Isn't it applicable network wide for MN, BN, SDK repos etc and any 3rd party users?

assets/hip-1037/Specification-Format-Style-Guidelines.md
```diff
@@ -0,0 +1,408 @@
...
+  custom processor 'PBJ'. While no other entity is required to use `PBJ`, all
```
Nana-EC, line 136: Hmm, is this essentially enforcing the community to have a comfortable understanding of PBJ outputs. This may be true but do we want to require this?
Should we call out the practical implications here so users know if they are compatible. How can they have confidence, does it require ensuring certain reserved types aren't used or does it require passing some tests on PBJ?

assets/hip-1037/Specification-Format-Style-Guidelines.md
```diff
@@ -0,0 +1,408 @@
...
+   - For messages (including `enum` or `service` messages), we prefer to use
+     blank lines exclusively, and _not_ use a `<p>` tag.
+- We use line breaks (`<br/>`) in several areas of the specification text.
+  All such line breaks should be placed at the _end_ of the prior line, and
+  should not be placed alone on a line, in the middle of a line, or at the
+  beginning of a line.
+- Certain elements (headings, code blocks, lists, etc...) are "block" elements
+  in markdown and HTML.  Such elements _should not_ be followed by either
+  a break (`<br/>`) or paragraph (`<p>`) tag (or equivalent markdown).
+- Use formal US-English spelling and grammar to the maximum extent possible.
```
Nana-EC, line 29: minor: Do we need to require locale?
```suggestion
- Use formal English spelling and grammar to the maximum extent possible.
```

Nana-EC (commented):
nice work, some considerations

HIP/hip-1037.md
jsync-swirlds: Expanded slightly and added.

assets/hip-1037/Specification-Format-Style-Guidelines.md
jsync-swirlds: PBJ is only used by Services at the moment; eventually Block Node will be added.  I do not know if MN uses PBJ, but I do not _think_ so.
SDKs do not use PBJ, only the Java SDK _could_ and it _should not_ for compatibility with the widest array of users.
3rd party systems written in Java _might_ choose to use PBJ, but I strongly doubt any currently are, and there are good reasons not to (not least that PBJ is not a complete protoc replacement, but is quite strongly focused on the CN use cases).

assets/hip-1037/Specification-Format-Style-Guidelines.md
jsync-swirlds: The simplest way is to pass the PBJ integration test, or pass compilation in the `hedera-services` codebase.  Probably we need to, when resources permit, split the PBJ integration test out into a useable compatibility test library for general use.

The `hedera-services` repo is already the primary repository for any proto under the `services` subfolder (hedera-protobufs is a manually maintained copy), and the current direction is to continue to split every group of proto files out to the various code repositories (so Block Node would have block stream; mirror node would have MN their files, etc...).

Ideally, PBJ _should_ compile anything that `protoc` would.  There are a decent array of open bugs and limitations that prevent this, however.  PBJ is also behind protoc standard by a bit.  PBJ supports proto3 while protoc (with the new "editions" versioning) is at the `2023 Edition`.

jsync-swirlds: I've added some explanation to describe the recommended current process.

assets/hip-1037/Specification-Format-Style-Guidelines.md
jsync-swirlds: Unless we want constant arguments over spelling of certain words, it's highly recommended to choose which version of English is used.

Nana-EC: Different locale spelling I don't mind as much (e.g. behaviour vs behavior) because contributors will come from all over the world but the others I agree can cause confusion so it's a fair consideration.

assets/hip-1037/Specification-Format-Style-Guidelines.md
```diff
@@ -0,0 +1,411 @@
...
+     each row to be a single long line.
+- Be _prescriptive_ in requirements.  Do not _describe_ the behavior.
+   - Declare what MUST be done for valid inputs, and what SHALL be present in
+     valid outputs (or in state messages).
+   - Declare what SHALL happen as a result of a successful, or failed,
+     transaction.
+   - The word `will` usually means `SHALL` or `MUST`.
+   - The word `is` often means `SHALL be`.
+   - Document both positive (`SHALL` or `MUST`) and negative (`SHALL NOT` or
+     `MUST NOT`) requirements.
```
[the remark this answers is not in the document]
jsync-swirlds: I copied the _formal_ definition of these (referring to RFC 2119 and 8174) from EIP-1, and we should update HIP-1 to include a similar reference.
Below is the EIP-1 text.
```
### RFC 2119 and RFC 8174

EIPs are encouraged to follow [RFC 2119](https://www.ietf.org/rfc/rfc2119.html) and [RFC 8174](https://www.ietf.org/rfc/rfc8174.html) for terminology and to insert the following at the beginning of the Specification section:

> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119 and RFC 8174.
```

mgarbs (approved):

PR #1045, opened 2024-09-11
HIP-991 - Updates

Neurone:
**Description**:

This PR:
 - Adds suggestion for API specs
 - Removed key_verification_mode (we postponed to a new HIP to make Topic compliant with HIP-540)
 - Fixes small statements
 - Makes the markdown linter happy

**Related issue(s)**:

N/A

**Notes for reviewer**:

N/A

**Checklist**

### DIFF
--- a/HIP/hip-991.md
+++ b/HIP/hip-991.md
@@ -87,9 +87,7 @@ We propose adding a fixed fee mechanism to the Hedera Consensus Service (HCS) fo
 * Fees can be changed (updated or removed) in a topic with an update topic transaction signed by the Fee Schedule Key.
 * Topics can have a Fee Schedule Key set at creation time.
 * The Fee Schedule Key can manage fee updates for the topic.
-* The Fee Schedule Key follows the same "Lower privilege key" rules introduced by [HIP-540](https://hips.hedera.com/hip/hip-540), i.e:
-  * Only the Admin Key should be able to remove the Fee Schedule Key.
-  * The Fee Schedule Key can change itself to another valid or unusable key.
+* The Fee Schedule Key can be updated according to the same rules that currently apply to the Submit key. In addition, to update the Fee Schedule Key, the new key must sign the transaction.
 * If the topic was created without a Fee Schedule Key, the key cannot be added later.
 
 #### Fee Payment Pre-Requisites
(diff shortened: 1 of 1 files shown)


kimbor (commented):
> The `ConsensusUpdateTopicTransactionBody` message is updated to include the optional Fee Schedule Key, the optional Fee Exempt Key List, and the `custom_fees` property for specifying fixed fees during topic creation.

Need to add the Fee Exempt Key List.

HIP/hip-991.md
```diff
@@ -121,21 +123,25 @@ We propose adding a fixed fee mechanism to the Hedera Consensus Service (HCS) fo
...
   
-* **Handling Duplicates**:
-   - If the FEKL list contains duplicate keys, the transaction will fail with an `FEKL_CONTAINS_DUPLICATED_KEYS` error. This ensures that duplicate entries are not silently ignored, preventing potential bugs or issues in the calling code.
+##### Handling Duplicates
+
+If the FEKL list contains duplicate keys, the transaction will fail with an `FEKL_CONTAINS_DUPLICATED_KEYS` error. This ensures that duplicate entries are not silently ignored, preventing potential bugs or issues in the calling code.
+
+##### Signatures for Invalid/Inactive/Deleted Accounts
+
+The FEKL list will only require keys to be formally valid as per protobuf specifications, meaning they must be correctly formatted key structures. These are keys, not accounts. Even if an account associated with a key is inactive, deleted, or non-existent, the key itself can still be added to the FEKL.
```
kimbor, line 132: Is `key_verification_mode` meant to change this behavior?

HIP/hip-991.md
```diff
@@ -105,13 +105,15 @@ We propose adding a fixed fee mechanism to the Hedera Consensus Service (HCS) fo
...
 * The `allowCustomFeesPayment` flag exists only in the SDK and isn't enforced by the network.
 
-* **Allowance Transactions**: 
-   - An account submits a transaction to grant an allowance (total fees and maximum fee per message) to a topic for both HBAR and HTS tokens. This transaction type reuses the existing allowance concept, with no need for a separate delete allowance transaction. To remove an allowance, the user will submit the same transaction with the allowance values set to 0 or a dedicated flag indicating allowance removal.
-   - This allowance only permits the topic to pay for the custom fees when messages are submitted, not for any other purpose.
+#### Allowance Transactions
+
+* An account submits a transaction to grant an allowance (total fees and maximum fee per message) to a topic for both HBAR and HTS tokens. This transaction type reuses the existing allowance concept, with no need for a separate delete allowance transaction. To remove an allowance, the user will submit the same transaction with the allowance values set to 0 or a dedicated flag indicating allowance removal.
```
kimbor, line 110: I don't see a flag specified in the FeeScheduleAllowance messages below, so I'm thinking that we decided against a dedicated flag

Neurone: Good catch, thanks! I forgot to remove that statement.

HIP/hip-991.md
netopyr: No, `key_verification_mode` only applies to keys on the topic that otherwise would have to sign the message, e.g. the `fee_schedule_key`.

kimbor: It would be useful to add an explanation of key_verification_mode to the hip text. It is present in the protobuf but not explained anywhere else.

Neurone: @kimbor is it ok for you if we postpone this note to a new PR, and we merge this in the meanwhile?
@netopyr can you take care of this additional note?

HIP/hip-991.md
```diff
@@ -87,9 +87,7 @@ We propose adding a fixed fee mechanism to the Hedera Consensus Service (HCS) fo
 * Fees can be changed (updated or removed) in a topic with an update topic transaction signed by the Fee Schedule Key.
 * Topics can have a Fee Schedule Key set at creation time.
 * The Fee Schedule Key can manage fee updates for the topic.
-* The Fee Schedule Key follows the same "Lower privilege key" rules introduced by [HIP-540](https://hips.hedera.com/hip/hip-540), i.e:
-  * Only the Admin Key should be able to remove the Fee Schedule Key.
-  * The Fee Schedule Key can change itself to another valid or unusable key.
+* The Fee Schedule key can be updated according to the same rules that currently apply to the Submit key.
```
netopyr, line 90: We should add that if the Fee Schedule Key is updated, the new one must sign the transaction. This will become mandatory once we implement HIP-540 for the Hedera Consensus Service. It will become a breaking change if we do not enforce this from the beginning.

netopyr (changes requested):
Please remove the field `ConsensusUpdateTopicTransactionBody.key_verification_mode`. It is part of HIP-540 which we decided to implement separately.

HIP/hip-991.md
Neurone: Done

HIP/hip-991.md
Neurone: Done

Neurone:
@netopyr I removed the `key_verification_mode` from the protobuf.

The only thing left is that I'm not sure I see the connection between `key_verification_mode` and the "Signatures for invalid/inactive/deleted accounts" section.

The FEKL should just be a list of keys, it should not be verified in the future, so in my mind, regardless of the value of the `key_verification_mode` parameter.

Does this make sense? Can we proceed with the current paragraph included, or should we change it?

Neurone:
@mgarbs can you merge this?

netopyr (approved):

mgarbs (approved):

PR #1052, opened 2024-09-23
HIP 1052 - Add `file_hash` field to `/api/v1/blocks` endpoints' results 

Neurone:
**Description**:

This HIP proposes to expose the SHA384 hash of the decompressed record file already present in the Mirror Node's DB using the REST APIs via the `/api/v1/blocks` endpoints.

**Related issue(s)**:

N/A

**Notes for reviewer**:

N/A

**Checklist**

### DIFF
--- a/HIP/hip-0000.md
+++ b/HIP/hip-0000.md
@@ -1,0 +1,93 @@
+---
+hip: <HIP number (assigned by the HIP editor), usually the PR number>
+title: Add the `file_hash` field to the `/api/v1/blocks` endpoints' results
+author: Giuseppe Bertone <giuseppe.bertone@hashgraph.com>
+working-group: Giuseppe Bertone <giuseppe.bertone@hashgraph.com>
+requested-by: Giuseppe Bertone <giuseppe.bertone@hashgraph.com>
+type: Standards Track
+category: Mirror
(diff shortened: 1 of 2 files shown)


HIP/hip-0000.md
```diff
@@ -0,0 +1,93 @@
+---
+hip: <HIP number (assigned by the HIP editor), usually the PR number>
+title: Add the `file_hash` field to the `/api/v1/blocks` endpoints' results
+author: Giuseppe Bertone <giuseppe.bertone@hashgraph.com>
+working-group: Giuseppe Bertone <giuseppe.bertone@hashgraph.com>
+requested-by: Giuseppe Bertone <giuseppe.bertone@hashgraph.com>
```
steven-sheehy, line 6: The requested by needs to be an external user needing this change. I don't think we should move forward without an actual user requiring this.

Neurone:
We decided not to proceed with this change. The structure we use to store data or block information is flexible, and this would set a precedent that will lock users and the APIs to report this structure essentially forever.

As a spin-off discussion, we can start exploring designing a Mirror Node that can be extended with API extensions that are not part of the official public API, and that can be enabled or disabled at the discretion of the Mirror Node operator.

PR #1058, opened 2024-10-09
HIP-980: Add shadow fork support as a rejected idea

IvanKavaldzhiev:
**Description**:
After analyzing the efforts for implementing HIP-980, it was proposed to move the idea of shadow fork support in mirror-node as a rejected idea. That is, because it would have introduced too much REST API calls against public mirror node instances and would harm the performance. In addition, the UX would have included more efforts from a user perspective and wouldn't have been so easy to use.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-980.md
+++ b/HIP/hip-980.md
@@ -6,7 +6,7 @@ working-group: Steven Sheehy <@steven-sheehy>, Nana Essilfie-Conduah <nana@swirl
 type: Standards Track
 category: Mirror
 needs-council-approval: Yes
-status: Accepted
+status: In review
 created: 2024-06-03
 requested-by: Solo Labs
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/981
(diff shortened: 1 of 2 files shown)


HIP/hip-980.md
```diff
@@ -11,55 +11,67 @@ last-call-date-time: 2024-07-23T07:00:00Z
 created: 2024-06-03
 requested-by: Solo Labs
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/981
-updated: 2024-08-16
+updated: 2024-10-09
```
Nana-EC, line 14: @mgarbs and @steven-sheehy should we be updating the Status to be Rejected?

HIP/hip-980.md
```diff
@@ -11,55 +11,67 @@ last-call-date-time: 2024-07-23T07:00:00Z
...
+This HIP enhances part of the mirror node REST APIs to support querying by timestamp. This will allow users to query for
+the historical state of any entity info on the network that currently misses this functionality.
 
 ## Motivation
 
-To enhance user experience and allow users or developers to query the state of an entity at a specific point in time we will add a timestamp query parameter to all of the entity-based REST APIs that currently don't have one.
-This change will be especially valuable for building the foundation of the so-called shadow fork functionality, where a user or developer can use a local mirror node instance but run it against a production environment state.
-Having this support, users or developers will have the ability to use public state data to simulate and troubleshoot existing issues or replay historical transactions using local mirror node instance against production state.
+To enhance user experience and allow users or developers to query the state of an entity at a specific point in time we
+will add a timestamp query parameter to all of the entity-based REST APIs that currently don't have one.
```
Nana-EC, line 25: Q: why was the motivation section cut from 3 points to 1?
Seems strange for motivation to have changed now.

IvanKavaldzhiev: It's because the shadow fork initiative is being rejected in this HIP. Most of the points in the motivation section were describing the motivation of having shadow fork functionality. If we remove this part of the section what is left is the extension of reading historical data for all entity types via the REST API.

Nana-EC: Ah i see, the HIP is essentially being altered, or scoped in.
So we reject the shadow forking but we still commit to the historical part.
Gotcha

HIP/hip-980.md
Nana-EC: No longer applicable. HIP is not being rejected but scoped in

mgarbs (approved):

PR #1066, opened 2024-10-23
Update hip-1046.md

Mark-Swirlds:
Based on feedback I've edited the following:

* Updated Abstract to make problem statement clear.
* Change entries containing "grpc_web nodes" to "grpc_web proxies"

**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-1046.md
+++ b/HIP/hip-1046.md
@@ -15,11 +15,11 @@ updated: 2024-10-15
 
 ## Abstract
 
-Frontend applications using the Hedera JavaScript SDK face challenges due to browser security sandbox rules, which block mixed content (i.e., HTTP requests from HTTPS pages) and do not permit direct gRPC requests.  Browsers also require TLS endpoints to be signed by trusted certificate authorities, necessitating fully qualified domain names (FQDNs).
+Frontend applications using the Hedera JavaScript SDK face challenges because of lack of support for direct gRPC requests. 
 
-Currently, gRPC-Web servers work around these limitations, but the endpoints are not supported in the Hedera Address Book. To circumvent this, gRPC-Web endpoints are hard-coded in the JavaScript SDK, which is neither scalable nor efficient. This approach requires manual updates and coordination, which will become increasingly problematic as Hedera decentralizes and more independent nodes operate.
+Currently, gRPC-Web servers bypass these restrictions, but their endpoints are not included in the Hedera Address Book. As a result, the JavaScript SDK uses hard-coded gRPC-Web endpoints, which is inefficient and requires manual updates. This becomes increasingly unsustainable as Hedera decentralizes and more independent nodes are added.
 
-To address this issue, we propose enhancing the Hedera address book to include gRPC-Web endpoint information, allowing the JavaScript SDK to dynamically determine the appropriate endpoints
+We propose enhancing the Hedera Address Book to include gRPC-Web endpoints, enabling the JavaScript SDK to dynamically select the appropriate ones, improving scalability and efficiency.
 
 ## Motivation
 
@@ -63,18 +63,18 @@ Given that this proposal involves an incremental change to the already existing 
     - Node operator can update gRPC-Web endpoint in the address book
     - The updated endpoint is propagated to all frontend applications using the Hedera JavaScript SDK.
     - The process is simple and does not require extensive coordination with SDK maintainers
-3. As an Hedera community member I want to run my own gRPC-Web node for my consensus node and make it visible to the community to assist in growing the network.
+3. As an Hedera community member I want to run my own gRPC-Web proxy for my consensus node and make it visible to the community to assist in growing the network.
     
     Acceptance Criteria:
     
-    - A community member can launch a gRPC-Web node without external interaction.
-    - A community member can submit a request to have their gRPC-Web node added to the network address book.
+    - A community member can launch a gRPC-Web proxy without external interaction.
+    - A community member can submit a request to have their gRPC-Web proxy added to the network address book.
   
 ## Specification
 
-The dynamic address book stores an entry in state for each node (only consensus, currently).  This HIP proposes to extend that “address book” concept to include gRPC-Web node endpoints as an additional field of the `Node` entry.  The expectation is that these additional node endpoints will be managed with the same `nodeCreate`, `nodeUpdate`, and `nodeDelete` transactions that manage the other endpoints for consensus nodes.  The only changes needed are one additional field in the `NodeCreateTransactionBody`, the same field in `NodeUpdateTransactionBody`, and a matching additional field in the `Node` entry in state.
+The dynamic address book stores an entry in state for each node (only consensus, currently).  This HIP proposes to extend that “address book” concept to include gRPC-Web proxy endpoints as an additional field of the `Node` entry.  The expectation is that these additional node endpoints will be managed with the same `nodeCreate`, `nodeUpdate`, and `nodeDelete` transactions that manage the other endpoints for consensus nodes.  The only changes needed are one additional field in the `NodeCreateTransactionBody`, the same field in `NodeUpdateTransactionBody`, and a matching additional field in the `Node` entry in state.
 
-Any entity wishing to operate a gRPC-Web proxy node would request to have that node added to the network address book via a `nodeCreate` (which, initially, requires council approval), and would manage that node via `nodeUpdate` as needed to maintain current and up-to-date information for the service endpoints of that node.
+Any entity wishing to operate a gRPC-Web proxy would request to have that node added to the network address book via a `nodeCreate` (which, initially, requires council approval), and would manage that node via `nodeUpdate` as needed to maintain current and up-to-date information for the service endpoints of that node.
 
 ### Protobufs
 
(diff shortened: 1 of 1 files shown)


HIP/hip-1046.md
```diff
@@ -154,11 +226,16 @@ actions is extremely low.
...
+ensuring developers understand it and are aware that it will not affect
+existing SDK-based applications.
 
 ## Reference Implementation
 
-The reference implementation must be complete before any HIP is given the status of “Final”. The final implementation must include test code and documentation.
+The reference implementation must be complete before any HIP is given the status
+of “Final”. The final implementation must include test code and documentation.
 
 ## Rejected Ideas
```
netopyr, line 240: I would have expected a new field in `node_create.proto` and `node_update.proto` like:
```
repeated proto.ServiceEndpoint grpc_web_proxy_endpoint;
```

There are probably good reasons why this approach was not taken, even though it seems less error-prone to use. Can we please add an entry under "Rejected Ideas" explaining why this approach was not taken?

netopyr (changes requested):
Why do we want to change the API like this? This looks more complicated and error-prone than the current definition.

HIP/hip-1046.md
```diff
@@ -2,147 +2,219 @@
...
+frontend applications to dynamically determine the appropriate endpoints.
 
-*Why:* Updating the JavaScript SDK to dynamically retrieve gRPC-Web endpoints from the address book eliminates the need for hard-coded endpoints. This design decision enhances flexibility and scalability, allowing frontend applications to adapt to changes in the network configuration without requiring manual updates.
+*Why:* Updating the JavaScript SDK to dynamically retrieve gRPC-Web endpoints
+from the address book eliminates the need for hard-coded endpoints. This design
+decision enhances flexibility and scalability, allowing frontend applications to
+adapt to changes in the network configuration without requiring manual updates.
 
-*Alternatives Considered:* Another alternative was to continue with the current practice of hard-coding endpoints within the SDK. However, this method is not sustainable as it requires frequent updates and coordination, particularly as Hedera continues to decentralize. Dynamic discovery provides a more robust and future-proof solution.
+*Alternatives Considered:* Another alternative was to continue with the current
```
steven-sheehy, line 81: This 80 character limit seems too restrictive and conflicts with our Java coding [standards](https://www.notion.so/swirldslabs/007-Java-Code-Style-a0130a7059b044f68447b9ca3a5242b3#9bc7be511ea94b588d407491c28068c3) to use 120. The standards just mention java so not sure about other formats. Is it documented somewhere that we should use 80 for markdown?

HIP/hip-1046.md
```diff
@@ -2,147 +2,219 @@
...
+     * This is one of several options, including a gossip endpoint,
+     * a gRPC endpoint, or a gRPC-Web Proxy endpoint, among others.
+     * <p>
+     * This field is REQUIRED.<br/>
+     * This field MUST match the actual purpose of this endpoint.<br/>
+     * There MUST NOT be more than one "gossip" endpoint for each `Node`.<br/>
+     * There MUST NOT be more than one "secure gRPC" endpoint for
+     * each `Node`.<br/>
+     * If a `Node` has a "secure gRPC" endpoint, it SHOULD NOT have an
+     * "insecure gRPC endpoint.
```
steven-sheehy, line 156: These statements are not really true. A node can have more than one secure port on different proxy IPs. A node can also have both secure and insecure on the same IP since they are different ports. I imagine in the future with permissionless some nodes might even want multiple secure per IP or multiple insecure per IP. Here's node 0.0.3 service endpoints on mainnet right now:

```json
"service_endpoints":[
{"domain_name":"","ip_address_v4":"34.239.82.6","port":50211},
{"domain_name":"","ip_address_v4":"34.239.82.6","port":50212},
{"domain_name":"","ip_address_v4":"35.237.200.180","port":50211},
{"domain_name":"","ip_address_v4":"35.237.200.180","port":50212}]
```

```

**Review by steven-sheehy — CHANGES_REQUESTED** [review]

File: HIP/hip-1046.md (hunk 2 above)
Comment (jsync-swirlds) [reply HIP/hip-1046.md:81]: There is no standard for markdown that I know of, but github enforces 80 characters in much of the review interface (particularly suggestions), so I habitually reformat HIP markdown to 80 characters for ease of review.
I do not think there is any requirement to do so, however.

File: HIP/hip-1046.md (hunk 3 above)
Comment (jsync-swirlds) [reply HIP/hip-1046.md:156]: It is permitted to do so (SHOULD is a recommendation keyword).  It is not best practice, however.

Comment (steven-sheehy) [reply HIP/hip-1046.md:156]: I was commenting on both statements. There is a MUST NOT statement that is not true.

And I disagree with your SHOULD recommendation as it can't be considered a bad practice if every node in production does it. The statements should reflect the reality on mainnet.

Comment (jsync-swirlds) [reply HIP/hip-1046.md:156]: I see.  That is not ideal (having both secure and insecure endpoints is bad security practice), but I can remove the specification text.

File: HIP/hip-1046.md
Comment (rbair23) [inline HIP/hip-1046.md:1]: I don't understand the new proposal. The `NodeCreateTransactionBody` for example, which adds a node to the address book, has `ServiceEndpoints` such as `gossip_endpoint` and `service_endpoint`. This HIP doesn't change that. But it does add `ServiceEndpointPurpose`. So what happens if I set a ServiceEndpoint on `gossip_endpoint` with a purpose of `SECURE_GRPC`?

I'm assuming the node transaction will have to be updated in an incompatible way? That isn't a problem since it is not currently used, but I didn't see that change in the specification section.

**Comment by netopyr** [comment]
I think we should keep the different types of endpoints separated. Instead of putting everything into a single list and adding a type, I suggest defining the new field in `NodeCreateTransactionBody` and `NodeUpdateTransactionBody` as lists:
```
repeated proto.ServiceEndpoint grpc_proxy_endpoint;
```
Both solutions provide the same functionality. Everything that can be configured with a single list can also be configured with two lists and vice versa. Storing the gRPC proxies in a separate list has several advantages, though:

1. A single list makes errors possible that would otherwise not occur. As @rbair23, for example, notes, the API does not strongly discourage me from putting a gossip endpoint into the list of service endpoints. Another point of failure is the additional mandatory `purpose` field that one may forget.
2. It is inconsistent and unexpected that one type of `ServiceEndpoint` has a separate list while the others do not.
3. A single list makes it harder to run checks like "Does the node define at least one service point?". These checks will probably be implemented on the server and in all SDKs.

I agree with @steven-sheehy's observation that we should not rely on port ranges to determine if an endpoint is secured. How about adding a boolean flag in `ServiceEndpoint`?

HIP/hip-1046.md
jsync-swirlds: After this HIP, we would need to enforce the correct `purpose` value for `nodeCreate` and `nodeUpdate` endpoint lists in consensus node software.

All changes needed would be compatible from the perspective of protobuf.  The node transaction definitions do not need to be changed here as we're just adding a new field.
During upgrade we might need perform a state migration to set the purpose for each endpoint (which can be determined from the containing list and port, as Mirror Node does currently).

jsync-swirlds:
> I think we should keep the different types of endpoints separated. Instead of putting everything into a single list and adding a type, I suggest defining the new field in `NodeCreateTransactionBody` and `NodeUpdateTransactionBody` as lists:
> 
> ```
> repeated proto.ServiceEndpoint grpc_proxy_endpoint;
> ```
> 
> Both solutions provide the same functionality. Everything that can be configured with a single list can also be configured with two lists and vice versa. Storing the gRPC proxies in a separate list has several advantages, though:
> 
> 1. A single list makes errors possible that would otherwise not occur. As @rbair23, for example, notes, the API does not strongly discourage me from putting a gossip endpoint into the list of service endpoints. Another point of failure is the additional mandatory `purpose` field that one may forget.
> 2. It is inconsistent and unexpected that one type of `ServiceEndpoint` has a separate list while the others do not.
> 3. A single list makes it harder to run checks like "Does the node define at least one service point?". These checks will probably be implemented on the server and in all SDKs.
> 
> I agree with @steven-sheehy's observation that we should not rely on port ranges to determine if an endpoint is secured. How about adding a boolean flag in `ServiceEndpoint`?

This is exactly what the current [HIP text](https://github.com/hashgraph/hedera-improvement-proposal/blob/main/HIP/hip-1046.md) does (aside from eliding the `repeated` keyword, which was requested during technical review, as we do not expect to have more than one proxy per node), prior to this update.

We probably also need to think about whether we expect to have _more_ node types in the future.  Already the `Node` object is very specific to consensus _as currently implemented_, and would be difficult to adjust to handle, for instance, a split between consensus and execution, or the addition of Block Nodes to the address book.

netopyr:
> This is exactly what the current [HIP text](https://github.com/hashgraph/hedera-improvement-proposal/blob/main/HIP/hip-1046.md) does (aside from eliding the `repeated` keyword, which was requested during technical review, as we do not expect to have more than one proxy per node), prior to this update.

I know. I think the new proposal is worse than the current HIP.


> We probably also need to think about whether we expect to have _more_ node types in the future. Already the `Node` object is very specific to consensus _as currently implemented_, and would be difficult to adjust to handle, for instance, a split between consensus and execution, or the addition of Block Nodes to the address book.

I do not see how the split between consensus and execution would influence the transactions. For block nodes, we probably want specialized transaction types. I do not see much commonality between consensus nodes and block nodes.

bugbytesinc:
In general this seems to be bad architecture to require nodes that only communicate via gRPC to maintain a list of endpoints that they do not control, nor can validate and attest to their correctness.  If the problem is due to SDK design, the correction should be mitigated in that tech stack not the consensus node protocol.

jsync-swirlds:
> In general this seems to be bad architecture to require nodes that only communicate via gRPC to maintain a list of endpoints that they do not control, nor can validate and attest to their correctness. If the problem is due to SDK design, the correction should be mitigated in that tech stack not the consensus node protocol.

Node operators control all three types (gRPC, Proxy, gossip) endpoints for their node via nodeUpdate transactions.

mgarbs:
Hello Everyone,

I hope this message finds you well. I wanted to bring your attention back to the PR titled 'Update hip-1046.md'. We've been making great progress so far, but there are still 9 discussions that need resolution.

I am certain that each one of your thoughtful insights would be extremely valuable in resolving these discussions and moving this PR forward. 

I would like to extend my sincere appreciation for your time and participation in this process. It's your consorted efforts that will help improve the quality of the changes proposed in this PR.

Looking forward to hearing from you!

Best regards@netopyr - There's an unresolved discussion from 2024-10-30. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1066#discussion_r1822685029)

@steven-sheehy - There's an unresolved discussion from 2024-10-30. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1066#discussion_r1823520439)

@steven-sheehy - There's an unresolved discussion from 2024-10-30. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1066#discussion_r1823527722)

@jsync-swirlds - There's an unresolved discussion from 2024-10-31. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1066#discussion_r1824904278)

@jsync-swirlds - There's an unresolved discussion from 2024-10-31. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1066#discussion_r1824905217)

@steven-sheehy - There's an unresolved discussion from 2024-10-31. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1066#discussion_r1824909142)

@jsync-swirlds - There's an unresolved discussion from 2024-10-31. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1066#discussion_r1824915328)

@rbair23 - There's an unresolved discussion from 2024-11-01. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1066#discussion_r1825338142)

@jsync-swirlds - There's an unresolved discussion from 2024-11-01. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1066#discussion_r1826310656)

Mark-Swirlds:
Hello Michael,
I'll set-up a meeting with stakeholders this week to discuss and hopefully
resolve.
Mark

PR #1068, opened 2024-10-25
Sponsored Fees

Nana-EC:
**Description**:
This HIP proposes an expansion to the approval and allowance network logic (i.e. CryptoApproveAllowance, NftAllowance
and TokenAllowance) by allowing any account on the network to provide allowances that will sponsor the payment of the
fees for another account’s submitted transactions.
In this way account Alice may assign an HBAR / token transaction fee allowance for EOA account Bob or Contract account
Carol and pay the fees due to nodes, the network and accounts for any transactions they submit, thus allowing Bob or
Carol to have zero balance and still transact on chain.

This proposal extends the existing Approval and Allowance API by adding the concept of a scope that designates the type
of value transfer an allowance applies to.
The scopes will include CryptoTransfer, HAPI transaction fee, token custom fees and gas fees.
This enables value transfer and fee payments to another account without requiring one-off
approval for each transaction. This feature enhances flexibility and efficiency in managing transaction costs,
particularly for high volume applications.

**Related issue(s)**:

Fixes #912 

**Notes for reviewer**:
Expect more changes soon.
There's an outstanding item to condense the Allowances types to instead be scopes on the existing allowances as a cleaner API form

**Checklist**


Nana-EC:
Updated HIP with scope concept and fixed PR description with updated Abstract

tud0r6:
Would it be possible for sponsored fees to be optionally set towards spending by a smart contract, thus limiting the scope of the approval to executing only 1 particular smart contract.

gregscullard:
> Would it be possible for sponsored fees to be optionally set towards spending by a smart contract, thus limiting the scope of the approval to executing only 1 particular smart contract.

I like this idea, to be clear, you mean that if I grant some hbar to Alice, I would be able to specify that Alice can only spend the hBar on executing transactions related to a specific contract, nothing else. She would not be able to execute another contract, or create an NFT, etc...

Note, hbar in this case is gas for contracts, does the allowance extend to paying for gas (through the SDK and/or relay), or just native fees ? If only native fees, we may need to consider supporting gas payments.

Nana-EC:
> > Would it be possible for sponsored fees to be optionally set towards spending by a smart contract, thus limiting the scope of the approval to executing only 1 particular smart contract.
> 
> @tud0r6 I like this idea, to be clear, you mean that if I grant some hbar to Alice, I would be able to specify that Alice can only spend the hBar on executing transactions related to a specific contract, nothing else. She would not be able to execute another contract, or create an NFT, etc...
> 
> Note, hbar in this case is gas for contracts, does the allowance extend to paying for gas (through the SDK and/or relay), or just native fees ? If only native fees, we may need to consider supporting gas payments.

Great considerations.
So the intention of this HIP was to cover value transfers that cover fees. 
Initially it was not intended for gas, however, gas is eventually settled used to define the work but it's paid in the form of hbar.
So there's a possibility to extend it to cover gas payments with a specific scope or as GAS_PAYMENT 🤔.
I like it, though we need to think on any unintended emergent features here 😅.
For example the Ethereum JSON RPC API doesn't support the ability for an EOA to specify the transaction is sponsored. Which would mean it might only be available via `ContractCall`/`ContractCreate` HAPI and not via `EthereumTransaction`

tud0r6:
> tention of this HIP was to cover value transfers that cover fees.
> Initially it was not intended for gas, however, gas is eventually settled used to define the work but it's paid in the form of hbar.

Indeed, I won't want to sponsor general-use gas(HBAR), rather application(SC)-specific gas(HBAR).

Nana-EC:
@gregscullard and @tud0r6 I've added the gas scope but also added thoughts in the open questions section around some challenges here.
Any ideas on how to address this?

gregscullard:
@Nana-EC on the gas scope and the EVM tooling question, could we not consider that if an EOA account has 0 balance, then gas fees should be paid out of the allowance, or if the account has balance, it consumes that balance first, then dips into allowances if they exist ?
Given the inference is that accounts would have 0 balance, this seems to be in line with the ethos of the hip and this would make the solution compatible with existing tooling.

I think we're still missing something with regards to the applicability of allowances for gas. Specifically the ability to be explicit about which contracts can be executed with a given allowance. User-stories-ish for example:

- Alice allows Bob to spend up to 100 hbar on gas costs (implies any contract)
- Carol allows Dave to spend up to 1000 hbar on gas costs only for contract 0.0.1234
- Carol allows Grace to spend up to 1000 hbar on gas costs for either contract 0.0.1235 and 0.0.1243 (how many can be specified here could be subject to a limit of 10 for example).
- Alternatively, Carol allows grace to spend up to 500 hbar on gas costs for contract 0.0.1235 and separately allows Grace to spend up to 500 hbar on gas costs for contract 0.0.1243

Nana-EC:
> @Nana-EC on the gas scope and the EVM tooling question, could we not consider that if an EOA account has 0 balance, then gas fees should be paid out of the allowance, or if the account has balance, it consumes that balance first, then dips into allowances if they exist ? Given the inference is that accounts would have 0 balance, this seems to be in line with the ethos of the hip and this would make the solution compatible with existing tooling.

Hey @gregscullard ,
I see what you're going for. To date though the network requires explicit intent from an account to utilize an allowance vs its own balance. This is the case for `CryptoTransfer` so it'd be inconsistent to do so with gas expenses.
It might not be right for the network to infer as different users may have different intents or be held to different policies offchain.
As such even with 0 balance but non zero allowance a node should still reject a transaction where there's a requirement to transfer value from a spender.
For example with ERCs a user must call transferFrom and with HAPI a user must set is_approval to true to tap into an allowance.
Additionally, a user may have multiple allowances, a node would not be able to infer the correct allowance to pull from even if it did want to.
I'm still pondering though to see if there's a pathway for this HIP, but I think it should be a proposal for another HIP.
@rbair23 thoughts here?

Nana-EC:
> I think we're still missing something with regards to the applicability of allowances for gas. Specifically the ability to be explicit about which contracts can be executed with a given allowance. User-stories-ish for example:
> 
> * Alice allows Bob to spend up to 100 hbar on gas costs (implies any contract)
> * Carol allows Dave to spend up to 1000 hbar on gas costs only for contract 0.0.1234
> * Carol allows Grace to spend up to 1000 hbar on gas costs for either contract 0.0.1235 and 0.0.1243 (how many can be specified here could be subject to a limit of 10 for example).
> * Alternatively, Carol allows grace to spend up to 500 hbar on gas costs for contract 0.0.1235 and separately allows Grace to spend up to 500 hbar on gas costs for contract 0.0.1243

Hey @gregscullard ,
I get the user scenarios and the value. I just don't believe that it connects with this HIP which utilizes allowances logic as is but applies them for nameable criteria regardless of the other entities involved in the transaction.
The requested changes to the network for an account to be able to specify allowance to another account only for expenses on other entities is an extra dimension for the nodes to manage with state and complexity implications.
I think it'd be a great feature. 
I just think it should be in a separate HIP as the logic differs from this and it needs to be considered separately and deeply.
@rbair23 thoughts?

HIP/hip-1068.md
```diff
@@ -0,0 +1,528 @@
...
+onboarding process. 
+
+The proposed changes are made in alignment with the current implementation of allowances for similar capabilities
+(as seen in [HIP 336](https://github.com/hashgraph/hedera-improvement-proposal/blob/main/HIP/hip-336.md)).
+
+## User stories
+
+1. As an existing account on the network I want to be able to sponsor up to a certain HBAR limit the cost of future
+    transaction fees due by another account by granting an explicit allowance.
+2. As an existing account on the network I want to be able to sponsor an unlimited amount of HBARs to cover the cost
```
IvanKavaldzhiev, line 89: This behaviour seems similar to the behaviour we have for NFT allowances, more specifically approvedForAll=true, where each new serial number of a given NFT will have been granted as an allowance to the spender.

Is the concept here the same? Since hbars are fungible there might be some applications like:

1. If a sponsor gives allowance to multiple accounts with unlimited HBAR amounts - the granted accounts might use a competative greedy approach to try to use a part or the whole balance for their personal use before another account and thus having unfair share of the sponsored balance
2. Defend the sponsor against a malicious account which performs operations in a loop, exhausting all of the balance

There might be safe use cases for this unlimited amount sponsorship, but I wanted to raise some security implications and corner cases, which can be thought of.

HIP/hip-1068.md
```diff
@@ -0,0 +1,528 @@
...
+    2. Otherwise the transaction fails.
+5. Where the transaction succeeds, the HBAR fees for the transaction are deducted from the Fee Sponsors account (i.e.
+    the software provider).
+    1. The customer’s HBAR account balance (used to sign the transaction) is left untouched
+    2. The customer’s HBAR balance can be zero.
+6. The transaction is successful.
+
+### Protobuf
+
+A new new allowance scope will be added to help define which type of fee an allowance can be used for
```
IvanKavaldzhiev, line 179: Typo - new repetition

HIP/hip-1068.md
```diff
@@ -0,0 +1,528 @@
...
+        ...
+
+        /**
+        * The applicable scope for this TokenAllowance. 
+        */
+        repeated AllowanceScope scope = 5;
+    }
+```
+
+By default existing nft allowances will map to an `NftAllowance` with scope `CRYPTO_TRANSFER`
```
IvanKavaldzhiev, line 245: Typo: NftAllowance -> TokenAllowance

HIP/hip-1068.md
```diff
@@ -0,0 +1,528 @@
...
+`CRYPTO_TRANSFER` allowances.
+
+The updated TransactionBody will contain the sponsoring account identifier
+```protobuf
+    /**
+     * A value transfer sponsor claim.
+     */
+    message TransactionSponsorClaim {
+        /**
+         * The account a spender is noting has provided prior willingness through an transaction fee allowance to
```
IvanKavaldzhiev, line 257: Typo: an -> a

I think the field description is a bit too complex and can be simplified. Part of the description can be moved in the description for the TransactionSponsorClaim operation

HIP/hip-1068.md
```diff
@@ -0,0 +1,528 @@
...
+    the software provider).
+    1. The customer’s HBAR account balance (used to sign the transaction) is left untouched
+    2. The customer’s HBAR balance can be zero.
+6. The transaction is successful.
+
+### Protobuf
+
+A new new allowance scope will be added to help define which type of fee an allowance can be used for
+```protobuf
+    enum AllowanceScope {
```
IvanKavaldzhiev, line 181: The `GAS` scope seems to be missing.

hendrikebbers: Isn't GAS part of the TRANSACTION_FEE for a Smart contract call?

HIP/hip-1068.md
```diff
@@ -0,0 +1,528 @@
...
+### System Contract Functions
+
+Similar to [HIP 906](https://hips.hedera.com/hip/hip-906), the system contract interface would need to be updated to
+support the management of allowances with scope by a smart contract.
+
+The following functions are introduced in this HIP
+```solidity
+    interface IHRC1068 {
+        /// extensions of HIP 514
+        function approve(address token, address spender, uint256 amount, AllowanceScope allowanceScope) external
```
hendrikebbers, line 339: Definition of AllowanceScope enum in solidity is missing

HIP/hip-1068.md
```diff
@@ -0,0 +1,528 @@
...
+
+        /// extensions of HIP 906
+        function hbarAllowance(address spender, AllowanceScope allowanceScope) external returns (responseCode,
+            int256 amount);
+        function hbarApprove(address spender, int256 amount, AllowanceScope allowanceScope) external
+            returns (responseCode);
+    }
+```
+
+`IHRC1068` token methods will be callable via the HTS system contract at `0x167`.
```
hendrikebbers, line 354: While I know HTS I do not know what HAS or EOA stand for. Can we add a link or the full name at first usage?

HIP/hip-1068.md
IvanKavaldzhiev: There is an interesting case that we should make sure is covered properly by this scope concept.

If we performa a precompile smart contract call that creates a token - the precompile requires gas amount in tinybars + separate value sent in tinybars, since the gas amount will be above some of the limits. This implies that calling createFungibleToken precompile call might make use of both `GAS` and `TRANSACTION_FEE` scopes.

We can elaborate more of what `TRANSACTION_FEE` covers in this HIP as a scope? I understand it as hbar amount needed to cover transaction expenses different than the gas itself. Or is it actually the gas amount itself?

HIP/hip-1068.md
IvanKavaldzhiev: HAS -> Hedera Account Service
EAO -> Externally owned account

I agree with your comment for adding more info about the full names.

HIP/hip-1068.md
```diff
@@ -0,0 +1,528 @@
...
+- AccountAllowanceApproveTransaction
+- AccountAllowanceAdjustTransaction
+- AccountAllowanceDeletionTransaction
+
+### Mirror Node API
+
+The Mirror Node allowance REST APIs would need to be updated to support the applicable scope concept values.
+
+#### getCryptoAllowances
+The `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/crypto` would need to be updated to to support all scopes.
```
hendrikebbers, line 394: Do I call this endpoint when want to get information about an allowance for calling smart contracts for example? If that is correct I dislike the "/crypto" at the end. Since we have the scope I would assume "/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances" as endpoint.

HIP/hip-1068.md
hendrikebbers: But at some other points GAS is used as a scope (Mirror Node endpoint, for example). Here it must be clear what is the difference and if GAS must be included.

IvanKavaldzhiev: I agree, I'm a bit confused regarding the difference between `GAS` and `TRANSACTION_FEE` scope.

Nana-EC: To clarify, gas is a concept of resource utilization. 
Gas fee is how much each unit of gas costs a user in HBAR.
Here the allowance is capturing the cost in hbar when settlement occurs

HIP/hip-1068.md
Nana-EC: This is an existing endpoint that lists all the allowances with relation to hbar crypto.
This is distinguished from the `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/nfts` and `/api/v1/accounts/{idOrAliasOrEvmAddress}/allowances/tokens` endpoints. 
See https://testnet.mirrornode.hedera.com/api/v1/docs/

HIP/hip-1068.md
Nana-EC: Added

HIP/hip-1068.md
Nana-EC: Added

HIP/hip-1068.md
Nana-EC: Fixed

HIP/hip-1068.md
Nana-EC: Fixed

HIP/hip-1068.md
Nana-EC: Interesting points.
If by similar you mean will new hbars or FTs made available by a treasury be applicable to an existing allowance then yes, this would continue to be the case.

The cases noted are the same today I believe if an account was to provide one or multiple allowances with large enough values in comparison to their balance, so I don't believe this is a new edge case.
Solid consideration but I think it's already covered within the service capabilities.

It's the responsibility of the account to consider its awarded allowances.

HIP/hip-1068.md
Nana-EC: Gas is an EVM concept which is a proxy for the amount of work the EVM needs to do to porcesss your transaction, which is not the same as the Hedera transaction fee (cost assigned to each HAPI transaction based on the feeSchedule)
Gas is indeed settled in HBAR and charged to an EOA so this might be where things are getting mixed up.
They are separate concepts and scopes so we have a TRANSACTION_FEE and a GAS_FEE to cover each separately.
Both cases are existing concepts on Hedera and are noted in summary in the HIP in the protobuf summary.
Maybe additional illustrations of the Scopes would be helpful.

It is possible as @IvanKavaldzhiev noted that a single transaction could utilize a combination of scopes. If that user has the appropriate allowances then the network will support it.

rbair23: For the EVM this is very interesting. We have two scopes: `GAS_FEES` and `TRANSACTION_FEES`. So we should consider (and document) the behavior of three variations:

1. A `ContractCall` transaction has only a single allowance for `GAS_FEES`
2. A `ContractCall` transaction has only a single allowance for `TRANSACTION_FEES`
3. A `ContractCall` transaction has two scopes: `GAS_FEES` and `TRANSACTION_FEES` with the same sponsor
4. A `ContractCall` transaction has two scopes: `GAS_FEES` and `GAS_FEES` by two sponsors
5. A `ContractCall` transaction has two scopes: `GAS_FEES` and `TOKEN_CUSTOM_FEES` by one or more sponsors

The first case is the base expected case: the sponsor will pay for a certain amount of gas. Gas is denominated in USD, and so the amount of hbars per gas is variable. So the sponsor is saying "I will pay whatever the USD equivalent of HBAR is up to this limit of gas". So this works.

For the second case, gas fees are a subset of possible transaction fees, so it seems very reasonable that a sponsor willing to pay for transaction fees would expect it to also cover gas. For example, as a sponsor, I could say I am willing to pay a certain amount of HBAR and I don't care whether it is for crypto transfer fees for gas fees. We could instead say that gas fees are special and must be specified separately.

The third case is dependent on our answer to the second case. If we say that TRANSACTION_FEES is illegal, then the third case is also illegal. If however we say that TRANSACTION_FEES is legal, then we have a situation here where the sponsor has provided two different allowances and we have a decision to make. Is this combination saying "you can spend X gas AND Y transaction fees", or is it saying "you can spend X gas but not more than Y hbars when converted to fees".

The fourth case we should just define the behavior. I think we will go to the first sponsor and use their funds first, and only if there is insufficient, then we go to the second sponsor. This seems simple enough.

The fifth case should be a failure because `TOKEN_CUSTOM_FEES` is not applicable to `ContractCall`. We should fail transactions that have inapplicable scopes. This is a little frustrating for dapps because they have to analyze what the transaction type is and then select the right sponsor and claim type, they cannot just put all claims in there and let the network figure it out. But this has been our policy to this point for all our transactions, to be strict on what we accept, to have the greatest forward compatibility. If this becomes a problem we can always loosen our position, but if we start off loose, we cannot tighten it.

So I think the pivotal cases are the second and third.

mgarbs (dismissed):

HIP/hip-1068.md
Nana-EC: Initially failing on the use of non applicable allowance vs ignoring them seem appropriate. it will encourage strictness and will help preserved deterministic of execution of transactions.

Nana-EC: Also i think it might be worth considering accounting for the `GAS_FEES` in gas as there is $/gas that is required prior to conversion to HBAR. This will allow EVM players to estimate costs/allowances without consideration of exchange rates.
Of course the opposite could be true and sponsors may want to only know how much HBARs they are sponsoring 🤔

admin-hedera:
There's been some conversation regarding allowances/allocations for smart contracts to be denominated in gas vs hbar. I would like to offer up consideration of providing an option for the user setting the allowance in either denomination. This could be set implicitly (via a default if the option is null, say hbar) or be required to be set explicitly with an expected failure of the allowance if not set. Given we already will need to set allowances beyond hbar (e.g. HTS allowances), would suggest this would be consistent with setting the unit of allowance.

popowycz:
There's been some conversation regarding allowances/allocations for smart contracts to be denominated in gas vs hbar. I would like to offer up consideration of providing an option for the user setting the allowance in either denomination. This could be set implicitly (via a default if the option is null, say hbar) or be required to be set explicitly with an expected failure of the allowance if not set. Given we already will need to set allowances beyond hbar (e.g. HTS allowances), would suggest this would be consistent with setting the unit of allowance.

HIP/hip-1068.md
```diff
@@ -0,0 +1,540 @@
...
+Ultimately the current Hedera ecosystem lacks a native capability for an Owner account to cover transaction fees of a
+Spender without co-signing for every transaction. This limitation hinders the development of user-friendly applications
+and services that could benefit from a more flexible transaction fee payment flow.
+
+## Rationale
+
+Today the network supports 3 types of allowances `HbarAllowance`, `TokenAllowance` and `NftAllowance` . An account may
+submit a `CryptoApproveAllowance` transaction and specify the amount of value they are willing to allow another account
+to transfer out of their own balance. This HIP builds upon the allowance logic with the introduction of another
+allowance type `TransactionFeeAllowance`.
```
stoqnkpL, line 62: This is mentioned as a rejected idea below and other protobuf types are proposed instead in the `Specification` section. Should this be updated?

mgarbs:
Hi Team,

I hope this message finds you well. I wanted to follow up on PR # [Sponsored Fees]. It's incredibly exciting to see all the insightful discussions and recommendations we've been having on this front. Brilliant work everyone!

However, there are still 28 unresolved discussions remaining that need our attention and resolution. If we can make an extra push to clear these up, we can make significant progress on our PR. Specifically, I wanted to kindly ask @mgarbs, could you please provide your valuable inputs when you get time? Your perspective is crucial for us to move forward.

Once again, thank you all for your dedicated participation and thoughtful considerations in this PR. It's been amazing to see everyone's level of engagement. Let's bring it across the finish line together!

Best regards,
[Your name]
📋 **Pending Reviews:**
@mgarbs - Please review this PR when you get a chance.

@IvanKavaldzhiev - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1879690309)

@IvanKavaldzhiev - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1879872952)

@IvanKavaldzhiev - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1879880842)

@IvanKavaldzhiev - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1879885011)

@IvanKavaldzhiev - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1880054587)

@hendrikebbers - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1880086967)

@hendrikebbers - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1880096685)

@hendrikebbers - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1880098716)

@IvanKavaldzhiev - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1880099188)

@IvanKavaldzhiev - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1880101886)

@hendrikebbers - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1880105611)

@hendrikebbers - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1880106844)

@IvanKavaldzhiev - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1880123585)

@Nana-EC - There's an unresolved discussion from 2024-12-16. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1887456421)

@Nana-EC - There's an unresolved discussion from 2024-12-16. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1887518182)

@Nana-EC - There's an unresolved discussion from 2024-12-16. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1887522235)

@Nana-EC - There's an unresolved discussion from 2024-12-16. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1887522273)

@Nana-EC - There's an unresolved discussion from 2024-12-16. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1887522362)

@Nana-EC - There's an unresolved discussion from 2024-12-16. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1887522484)

@Nana-EC - There's an unresolved discussion from 2024-12-16. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1887522532)

@Nana-EC - There's an unresolved discussion from 2024-12-16. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1887563390)

@Nana-EC - There's an unresolved discussion from 2025-01-14. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1914296203)

@rbair23 - There's an unresolved discussion from 2025-01-14. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1915139228)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1923326154)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1923326305)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1923326381)

@hendrikebbers - There's an unresolved discussion from 2025-01-28. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1932243997)

@stoqnkpL - There's an unresolved discussion from 2025-01-28. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1068#discussion_r1932330139)

hendrikebbers:
Based on all the ongoing discussions on this HIP I suggest to drop it from the TSC agenda for now. I think this topic is super important and might be a gamechanger for web2 and enterprise to use web3 technology by making the complete payment process way easier and morw comforatble for those companies. Based on that I think that additional dicussions are needed before we bring it back to the TSC

HIP/hip-1068.md
```diff
@@ -0,0 +1,540 @@
...
+
+A new query param `?scope=(CRYPTO_TRANSFER|TOKEN_CUSTOM_FEES)` would be added to allow
+users to filter on the specific applicable scope type.
+
+## Backwards Compatibility
+
+Currently the ledger allows for sponsored transaction fees by charging the transaction submitter/payer. This HIP does
+not conflict with this flow, but rather adds an additional flow.
+
+There are minor changes to existing allowance concepts in teh form of an addition so there should be no regression in
```
edward-swirldslabs, line 492: typo `teh -> the`

HIP/hip-1068.md
```diff
@@ -0,0 +1,540 @@
...
+    }
+```
+
+A new query param `?scope=(CRYPTO_TRANSFER|TOKEN_CUSTOM_FEES)` would be added to allow
+users to filter on the specific applicable scope type.
+
+## Backwards Compatibility
+
+Currently the ledger allows for sponsored transaction fees by charging the transaction submitter/payer. This HIP does
+not conflict with this flow, but rather adds an additional flow.
```
edward-swirldslabs, line 490: The exact mechanism of backwards compatibility was not specified.  I assume this is through leaving the `repeated TransactionSponsorClaim` as an empty lists?

HIP/hip-1068.md
```diff
@@ -0,0 +1,540 @@
...
+
+## How to Teach This
+
+SDK examples, blogs and tutorials on docs.hedera.com
+
+## Reference Implementation
+
+
+## Rejected Ideas
+
```
edward-swirldslabs, line 511: If daisy chained allowances was not considered, it should be considered and then explicitly rejected or accepted.

Alice gives an allowance to Bob.   Bob gives an allowance to Fred in the same scope.   Can Fred make a claim on Alice's allowance to Bob?   Why or Why not?  

If daisy chaining is allowed: 
1. Alice should be able to prevent it or allow it explicitly in her sponsorship settings. 
2. We would need to detect for loops at the time of creating sponsorship.


I suspect that we should reject this.  But its a natural concept stemming from this idea.

edward-swirldslabs: Use Case FOR daisy chaining: 
1. ORG A sponsors ORG B,  then ORG B sponsors employees of ORG B to make claims to ORG A. 

Scenario Against Daisy Chaining: 
1. A sponsors B.   C socially engineers B into sponsoring C.  C now drains A's allowance for B.

Nana-EC:
HI all,
[HIP 1195 - Hooks](https://github.com/hiero-ledger/hiero-improvement-proposals/pull/1195) has been put up in draft mode and as previously noted it is proposed as a potential cleaner and more extensible approach to handling delegation of fees amongst many other items.

The community is encouraged to give HIP 1195 a read and confirm that the interests you had for this HIP (1068) can still be satisfied with HIP 1195. If there are any gaps or questions please share those also.

FYI @tud0r6 @brendonv @app-matt

pathornteng:
I believe Hook is a very good HIP. However, I do not think it can fully replace this HIP. For many enterprise use cases, where user keys are held by a custodian, this HIP is still very important. It enables administrators to avoid concerns about individual account balances and to manage a single account as the payer, thereby significantly reducing operational overhead. Additionally, as Keith mentioned, Hook might be unnecessarily complex for many use cases.

EMerchant90:
To avoid this being stagnated and closed, dropping a comment here. 

But I am curious if there will be a way to get alerts for high-volume deviations. 

From the POV of integrations this will be important to update to partners such as wallets, bridges, and any other partners that are querying the mirror node.

piotrswierzy:
I think this HIP is great because it enables subsidizing transaction costs for users.
If we want to onboard a large number of users and have them execute transactions—such as interacting with an application, casting a DAO vote, or playing an on-chain game—this approach fits very well.
I don’t think lambdas or hooks solve this in such a clean way.
So, if the implementation isn’t overly complicated, I’d love to see this HIP deployed on mainnet.

nubeasado:
This would be great! [Fee sponsorship ](https://solana.com/developers/cookbook/transactions/fee-sponsorship)on Solana is quite useful

HIP/hip-1068.md
ty-swirldslabs: Yes, this HIP builds upon the existing allowance logic by introducing scoped allowances, this is not how the feature should be implemented and needs to be removed, great callout

HIP/hip-1068.md
```diff
@@ -0,0 +1,554 @@
...
+### Additional Allowance Types
+Initially 2 new Allowance types were proposed (`TransactionFeeAllowance` and `CustomFeeAllowance`) to capture the
+ability to sponsor transaction and token custom fees. Though functional this would result in a larger API surface
+to specify the multiple scenarios. Instead teh simple concept of a scope of applicability of each allowance was
+discussed and agreed upon. The simple scope concept is cleaner and can be added to the existing allowance types.
+
+## Open Issues
+
+- [ ] Q: How would EVM tools utilize teh ability to sponsor gas settlement fees via the the `eth_sendRawTransaction`
```
edward-swirldslabs: typo "teh" -> the

HIP/hip-1068.md
ty-swirldslabs: Daisy chaining is not allowed in this HIP iteration and has been added to the wording.

PR #1069, opened 2024-10-31
Enable KeyList Support in HTS  Precompile

mgarbs:
### Enable KeyList Support in HTS updateTokenKeys Precompile

This update adds support for using complex keys (`KeyLists` ) within the `updateTokenKeys` HTS precompile function, allowing developers to set token keys with multiple signatories directly in smart contracts. While complex keys are supported in the SDK, this enhancement allows on-chain token management with multi-signature and threshold-controlled permissions, expanding flexibility for applications requiring decentralized control over token keys.

### DIFF
--- a/HIP/hip-x.md
+++ b/HIP/hip-x.md
@@ -1,0 +1,94 @@
+---
+hip: <HIP number (assigned by the HIP editor)>
+title: Enable KeyList Support in HTS `updateTokenKeys` Precompile
+author: Michael Garber <@michael.garber>
+working-group: Giuseppe Bertone <@giuseppebertone>, Nikolaos Kamarinakis <@nikolaoskamarinakis>
+requested-by: Axelar
+type: Standards Track
+category: Service
+needs-council-approval: Yes
+status: Draft
+created: YYYY-MM-DD
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/xxx
+updated: YYYY-MM-DD
+requires: <HIP numbers, if applicable>
+replaces: <HIP numbers, if applicable>
+superseded-by: <HIP numbers, if applicable>
+---
+
+## Abstract
+This HIP proposes an enhancement to the Hedera Token Service (HTS) precompile function `updateTokenKeys` to support **KeyLists**. Although KeyLists are supported within the SDK, they cannot currently be set through the HTS precompiles, which limits their usability in smart contracts for token management. Enabling KeyList support within the HTS precompile would allow token keys managed by KeyLists to be updated on-chain, adding significant flexibility for contract-based token operations.
+
+## Motivation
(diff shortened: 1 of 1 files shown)


HIP/hip-1069.md
```diff
@@ -0,0 +1,91 @@
...
+    (responseCode) = success
+        ? abi.decode(result, (int32))
+        : HederaResponseCodes.UNKNOWN;
+}
+
+Example Usage: 
+
+IHederaTokenService.TokenKey memory key = IHederaTokenService.TokenKey({
+    keyType: KeyType.SUPPLY,
+    key: KeyList({
```
stoyanov-st, line 61: Q:  Are we aiming to support both KeyList and Threshold key types? I can see only KeyList mentioned but this example is using a Threshold type for the keyless.

mgarbs: May we get this change reviewed and approved @Neurone @stoyanov-st ?

Neurone: I cannot do it. It seems @Nana-EC and @lukelee-sl are assigned as reviewers.

Neurone:
Hi @mgarbs , it seems to me we solved all the pending issues. From my PoV we can proceed.

lukelee-sl:
From a technical point of view, I do not think this can be accommodated without system contract versioning as you would be potentially breaking existing contracts.  If this is indeed the case, then we need to decide if this feature should be included in the next HTS contract version.

stoyanov-st:
I totally agree with @lukelee-sl, this is a great candidate for the contract versioning initiative. 
Also, does it make sense to include the solidity struct changes with the HIP?

stoyanov-st:
After carefully reviewing this HIP, the smart contracts team has agreed on several discussion topics that need to be addressed for the clarity and understanding of this feature. Those are:

1. 	Solidity Structure Updates
 -	A review of existing Solidity structures is necessary to ensure compatibility with the proposed changes.
  -   Notably, due to recursion issues in KeyList we must determine whether to modify existing structures or introduce new ones to handle key management more efficiently.
 -   If we follow the protobuf definitions for the structs modification there would be additional validation for the threshold = 1 that would require an appropriate response code for the unsuccessful scenarios.
 -   Also, KeyHelper.sol needs adjustments.
2. 	Other Threshold Limitations
  -	While a threshold of 1 is the obvious limit, we need to address the more complex scenarios, such as a threshold within a threshold.
 -	Defining these limitations early will help avoid ambiguity in implementation, misinterpretations and potential security concerns.

3.  Specification of Key Usage for Signing Transactions
 -	A clear specification is needed regarding who signs transactions and which keys are used in different scenarios or a reference to an existing specification.

4.  System Contract Address
-   As this feature is going to be leveraging the System Contract Versioning, there needs to be a section stating the address that is going to be exposing it.

Considering the impact that this proposal is going to make, we need to be careful and also be able to foresee most of the implications that it is going to produce.

PR #1072, opened 2024-11-04
Update hip-904.md to Final for release 0.54.2

mgarbs:


mgarbs:
This is only a partial functionality, not yet full

PR #1079, opened 2024-11-14
HIP-991 update according to latest discussion

netopyr:
**Description**:

This PR modifies the HIP according to the latest discussion:

- Removed allowance (This will be handled in a later HIP.)
- Added `max_custom_fee` instead

### DIFF
--- a/HIP/hip-991.md
+++ b/HIP/hip-991.md
@@ -10,7 +10,7 @@ status: Accepted
 last-call-date-time: 2024-07-24T07:00:00Z
 created: 2024-06-14
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/pull/991
-updated: 2024-09-11
+updated: 2024-11-14
 ---
 
 ## Abstract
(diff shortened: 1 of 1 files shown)


HIP/hip-991.md
```diff
@@ -139,16 +123,15 @@ When creating a topic, the FEKL list is independent and must be explicitly popul
...
 
 * `MAX_CUSTOM_FEE_ENTRIES_FOR_TOPICS = 10`
 * `MAX_ENTRIES_FOR_FEE_EXEMPT_KEY_LIST = 10`
-* `MAX_FEE = unsigned int 64`
 
 ### User Flows and Interaction
 
 * Users will specify the fee settings during the topic creation process through a simple interface in their Hedera client (refer to the creation of token custom fees/fixed fee for reference).
-* Before submitting a message to a topic through an application or wallet interface, users must set an allowance and a maximum fee per message.
+* When submitting a message to a topic with custom fees through an application or wallet interface, users must set the maximum fee for the message.
```
ty-swirldslabs, line 134: to accommodate the requirement of accepting all fees, should we add a line that is something like:
* Users submitting messages can add a flag to accept all custom fees from a topic id

HIP/hip-991.md
```diff
@@ -232,6 +215,20 @@ message ConsensusCustomFeeList {
 }
 ```
 
+#### ConsensusSubmitMessageTransactionBody
+
+The `ConsensusSubmitMessageTransactionBody` message is updated to include the optional `max_custom_fees` property for specifying the maximum fee that the user is willing to pay for the message.
```
ty-swirldslabs, line 220: Do we need to call out that there's a method to set accepts all fees in this section?

HIP/hip-991.md
netopyr: Added

HIP/hip-991.md
netopyr: Added the flag `accept_all_fees`

HIP/hip-991.md
```diff
@@ -123,7 +108,7 @@ We propose adding a fixed fee mechanism to the Hedera Consensus Service (HCS) fo
   
 ##### Handling Duplicates
 
-If the FEKL list contains duplicate keys, the transaction will fail with an `FEKL_CONTAINS_DUPLICATED_KEYS` error. This ensures that duplicate entries are not silently ignored, preventing potential bugs or issues in the calling code.
+If the FEKL list contains duplicate keys, the `ConsensusUpdateTopicTransaction` transaction will fail with an `FEKL_CONTAINS_DUPLICATED_KEYS` error. This ensures that duplicate entries are not silently ignored, preventing potential bugs or issues in the calling code.
```
vtronkov, line 111: How about the duplicates in the `ConsensusCreateTopicTransaction`? This validation should apply to both create and update.

HIP/hip-991.md
vtronkov: Do we want to specify what this flag is called?

HIP/hip-991.md
```diff
@@ -139,16 +124,16 @@ When creating a topic, the FEKL list is independent and must be explicitly popul
...
 
 * `MAX_CUSTOM_FEE_ENTRIES_FOR_TOPICS = 10`
 * `MAX_ENTRIES_FOR_FEE_EXEMPT_KEY_LIST = 10`
-* `MAX_FEE = unsigned int 64`
 
 ### User Flows and Interaction
 
 * Users will specify the fee settings during the topic creation process through a simple interface in their Hedera client (refer to the creation of token custom fees/fixed fee for reference).
-* Before submitting a message to a topic through an application or wallet interface, users must set an allowance and a maximum fee per message.
+* When submitting a message to a topic with custom fees through an application or wallet interface, users must set the maximum fee for the message.
```
vtronkov, line 135: _"users **must** set the maximum fee for the message"_ - when using the word **must** it seems that the field is required but the `max_custom_fees` is optional

HIP/hip-991.md
vtronkov: It seems that the user must set either a `max_custom_fees` or set the `accept_all_custom_fees` to **true**. We can add a validation to throw an error if `accept_all_custom_fees=false` and `max_custom_fees` is not set.

vtronkov:
There are a few more mentions of "allowances" in the **Mirror Node** and **SDK Reference Implementation** sections. You can remove those.

HIP/hip-991.md
netopyr: Right, thanks. I will update the sentence.

HIP/hip-991.md
netopyr: If the topic has custom fees and `accept_all_custom_fees` is `false,` a user must set `max_custom_fees`. In all other cases, the field is ignored and is therefore optional.

HIP/hip-991.md
netopyr: I don't think this is needed. If a topic has custom fees and neither `max_custom_fees` nor `accept_all_custom_fees` are set, the message will fail anyway because the fees cannot be paid.

HIP/hip-991.md
netopyr: Added a sentence for clarification.

HIP/hip-991.md
```diff
@@ -123,7 +108,7 @@ We propose adding a fixed fee mechanism to the Hedera Consensus Service (HCS) fo
   
 ##### Handling Duplicates
 
-If the FEKL list contains duplicate keys, the transaction will fail with an `FEKL_CONTAINS_DUPLICATED_KEYS` error. This ensures that duplicate entries are not silently ignored, preventing potential bugs or issues in the calling code.
+If the FEKL list contains duplicate keys, the `ConsensusCreateTopicTransaction`, respectively `ConsensusUpdateTopicTransaction`, will fail with a `FEKL_CONTAINS_DUPLICATED_KEYS` error. This ensures that duplicate entries are not silently ignored, preventing potential bugs or issues in the calling code.
```
vtronkov, line 111: The `FEKL_CONTAINS_DUPLICATED_KEYS` can be renamed to `FEE_EXEMPT_KEY_LIST_CONTAINS_DUPLICATED_KEYS`.

netopyr: Good idea. Changed it.

vtronkov (approved):
LGTM!

steven-sheehy (approved):

Neurone (approved):

mgarbs (approved):

PR #1080, opened 2024-11-14
Update HIP-904 status and HSCS details

stoyanov-st:
**Description:**

- Update status to Final
- Update HIP-904 with the Smart Contract function selectors and structures .

### DIFF
--- a/HIP/hip-904.md
+++ b/HIP/hip-904.md
@@ -7,7 +7,8 @@ requested-by: DEXs, Wallets, External Web3 Users. Notably Reality+, Liithos, Dro
 type: Standards Track
 category: Core
 needs-council-approval: Yes
-status: Accepted
+status: Final
+release: v0.54.0
 last-call-date-time: 2024-04-08T07:00:00Z
 created: 2024-02-25
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/905
--- a/HIP/hip-904.md
+++ b/HIP/hip-904.md
@@ -11,7 +11,7 @@ status: Accepted
 last-call-date-time: 2024-04-08T07:00:00Z
 created: 2024-02-25
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/905
-updated: 2024-08-19
+updated: 2024-11-14
 replaces: 655, 777
 ---
 
--- a/HIP/hip-904.md
+++ b/HIP/hip-904.md
@@ -8,11 +8,11 @@ type: Standards Track
 category: Core
 needs-council-approval: Yes
 status: Final
-release: v0.54.0
+release: v0.56.0
 last-call-date-time: 2024-04-08T07:00:00Z
 created: 2024-02-25
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/905
-updated: 2024-11-14
+updated: 2024-11-15
 replaces: 655, 777
 ---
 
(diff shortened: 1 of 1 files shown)


steven-sheehy (approved):

jnels124 (commented):
LGTM

HIP/hip-904.md
```diff
@@ -7,11 +7,12 @@ requested-by: DEXs, Wallets, External Web3 Users. Notably Reality+, Liithos, Dro
 type: Standards Track
 category: Core
 needs-council-approval: Yes
-status: Accepted
+status: Final
+release: v0.54.0
```
Nana-EC, line 11: Should be v0.56 as that's when the system contract functions went in.
Agreed it's 0.54 for when HAPI functionality was completed but it might be confusing if users test this using local node or solo 😄

HIP/hip-904.md
```diff
@@ -1170,21 +1171,50 @@ HTS system contracts will need to support the new airdrop related transactions t
...
+| `0x2f348119`           | `function airdropTokens(TokenTransferList[] memory tokenTransfers) external returns (int64 responseCode)`                                      | TokenAirdrop     | `ResponseCode` | The response code from the call |
+| `0x012ebcaf`           | `function cancelAirdrops(PendingAirdrop[] memory pendingAirdrops) external returns (int64 responseCode)`                                       | TokenCancel      | `ResponseCode` | The response code from the call |
+| `0x05961641`           | `function claimAirdrops(PendingAirdrop[] memory pendingAirdrops) external returns (int64 responseCode)`                                        | TokenClaim       | `ResponseCode` | The response code from the call |
+| `0xebd595e0`           | `function rejectTokens(address rejectingAaddress, address[] memory ftAddresses,  NftID[] memory nftIDs) external returns (int64 responseCode)` | TokenReject      | `ResponseCode` | The response code from the call |
+
+New system contract functions to be included for a new `IHRC904` interface.
```
Nana-EC, line 1183: Please add a note here that the facade/proxy logic is used and as such the contract address used should be the Hedera Entity on note e.g. the HTS token address or the Account EVM address.

Nana-EC (commented):
2 suggestions for completion.
Thanks

lukelee-sl (approved):
LGTM

HIP/hip-904.md
```diff
@@ -1170,21 +1171,50 @@ HTS system contracts will need to support the new airdrop related transactions t
...
+
+`PendingAirdrop` - A struct that represents a pending airdrop request.
+
+```solidity
+struct PendingAirdrop {
+    address sender;
+    address receiver;
+
+    address token;
+    int64 serial;
```
david-bakin-sl, line 1206: It seems totally obvious that `serial` is only for NFTs - but still, maybe a comment that addresses that (and states explicitly that that depends on the token type addressed by the `token` field)?  (Reason is: It's a little odd that the very next lines introduce `NftID` for identifying "the Nft serial to be rejected" but there's nothing similar for fungible types (the address only?) and anyway it's not used here in `PendingAirdrop` because that's the way solidity works: No union types.). And must `serial == 0` for fungible tokens or is it a don't care?

HIP/hip-904.md
stoyanov-st: That makes sense. I copied it from the previous PR.

HIP/hip-904.md
stoyanov-st: Added the note.

HIP/hip-904.md
stoyanov-st: Added a comment for clarification.
For the latter comment, the stated functions are redirect calls that already have the token address so it makes it obsolete to use `NftID` there. The `address` parameters there are the token airdrop sender/receiver as stated in the naming.

david-bakin-sl: ah! my mistake

david-bakin-sl (approved):
lgtm

steven-sheehy (approved):

Nana-EC (approved):
LG

mgarbs (approved):

PR #1082, opened 2024-11-18
Allow Setting of EVM Address by EOA

Nana-EC:
**Description**:
HIP 583 opened the doors for greater account compatibility with EVM flows by utilizing the account alias to encompass
the EVM address. However, given the static nature of the alias and concerns of complexity, existing accounts were not
allowed to update their alias and set one if it was null. This HIP aims to rectify this by allowing EOAs who have no
alias set to set it to an ECDSA key derived evm address value they own.

This HIP also replaces the need for [HIP 631: Account Virtual Addresses](https://hips.hedera.com/hip/hip-631) as it
aimed to resolve this issue as well as provide even greater functionality. A form of HIP 631 could be revisited in the
future but not with the goal of providing address equivalence with the EVM.

**Related issue(s)**:

Fixes #1070 

**Notes for reviewer**:
1st draft, need to add a clarifying image and touch up

**Checklist**

### DIFF
--- a/HIP/hip-0000-set-evm-addres-on-eoa.md
+++ b/HIP/hip-0000-set-evm-addres-on-eoa.md
@@ -1,0 +1,210 @@
+---
+hip: <HIP number (assigned by the HIP editor), usually the PR number>
+title: Allow Setting of EVM Address by EOA
+author: Nana Essilfie-Conduah <@Nana-EC>
+working-group: Richard Bair <@rbair>, Atul Mahamuni <@atul-hedera>, Ty Smith <ty-swirldslabs>,
+Serg Metelin <@sergmetelin>, Ali Nikan <@alinik4n>, Greg Scullard <@gregscullard>
+requested-by: Hedera community
+type: Standards Track
+category: Service
+needs-council-approval: Yes
+status: Draft
+created: 11-18-24
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/1070
+updated: 11-18-24
+replaces: 631
+requires: 583
+---
+
+## Abstract
+
+HIP 583 opened the doors for greater account compatibility with EVM flows by utilizing the account alias to encompass
+the EVM address. However, given the static nature of the alias and concerns of complexity, existing accounts were not
+allowed to update their alias and set one if it was null. This HIP aims to rectify this by allowing EOAs who have no
+alias set to set it to an ECDSA key derived evm address value they own.
+
+This HIP also replaces the need for [HIP 631: Account Virtual Addresses](https://hips.hedera.com/hip/hip-631) as it
+aimed to resolve this issue as well as provide even greater functionality. A form of HIP 631 could be revisited in the
+future but not with the goal of providing address equivalence with the EVM.
+
+## Motivation
+
+The EVM address alias introduced in HIP 583 allowed for ECDSA based accounts to interact with the EVM in the same way
+other EVM chains do. Particularly, functionality like the `ecrecover` that relies on the relationship between key and
+address (pivotal to many DEXs) was unblocked and allowed greater EVM equivalence on the Hedera network.
+However, HIP 583 intentionally provided no pathway for accounts with no alias regardless of key type (ED, EC or
+complex keys). Though these accounts were afforded the hedera num account alias (long zero), this format would not
+allow accounts to utilize the `ecrecover` precompile in the expected manner.
+This continued incompatibility has remained and inhibited web3 EVM tooling support.
+
+Providing a solution to this gap is a necessity to make it easy for EVM dApps in the web3 community to come to 
+Hedera without needing to change their smart contracts.
+
+## Rationale
+
+[HIP 631: Account Virtual Addresses](https://hips.hedera.com/hip/hip-631) lays out the challenges faced by accounts
+without an ECDSA derived EVM address that want to interact with the EVM in a way that complies with the Hedera native
+model and the EVM model. However, HIP 631 offered feature rich complexity that may be difficult for DApps, wallets and
+users to follow completely.
+
+Instead of the full HIP 631, an intermediate approach i.e. a HIP 631 Lite (this HIP) could unblock users without an
+EVM address. It would enable accounts to interact with the EVM in a compliant manner regardless of the public key on
+its account.
+
+In a sense the current network already supports the ability to set the evm address, it just requires the logic to be 
+set on account creation. Thus by lifting the limit of not being able to set the alias after account creation, whiles
+still requiring that the current alias be unset this HIP ushers in the ability of those previously blocked accounts to
+interact with the EVM as expected.
+
+### Foundational Concepts
+
+The following are some clarifying concepts to increase clarity around aliases
+- Every account on the network has 1 or 2 evm addresses. Either only the Hedera num alias (long zero) or the Hedera num
+    alias (long zero) and an EVM address alias (standard ECDSA derived EVM address).
+- When a user interacts with the EVM the network will utilize the ECDSA derived EVM address alias on an account if
+    present. If not it will utilize the long zero.
+- Accounts with an ECDSA derived EVM address are fully compatible with expectations of contracts that utilize the
+    ecrecover precompile.
+- Accounts without an ECDSA derived EVM address alias will not match the expectation of the ecrecover precompile
+    calculations. This has effects on certain DApp behaviour.
+- Smart contracts can and do cache the evm address of an account and often use it to assign attributes such as access
+    and ownership. As a result it was important to maintain a static alias to avoid unintended differences in
+    expectation
+- Account key rotation has no impact on the evm address utilized when an account interacts with the EVM
+
+<aside>
+🚨 **Open Task:** : Add account, address key relationship diagram for clarity
+</aside>
+
+## User stories
+
+1. As an existing account with an ED key but no EVM address alias, I would like to set an ECDSA derived EVM address
+    alias on my account that will identify my account on the EVM and allow for the correct `ecrecover` functionality.
+2. As an existing account with an EC key but no EVM address alias, I would like to set an ECDSA derived EVM address
+    alias on my account that will identify my account on the EVM and allow for the correct `ecrecover` functionality.
+3. As an existing account with a complex key but no EVM address alias, I would like to set an ECDSA derived EVM address
+    alias on my account that will identify my account on the EVM and allow for the correct `ecrecover` functionality.
+4. As an account with an EVM address alias set after creation, I would like to set an evm address override to my
+    previous long zero address for a `ContractCreate` transaction.
+5. As an account with an EVM address alias set after creation, I would like to set an evm address override to my
+    previous long zero address for a `ContractCall` transaction.
+  
+## Specification
+
+The main change to the network is simple - `CryptoUpdate` transactions will support setting a new ECDSA derived
+`evm_address` if the current value is unset.
+
+### Protobuf
+```protobuf
+message CryptoUpdateTransactionBody {
+    ...
+
+    /**
+     * EOA 20-byte address to create that is derived from the keccak-256 hash of a ECDSA_SECP256K1 primitive key.
+     */
+    bytes evm_address = 19;
+}
+```
+
+Additionally, `ContractCreate` and `ContractCall` EVM transactions will need to support the ability to specify the
+a Hedera Num Alias alias EVM address to use on the transaction. This is because on the HAPI side all aliases are
+applicable and thus the flexibility of the API is important to maintain.
+Also, accounts that previously had no ECDSA derived EVM address were likely created via the HAPI flow and it may
+be important for an EOA to utilize the long zero address even after setting a new ECDSA derived EVM address.
+For transactions that utilize the `EthereumTransaction` a compliant ECDSA derived EVM address is provided by default
+when using EVM tools.
+
+```protobuf
+message ContractCreateTransactionBody {
+    ...
+    /**
+    * The 20-byte EVM address to use for the given transaction. This address must be the Hedera num alias of the account.
+    */
+    bytes evm_address_override = 21;
+}
+
+message ContractCallTransactionBody {
+	...
+    /**
+    * The 20-byte EVM address to use for the given transaction. This address must be the Hedera num alias of the account.
+    */
+    bytes evm_address_override = 5;
+}
+```
+
+No change is needed to expose the evm address as it is covered by the `evm_address` property on `TransactionRecord`
+
+### HSCS System Contract API
+The HAS system contract will need to be updated to add support for a `setEVmAddressAlias()` function to expose the
+`CryptoUpdate` HAPI functionality and allow EOAs to call this logic directly from dApps.
+
+### SDK
+
+SDKs will need to update the `AccountUpdateTransaction` to support the `setAlias()` function as it does on
+`AccountCreateTransaction`. This will populate the `evm_address` in the `CryptoUpdateTransactionBody`to be submitted.
+
+`ContractCreateTransaction` and `ContractExecuteTransaction` will need to support a new `setEVmAddressOverride()`
+function to set the `evm_address_override` property on the respective `ContractCreate` and `ContractCall` transactions. 
+
+### Mirror Node
+
+Mirror Nodes should be sure to support the setting of the evm address value on an account when processing a
+`CryptoUpdate` transaction. As in the case of a `CryptoCreate` the new value can be retrieved from the `evm_address`
+property on `TransactionRecord`.
+If no custom logic already exists to ignore this Mirror Nodes will be unimpaired by the change
+
+### Wallets
+
+The HIP provides opportunities for increased user functionality.
+- Wallets are encouraged to create accounts with an ECDSA derived address alias
+- Wallets are encouraged to support the ability for users to set an ECDSA derived alias in the future. This may require
+wallets to support multiple keys for users that may decide to interchange between addresses.
+
+## Backwards Compatibility
+
+This does not break previous functionality but rather adds new scoped support for a change that is possible at account
+creation time. However, it may be said that there is a behaviour change here that is important to note. This change
+is accepted as the benefits of unblocking existing accounts outweighs the cons. The ability to override the address
+also ensure there's a pathway for EOA to set their long zero address if needed.
+Finally, it is appropriate for clients to query mirror node for accurate account details. Clients that do this will
+avoid not be impacted by this change.
+
+## Security Implications
+
+Smart contracts have their logic self contained in byte code. Therefore, the implications of address changes are
+dictated by each smart contract and not the network. A smart contract may cache an EOA address on first interaction and
+utilize that to identify them in future. The cached address could dictate balance access or other privileged access.
+
+As such, an EOA that has interacted with a smart contract using their long zero address and then sets an ECDSA derived
+EVM address after will be responsible for any unexpected interactions with the smart contract
+
+## How to Teach This
+
+SDK examples, blogs and tutorials on docs.hedera.com
+
+## Reference Implementation
+
+
+## Rejected Ideas
+
+### Implement HIP 631
+[HIP 631: Account Virtual Addresses](https://hips.hedera.com/hip/hip-631) would have added the ability add evm address
+and specify any one of them as the desired address for use. However, this didn't unblock existing account and offered
+likely complex UX that would overcomplicated the issue at this time. Focusing on setting aliases that were already null
+and supporting evm address only for accounts that migrated themselves provides a solution that is focused and less
+complex.
+
+## Open Issues
+
+While a HIP is in draft, ideas can come up which warrant further discussion. Those ideas should be recorded so people
+know that they are being thought about but do not have a concrete resolution. This helps make sure all issues required
+for the HIP to be ready for consideration are complete and reduces people duplicating prior discussions.
+
+## References
+
+
+
+## Copyright/license
+
+This document is licensed under the Apache License, Version 2.0 -- see [LICENSE](../LICENSE) or
+(https://www.apache.org/licenses/LICENSE-2.0)


HIP/hip-1082.md
```diff
@@ -0,0 +1,210 @@
...
+avoid not be impacted by this change.
+
+## Security Implications
+
+Smart contracts have their logic self contained in byte code. Therefore, the implications of address changes are
+dictated by each smart contract and not the network. A smart contract may cache an EOA address on first interaction and
+utilize that to identify them in future. The cached address could dictate balance access or other privileged access.
+
+As such, an EOA that has interacted with a smart contract using their long zero address and then sets an ECDSA derived
+EVM address after will be responsible for any unexpected interactions with the smart contract
```
gregscullard, line 179: Maybe add a suggestion that such accounts should first transfer all contract assets to the EVM address, then update the account's EVM address ?

HIP/hip-1082.md
```diff
@@ -0,0 +1,210 @@
...
+    expectation
+- Account key rotation has no impact on the evm address utilized when an account interacts with the EVM
+
+<aside>
+🚨 **Open Task:** : Add account, address key relationship diagram for clarity
+</aside>
+
+## User stories
+
+1. As an existing account with an ED key but no EVM address alias, I would like to set an ECDSA derived EVM address
```
ty-swirldslabs, line 81: In this user story, wouldn't the user have to update the ED key to the new ECDSA key that corresponds with the EVM address for 'ecrecover' to function? Maybe I'm misunderstanding the phrase 'correct ecrecover functionality', or is the implication that the key is being updated as well

HIP/hip-1082.md
```diff
@@ -0,0 +1,210 @@
...
+    alias on my account that will identify my account on the EVM and allow for the correct `ecrecover` functionality.
+2. As an existing account with an EC key but no EVM address alias, I would like to set an ECDSA derived EVM address
+    alias on my account that will identify my account on the EVM and allow for the correct `ecrecover` functionality.
+3. As an existing account with a complex key but no EVM address alias, I would like to set an ECDSA derived EVM address
+    alias on my account that will identify my account on the EVM and allow for the correct `ecrecover` functionality.
```
ty-swirldslabs, line 86: Same concern here as user story 1. Maybe we add the caveat to the stories, that the ECDSA signature that corresponds with the EVM address for ecrecover to function would need to be utilized on signatures for EVM of those transactions and not the key associated with the account on the hedera network.

HIP/hip-1082.md
```diff
@@ -0,0 +1,210 @@
...
+4. As an account with an EVM address alias set after creation, I would like to set an evm address override to my
+    previous long zero address for a `ContractCreate` transaction.
+5. As an account with an EVM address alias set after creation, I would like to set an evm address override to my
+    previous long zero address for a `ContractCall` transaction.
```
ty-swirldslabs, line 90: Is this out of scope / too complex for this HIPs user story?
As an existing account with an ED key but no EVM address alias, I would like to update my key to an ECDSA key and set my EVM address to be derived from the new ECDSA key.

ty-swirldslabs (commented):
Great work! I added some clarifying questions to user stories section

HIP/hip-1082.md
Nana-EC: No, the user would just use their ECDSA key in a wallet to sign. So far as the EVM address can be extracted from the signature the ecercover flow would work

HIP/hip-1082.md
Nana-EC: Out of scope, not too complex build would be enabled by this HIP and current network functionality.
Such a user could rotate their keys before or after setting the EVM address.
The could actually do it in one single CryptoUpdate also.
Key rotation is not tied to alias value

HIP/hip-1082.md
```diff
@@ -0,0 +1,215 @@
...
+created: 11-18-24
+discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/1070
+updated: 12-06-24
+replaces: 631
+requires: 583
+---
+
+## Abstract
+
+HIP 583 opened the doors for greater account compatibility with EVM flows by utilizing the account alias to encompass
```
hendrikebbers, line 21:
```suggestion
[HIP 583](https://hips.hedera.com/hip/hip-583) opened the doors for greater account compatibility with EVM flows by utilizing the account alias to encompass
```

HIP/hip-1082.md
```diff
@@ -0,0 +1,215 @@
...
+
+    /**
+     * EOA 20-byte address to create that is derived from the keccak-256 hash of a ECDSA_SECP256K1 primitive key.
+     */
+    bytes evm_address = 19;
+}
+```
+
+Additionally, `ContractCreate` and `ContractCall` EVM transactions will need to support the ability to specify the
+a Hedera Num Alias alias EVM address to use on the transaction. This is because on the HAPI side all aliases are
```
hendrikebbers, line 113: "a Hedera Num Alias alias EVM addres"
Is that correct?

HIP/hip-1082.md
IvanKavaldzhiev: Do you mean a scenario when an account has a KeyList containing existing ED key and a new ECDSA key is added that is added there and is linked to this account.

How this wallet ECDSA key will be connected to the account?

HIP/hip-1082.md
IvanKavaldzhiev, line 113: typo: a

Nana-EC: Fixed

Nana-EC: fixed

HIP/hip-1082.md
Nana-EC: Fixed

HIP/hip-1082.md
```diff
@@ -0,0 +1,215 @@
...
+updated: 12-06-24
+replaces: 631
+requires: 583
+---
+
+## Abstract
+
+HIP 583 opened the doors for greater account compatibility with EVM flows by utilizing the account alias to encompass
+the EVM address. However, given the static nature of the alias and concerns of complexity, existing accounts were not
+allowed to update their alias and set one if it was null. This HIP aims to rectify this by allowing EOAs who have no
```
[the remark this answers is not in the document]
Nana-EC: Updated

HIP/hip-1082.md
Nana-EC: Updated

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Protobuf
+```protobuf
+message CryptoUpdateTransactionBody {
+    ...
+
+    /**
+     * EOA 20-byte address to create that is derived from the keccak-256 hash of a ECDSA_SECP256K1 primitive key.
+     */
+    bytes evm_address = 19;
```
rbair23, line 109:
```suggestion
    /**
     * An EVM-compatible 20-byte address derived the keccak-256 hash of a ECDSA_SECP256K1 primitive key.
     * The transaction MUST be signed by the corresponding private key, and the `alias` on the account MUST
     * be unset prior to this call. Once set, this value cannot be changed.
     */
    bytes alias = 19;
```

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+}
+
+message ContractCallTransactionBody {
+	...
+    /**
+    * The 20-byte EVM address to use for the given transaction. This address must be the Hedera num alias of the account.
+    */
+    bytes evm_address_override = 5;
+}
+```
```
rbair23, line 137:
```suggestion
The `ContractCall` transaction will allow the user to specify an `evm_address_override` to explicitly specify
the use of the long-zero EVM address to represent the account in the transaction.

```protobuf
message ContractCallTransactionBody {
	...
    /**
     * The 20-byte EVM address to use for the given transaction. This address must be the Hedera "long-zero" alias
     *  of the account making this call.
     */
    bytes evm_address_override = 5;
}
```
```

File: HIP/hip-1082.md [hunk 10]
```diff
@@ -0,0 +1,225 @@
...
+
+No change is needed to expose the evm address as it is covered by the `evm_address` property on `TransactionRecord`
```
rbair23, line 139: We have both `evm_address` and `alias` in the `TransactionRecord`. The documentation for `evm_address` is suspicious:

> The new default EVM address of the account created by this transaction. This field is populated only when the EVM address is not specified in the related transaction body

I'm not exactly sure what that means. We should double check on whether we write the evm_address, alias, or both in the transaction record after an update. And we should move this comment into the previous section for `CryptoUpdateTransactionBody`.

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+
+### HSCS System Contract API
+The HAS system contract will need to be updated to add support for a `setEVmAddressAlias()` function to expose the
+`CryptoUpdate` HAPI functionality and allow EOAs to call this logic directly from dApps.
+
```
rbair23, line 144:
```suggestion
The Hedera Account Service (HAS) system contract will need to be updated to add support for a `setEvmAddressAlias()` function to expose the
`CryptoUpdate` HAPI functionality and allow EOAs to call this logic directly from dApps.

```

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+function to set the `evm_address_override` property on the respective `ContractCreate` and `ContractCall` transactions. 
+
+In addition to updated `AccountUpdateTransaction` functionality the SDK will want to add new methods to streamline
+account key specifications by developers with considerations for alias
+
+- `setKeyWithAlias(ECDSAKey)` - Explicitly calls out the alias set and converts public key to address in the background
+- `setKeyWithAlias(Key, ECDSAKey)` - Allows for setting the account key and a separate key that the evmAddress should
+    be derived from. A user must sign the transaction with both keys for this flow to be successful.
+- `setKeyWithoutAlias(Key)` - Explicitly calls out that the alias is not set. This is useful for wallets that create
+    accounts with default keys and rotate keys later.
```
rbair23, line 160:
```suggestion
SDKs will need to update the `AccountUpdateTransaction` to support the `setAlias()` function as it does on
`AccountCreateTransaction`. This will populate the `evm_address` in the `CryptoUpdateTransactionBody`to be submitted.

`ContractExecuteTransaction` will need to support a new `setEVmAddressOverride()`
function to set the `evm_address_override` property on the `ContractCall` transaction. 

In addition to updated `AccountUpdateTransaction` functionality the SDK will want to add new methods to
streamline account key specifications by developers with considerations for alias:

- `setKeyWithAlias(ECDSAKey)` - Explicitly calls out the alias will be set and converts the public key to an
   address in the background
- `setKeyWithAlias(Key, ECDSAKey)` - Allows for setting the account key and a separate key that the evmAddress
   should be derived from. A user must sign the transaction with both keys for this flow to be successful.
- `setKeyWithoutAlias(Key)` - Explicitly calls out that the alias is not set. This is useful for wallets that create
    accounts with default keys and rotate keys later.
```

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Mirror Node
+
+Mirror Nodes should be sure to support the setting of the evm address value on an account when processing a
+`CryptoUpdate` transaction. As in the case of a `CryptoCreate` the new value can be retrieved from the `evm_address`
+property on `TransactionRecord`.
+If no custom logic already exists to ignore this Mirror Nodes will be unimpaired by the change
```
rbair23, line 167:
```suggestion
Mirror Nodes should be sure to support the setting of the evm address / alias value on an account when processing
a `CryptoUpdate` transaction. As in the case of a `CryptoCreate` the new value can be retrieved from the `evm_address` property on `TransactionRecord`.
If no custom logic already exists to ignore this Mirror Nodes will be unimpaired by the change
```

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+wallets to support multiple keys for users that may decide to interchange between addresses.
+
+## Backwards Compatibility
+
+This does not break previous functionality but rather adds new scoped support for a change that is possible at account
+creation time. However, it may be said that there is a behaviour change here that is important to note. This change
+is accepted as the benefits of unblocking existing accounts outweighs the cons. The ability to override the address
+also ensure there's a pathway for EOA to set their long zero address if needed.
+Finally, it is appropriate for clients to query mirror node for accurate account details. Clients that do this will
+avoid not be impacted by this change.
```
rbair23, line 183:
```suggestion
Any account interacting with a smart contract using its long-zero address, that subsequently sets an ECDSA derived alias, MAY need to use the `evm_address_override` when working with that smart contract,
if the smart contract maintained a reference to the account. Alternatively, the user may call the smart
contract using the `evm_address_override` to move assets or data contained within the smart contract
from the long-zero address to their new ECDSA-derived address.
```

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+## Security Implications
+
+Smart contracts have their logic self contained in byte code. Therefore, the implications of address changes are
+dictated by each smart contract and not the network. A smart contract may cache an EOA address on first interaction and
+utilize that to identify them in future. The cached address could dictate balance access or other privileged access.
+
+As such, an EOA that has interacted with a smart contract using their long zero address and then sets an ECDSA derived
+EVM address after will be responsible for any unexpected interactions with the smart contract.
+It is advised that an EOA in this case first transfer all in contract storage assets to the EVM address prior to
+setting their alias.
```
rbair23, line 194:
```suggestion
None.
```

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+
+## How to Teach This
+
+SDK examples, blogs and tutorials on docs.hedera.com
```
rbair23, line 198:
```suggestion
SDK examples, blogs and tutorials on docs.hedera.com. We recommend updating all samples and tutorials to use
ECDSA derived aliases in all cases with the matching ECDSA key being used as the account admin key, as this is
the expected normal flow. We strongly recommend only **advanced** users set the alias on existing accounts,
and only once their wallets support it. Otherwise, we recommend setting an alias on an existing account only be
undertaken by wallets or dapps that pre-create accounts but never use them, until given to a user where the alias
and key are then set.
```

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+
+
+## Rejected Ideas
+
+### Implement HIP 631
+[HIP 631: Account Virtual Addresses](https://hips.hedera.com/hip/hip-631) would have added the ability add evm address
+and specify any one of them as the desired address for use. However, this didn't unblock existing account and offered
+likely complex UX that would overcomplicated the issue at this time. Focusing on setting aliases that were already null
+and supporting evm address only for accounts that migrated themselves provides a solution that is focused and less
+complex.
```
rbair23, line 210: Is this true? I thought 631 allowed having multiple aliases, and thus would support going from null to an alias as well. The only problem with 631 was that it was complicated by having multiple aliases, and we decided to try a simpler approach with just 1 alias (in addition to the long-zero).

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Open Issues
+
+While a HIP is in draft, ideas can come up which warrant further discussion. Those ideas should be recorded so people
+know that they are being thought about but do not have a concrete resolution. This helps make sure all issues required
+for the HIP to be ready for consideration are complete and reduces people duplicating prior discussions.
```
rbair23, line 216: Either we should remove this or list the open issues.

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+
+## References
+
+
```
rbair23, line 220: It would be good to add the references to the other HIPs.

stoqnkpL:
IMO its important to consider two things:
1. From an ecosystem perspective, having a single real-world user appear as multiple distinct participants can break the expected user flows. DApps, wallets, and explorers might show contradictory data about the user’s “identity,” transaction history, or holdings. For example, if I own some amount of token X as my long-zero address, and some amounts of token Y as my newly-set alias, my wallet should be able to detect those and display both, then know how to generate transactions that operate with those balances. Explorers should also index events from both addresses and associate them to the same account.
3. The EVM should handle edge-case scenarios that will now be possible where for example I attempt to transfer HBAR from my long-zero address to my new alias address, both pointing to the same hedera account.

HIP/hip-1082.md
Nana-EC: [HIP 631](https://hips.hedera.com/hip/hip-631) specifically reserved a different space in state for the multiple virtual addresses but maintained a single account.alias field as immutable

> With Virtual Addresses, keys remain the signing permission authority and can be changed but Virtual Addresses (not tied to the alias) provide the value for a transactions EVM address.

Pulled from HIP 631, also the account diagram notes virtual addresses as a different field altogether. We never stated the ability to set a the null alias, so even with multiple virtual addresses an account could still have a null alias

Perseverance:
I agree with this HIP's proposed solution and believe it is the right way to address many possible issues.

However, I would like to highlight that the rationale and, therefore, the impact of this change are likely less severe than they appear.

Firstly, I strongly believe that the `ecrecover` precompile should not be modified and its behaviour should be kept the same as Ethereum. `ecrecover` is meant to be a pure cryptographic function - a stateless black-box one. The rationale mentions that this HIP will enable `ecrecover` to "work". I believe that `ecrecover` works exactly as intended right now.

The real reasoning behind this HIP, for me, comes from the use-cases of `ecrecover` rather than the inner workings of `ecrecover`. Many Ethereum applications use "signed-message"-based authorization - Based on a signed message by the user, the dapps enable certain actions to be authorized on the behalf of the signing user. The most famous example is the ERC20 add-on - ERC-2612 - commonly known as "permits". In this use-case the owner of certain amount of tokens, signs a message authorizing another user to use them.

This HIP solution will help enable this use case, but it is **not** because `ecrecover` is somehow broken, but because it will enable our users with non-ecdsa keys to set their ecdsa representing addresses, therefore be eligible to use this authorization scheme.

In respect of fairness, it is important to highlight the negative side-effects of this change. As @rbair23 pointed, this change, and our aim to support the "signed-message"-based authorization is in fact leading to hard-coupling between user and their ECDSA keys. This disables the much desired key rotation.

Ferparishuertas:
From an evm perspective, this is the most compatible version, and allows the signed message schemes at smart contract level. That said as @rbair23 and @Perseverance mention, this intermediate HIP is a step forward,  but we need to keep an eye on HIP -631, to enable alias rotation, or multiple evm adressess alias to  identity/account mappings. 

Apologizes in forehand for next comments, if there is anything out of context.

As a newbie, I want to suggest some extra adds to the context to make the HIP more comprehensive, added to a question

- Make it clear, that the alias (derived evm address) is unique and cannot be used within more than 1 account
- Separate the concepts of key rotating and alias. Documentation indicates the following:

 "If an alias is set during account creation, it becomes [immutable](https://docs.hedera.com/hedera/support-and-community/glossary#immutability), meaning it cannot be changed. If you plan to update or rotate keys in the future, do not set the alias at the time of initial account creation. The alias can be set after finalizing all key updates."

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
+---
+hip: 1082
+title: Allow Setting of EVM Address by EOA
+author: Nana Essilfie-Conduah <@Nana-EC>
+working-group: Richard Bair <@rbair>, Atul Mahamuni <@atul-hedera>, Ty Smith <ty-swirldslabs>,
+Serg Metelin <@sergmetelin>, Ali Nikan <@alinik4n>, Greg Scullard <@gregscullard>
```
steven-sheehy, line 6: Should stay on one line so github doesn't complain about invalid yaml when rendering.

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+replaces: 631
+requires: 583
+---
+
+## Abstract
+
+[HIP 583](https://hips.hedera.com/hip/hip-583) opened the doors for greater account compatibility with EVM flows by
+utilizing the account alias to encompass
+the EVM address. However, given the static nature of the alias and concerns of complexity, existing accounts were not
+allowed to update their alias and set one if it was `null`. This HIP aims to rectify this by allowing EOAs who have no
```
steven-sheehy, line 24: null is a bit of tech speak

```suggestion
allowed to update their alias and set one if it was missing. This HIP aims to rectify this by allowing EOAs who have no
```

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+For transactions that utilize the `EthereumTransaction` a compliant ECDSA derived EVM address is provided by default
+when using EVM tools.
+
+```protobuf
+message ContractCreateTransactionBody {
+    ...
+    /**
+    * The 20-byte EVM address to use for the given transaction. This address must be the Hedera num alias of the account.
+    */
+    bytes evm_address_override = 21;
```
steven-sheehy, line 127: This name seems ambiguous to me. It could be the address of the contract or the sender. What about `from` like in `eth_call`? Or other alternatives like `from_address` or `sender`?

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+- `setKeyWithAlias(ECDSAKey)` - Explicitly calls out the alias set and converts public key to address in the background
+- `setKeyWithAlias(Key, ECDSAKey)` - Allows for setting the account key and a separate key that the evmAddress should
+    be derived from. A user must sign the transaction with both keys for this flow to be successful.
+- `setKeyWithoutAlias(Key)` - Explicitly calls out that the alias is not set. This is useful for wallets that create
+    accounts with default keys and rotate keys later.
+
+### Mirror Node
+
+Mirror Nodes should be sure to support the setting of the evm address value on an account when processing a
+`CryptoUpdate` transaction. As in the case of a `CryptoCreate` the new value can be retrieved from the `evm_address`
```
steven-sheehy, line 165: This doesn't mention changes around processing contract call and create.

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+    /**
+     * EOA 20-byte address to create that is derived from the keccak-256 hash of a ECDSA_SECP256K1 primitive key.
+     */
+    bytes evm_address = 19;
+}
+```
+
+Additionally, `ContractCreate` and `ContractCall` EVM transactions will need to support the ability to specify the
+Hedera Num Alias address to use on the transaction. This is because on the HAPI side all aliases are applicable and
+thus the flexibility of the API is important to maintain.
```
steven-sheehy, line 115: I still don't understand why the HIP allows them to set an ECDSA derived EVM address on their existing account but then it doesn't allow them to use this new value on contract calls and instead requires the long zero form.

And even if Hedera Num Alias is required, why would they specify it as long zero bytes instead of just using `AccountID` in shard.realm.num?

HIP/hip-1082.md
steven-sheehy, line 165: Can we detail any API changes? My assumption that there's no new fields but that existing fields in `/api/v1/accounts/{id}` will show the updated evm_address and the existing `from` in contract result APIs will show the override?

steven-sheehy (changes requested):

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+
+[HIP 583](https://hips.hedera.com/hip/hip-583) opened the doors for greater account compatibility with EVM flows by
+utilizing the account alias to encompass
+the EVM address. However, given the static nature of the alias and concerns of complexity, existing accounts were not
+allowed to update their alias and set one if it was `null`. This HIP aims to rectify this by allowing EOAs who have no
+alias set to set it to an ECDSA key derived evm address value they own.
+
+This HIP also replaces the need for [HIP 631: Account Virtual Addresses](https://hips.hedera.com/hip/hip-631) as it
+aimed to resolve this issue as well as provide even greater functionality. A form of HIP 631 could be revisited in the
+future but not with the goal of providing address equivalence with the EVM.
```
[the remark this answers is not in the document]
Nana-EC: I think the context is still valuable. A reader of a HIP should be able to read the Abstract and get a general idea of the sections of the spec, for discussion. 
I adopted the 1 liner to give the answer to"What" but preserved the 2 short paragraphs to allow the idea to be presented.

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+The EVM address alias introduced in HIP 583 allowed for ECDSA based accounts to interact with the EVM in the same way
+other EVM chains do. Particularly, functionality like the `ecrecover` that relies on the relationship between key and
+address (pivotal to many DEXs) was unblocked and allowed greater EVM equivalence on the Hedera network.
+However, HIP 583 intentionally provided no solution for accounts with no alias regardless of key type (ED, EC or
+complex keys). Though these accounts were afforded the hedera num account alias (long zero), this format would not
+allow accounts to utilize the `ecrecover` precompile in the expected manner.
+This continued incompatibility has remained and inhibited web3 EVM tooling support.
+
+Providing a solution to this gap is a necessity to make it easy for EVM dApps in the web3 community to come to 
+Hedera without needing to change their smart contracts.
```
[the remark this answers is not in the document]
Nana-EC: Adopted change

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+users to follow completely.
+
+Instead of the full HIP 631, an intermediate approach i.e. a HIP 631 Lite (this HIP) could unblock users without an
+EVM address. It would enable accounts to interact with the EVM in a compliant manner regardless of the public key on
+its account.
+
+In a sense the current network already supports the ability to set the evm address, it just requires the logic to be 
+done on account creation. Thus by lifting the limit of not being able to set the alias after account creation, whiles
+still requiring that the current alias be unset this HIP ushers in the ability for those previously blocked accounts to
+interact with the EVM as expected.
```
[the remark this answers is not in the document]
Nana-EC: Adopted change

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+- When a user interacts with the EVM the network will utilize the ECDSA derived EVM address alias on an account if
+    present. If not it will utilize the long zero.
+- Accounts with an ECDSA derived EVM address are fully compatible with expectations of contracts that utilize the
+    ecrecover precompile.
+- Accounts without an ECDSA derived EVM address alias will not match the expectation of the ecrecover precompile
+    calculations. This has effects on certain DApp behaviour.
+- Smart contracts can and do cache the evm address of an account and often use it to assign attributes such as access
+    and ownership. As a result it was important to maintain a static alias to avoid unintended differences in
+    expectation
+- Account key rotation has no impact on the evm address utilized when an account interacts with the EVM
```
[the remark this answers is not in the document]
Nana-EC: Adopted change

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+3. As an existing account with a complex key but no EVM address alias, I would like to set an ECDSA derived EVM address
+    alias on my account that will identify my account on the EVM.
+5. As an existing account with a newly set ECDSA derived EVM address, I want my account to conform with `ecrecover`
+    expectations as long as I sign the transaction with the ECDSA key that corresponds to the EVM address alias.
+    This key may or may not be the key currently on my account.
+5. As an account with an EVM address alias set after creation, I would like to set an evm address override to my
+    previous long zero address for a `ContractCreate` transaction.
+6. As an account with an EVM address alias set after creation, I would like to set an evm address override to my
+    previous long zero address for a `ContractCall` transaction.
+
```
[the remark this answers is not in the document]
Nana-EC: Adopted change

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+    alias on my account that will identify my account on the EVM.
+3. As an existing account with a complex key but no EVM address alias, I would like to set an ECDSA derived EVM address
+    alias on my account that will identify my account on the EVM.
+5. As an existing account with a newly set ECDSA derived EVM address, I want my account to conform with `ecrecover`
+    expectations as long as I sign the transaction with the ECDSA key that corresponds to the EVM address alias.
+    This key may or may not be the key currently on my account.
+5. As an account with an EVM address alias set after creation, I would like to set an evm address override to my
+    previous long zero address for a `ContractCreate` transaction.
+6. As an account with an EVM address alias set after creation, I would like to set an evm address override to my
+    previous long zero address for a `ContractCall` transaction.
```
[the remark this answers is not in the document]
Nana-EC: Adopted change

HIP/hip-1082.md
```diff
@@ -0,0 +1,225 @@
...
+  
+## Specification
+
+The main change to the network is simple - `CryptoUpdate` transactions will support setting a new ECDSA derived
+`evm_address` if the current value is unset.
```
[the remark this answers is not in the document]
Nana-EC: Adopted change

HIP/hip-1082.md
Nana-EC: Adopted change

HIP/hip-1082.md
Nana-EC: Adopted change

HIP/hip-1082.md
Nana-EC: Adopted change

HIP/hip-1082.md
Nana-EC: Adopted change

HIP/hip-1082.md
Nana-EC: Adopted change

HIP/hip-1082.md
Nana-EC: Adopted change

HIP/hip-1082.md
Nana-EC: Adopted change

HIP/hip-1082.md
Nana-EC: Removed

HIP/hip-1082.md
Nana-EC: Added

Nana-EC:
> I agree with this HIP's proposed solution and believe it is the right way to address many possible issues.
> 
> However, I would like to highlight that the rationale and, therefore, the impact of this change are likely less severe than they appear.
> 
> Firstly, I strongly believe that the `ecrecover` precompile should not be modified and its behaviour should be kept the same as Ethereum. `ecrecover` is meant to be a pure cryptographic function - a stateless black-box one. The rationale mentions that this HIP will enable `ecrecover` to "work". I believe that `ecrecover` works exactly as intended right now.
> 
> The real reasoning behind this HIP, for me, comes from the use-cases of `ecrecover` rather than the inner workings of `ecrecover`. Many Ethereum applications use "signed-message"-based authorization - Based on a signed message by the user, the dapps enable certain actions to be authorized on the behalf of the signing user. The most famous example is the ERC20 add-on - ERC-2612 - commonly known as "permits". In this use-case the owner of certain amount of tokens, signs a message authorizing another user to use them.
> 
> This HIP solution will help enable this use case, but it is **not** because `ecrecover` is somehow broken, but because it will enable our users with non-ecdsa keys to set their ecdsa representing addresses, therefore be eligible to use this authorization scheme.
> 
> In respect of fairness, it is important to highlight the negative side-effects of this change. As @rbair23 pointed, this change, and our aim to support the "signed-message"-based authorization is in fact leading to hard-coupling between user and their ECDSA keys. This disables the much desired key rotation.

Thanks for the perspective and context @Perseverance, good points

> Firstly, I strongly believe that the `ecrecover` precompile should not be modified and its behaviour should be kept the same as Ethereum. `ecrecover` is meant to be a pure cryptographic function - a stateless black-box one. The rationale mentions that this HIP will enable `ecrecover` to "work". I believe that `ecrecover` works exactly as intended right now.

Agreed. `ecrecover` is not being altered in anyway by this HIP. The only change is an accounts ability to provide a conforming evm address to the EVM that when lining up with its ECDSA signed signature key will allow for `ecrecover` operation to produce similar results as expected on other EVM chains.


> The real reasoning behind this HIP, for me, comes from the use-cases of `ecrecover` rather than the inner workings of `ecrecover`. Many Ethereum applications use "signed-message"-based authorization - Based on a signed message by the user, the dapps enable certain actions to be authorized on the behalf of the signing user. The most famous example is the ERC20 add-on - ERC-2612 - commonly known as "permits". In this use-case the owner of certain amount of tokens, signs a message authorizing another user to use them.
> 
> This HIP solution will help enable this use case, but it is **not** because `ecrecover` is somehow broken, but because it will enable our users with non-ecdsa keys to set their ecdsa representing addresses, therefore be eligible to use this authorization scheme.

The reasoning originates from the lack of EVM address equivalence for accounts on Hedera when they do not have an ECDSA derived evm address. `ecrecover` is one of the loudest use cases that highlights the address incompatibility. The other comes in the user bases confusion when they see a non ECDSA derived addresses utilized on the EVM.
Agreed that `ecrecover` itself is not broken but just an accounts ability to have their details used as regular inputs with the expected operation behaviour.
Also agree that that this will enable an accounts ability to utilize this heavily utilized authorization scheme.

Thanks

HIP/hip-1082.md
Nana-EC: The CN should ensure that the override address is adopted fully by the EVM, so there should be no changes to Mirror Node processing on those calls.
The `sender` value in `ContractFunctionResult` should be that of the override utilized in the transaction body.
Do you see any processing changes as a result of this or in other areas?
I'm happy to call them out or address them if they were missed

HIP/hip-1082.md
Nana-EC: If a valid alias is set where one did not exist the new value is indeed used during Contract calls.
The override option is provided to set the older long-zero address in the case that a user would like to be perceived by the EVM as such this to unblock themselves.
This is why the address format is used and not the AccountID format.

For example, say an account with no alias has interacted with a contract that cached its long zero address. If they set the alias with this HIP and again interacted with the contract they would likely see the new address cached.
If for some reason that account wanted to migrate in contract details from the original long zero address to the new ECDSA derived address they could do so by utilizing this override in addition to the contract logic.

HIP/hip-1082.md
Nana-EC: Correct, the `evm_address` for an account will show the new value.
New contract results would also show the override.
Older accounts values and contract results should show the long zero if that's what was in the record file for correctness.
As such there shouldn't be any API field changes.
DO you see otherwise?

Added some details for this

HIP/hip-1082.md
Nana-EC: Fair point.
`sender` might be cleaner as in implementation this is setting the EOA `msg.sender` observed by the EVM

Nana-EC:
> From an evm perspective, this is the most compatible version, and allows the signed message schemes at smart contract level. That said as @rbair23 and @Perseverance mention, this intermediate HIP is a step forward, but we need to keep an eye on HIP -631, to enable alias rotation, or multiple evm adressess alias to identity/account mappings.
> 
> Apologizes in forehand for next comments, if there is anything out of context.
> 
> As a newbie, I want to suggest some extra adds to the context to make the HIP more comprehensive, added to a question
> 
> * Make it clear, that the alias (derived evm address) is unique and cannot be used within more than 1 account
> * Separate the concepts of key rotating and alias. Documentation indicates the following:
> 
> "If an alias is set during account creation, it becomes [immutable](https://docs.hedera.com/hedera/support-and-community/glossary#immutability), meaning it cannot be changed. If you plan to update or rotate keys in the future, do not set the alias at the time of initial account creation. The alias can be set after finalizing all key updates."

Great points @Ferparishuertas. I added some clarifications notes in earlier sections of the HIP to address this.
Thanks

HIP/hip-1082.md
Nana-EC: Updated

HIP/hip-1082.md
Nana-EC: Updated

HIP/hip-1082.md
mgarbs: Is this resolved @ty-swirldslabs @Nana-EC ?

HIP/hip-1082.md
```diff
@@ -0,0 +1,248 @@
...
+However, not all accounts on Hedera use ECDSA keys or were created after HIP-583 was defined. This HIP adds support for
+specifying an alias on an account, so long as that account does not already have a valid alias and the alias to be set
+is not already in use on the network. [HIP-631](https://hips.hedera.com/hip/hip-631) defines a more complex set of
+rules that would allow multiple aliases to be defined for any given account. This HIP differs from HIP-631 in
+that it supports only a single alias. HIP-631 could be implemented compatibly at a later date if desired.
+
+> Note 1: An ECDSA derived address alias is unique across the network. Only 1 account may utilize a given alias as a
+time. Attempts to create an account or update an account to use an already in use alias will be rejected by the network.
+
+> Note 2: Key rotation and aliases are separate concepts. A valid alias when set is immutable and remains for the
```
edward-swirldslabs, line 55: If the alias is immutable once set and the user loses their ECDSA private key, not only are the EVM assets associated with that key lost, but that Hedera account loses its ability to be associated with a different EVM address.  The user would have to create a new Hedera account, associate it with a new EVM address, and transfer all Hedera assets from the old account to the new account. 

I think this needs to be documented as an undesirable side effect of an immutable association and a disadvantage reduced target instead of the full dynamism in 631.

HIP/hip-1082.md
```diff
@@ -0,0 +1,248 @@
...
+aimed to resolve this issue as well as provide even greater functionality. A form of HIP 631 could be revisited in the
+future but not with the goal of providing address equivalence with the EVM.
+
+## Motivation
+
+On Ethereum, every account is addressed by an address which is derived from an ECDSA public key. The
+EVM also has a specific precompile function, `ecrecover`, that given an ECDSA signature can extract the public key
+and convert that public key into an address. This allows an EVM program to figure out the address of
+transaction signers.
+Hiero accounts are not limited to ECDSA keys, and Hiero accounts can rotate their keys. Both of these
```
david-bakin-sl, line 38: This section and the following two need some additional blank lines to introduce paragraphing in the rendered version.

HIP/hip-1082.md
Nana-EC: Additionally, this wouldn't change in an implementation of HIP 631, as the alias is still immutable and the new virtual addresses would again be tied up to a single account at a time. If they wanted to move the address though it would be a CryptoUpdate vs a CryptoDelta + CryptoCreate as is the case of the existing network pre and post this HIP.
HIP 631 could make the UX of moving an alias around easier but it's not difficult today with successive trnasactions.
Assets in an account would still have to be migrated in both cases

mgarbs:
Hey everyone,

I trust you're all doing well. I wanted to take a moment to address the PR titled "Allow Setting of EVM Address by EOA". It has seen quite a lot of fruitful discussion, and I appreciate the time and effort that everyone has put into this.

We currently have 73 unresolved discussions, and I believe it would be beneficial for the successful integration of this feature if we could revisit these and work collaboratively to resolve them.

I would really love to see this pull request to the finish line, and your cooperation would be incredibly valuable to achieve that. I'm immensely thankful for your participation and input.

Looking forward to our continued exchanges. Thanks, folks!
@gregscullard - There's an unresolved discussion from 2024-11-25. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1856920985)

@ty-swirldslabs - There's an unresolved discussion from 2024-11-25. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1857286050)

@ty-swirldslabs - There's an unresolved discussion from 2024-11-25. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1857290141)

@ty-swirldslabs - There's an unresolved discussion from 2024-11-25. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1857294040)

@Nana-EC - There's an unresolved discussion from 2024-12-04. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1870328111)

@Nana-EC - There's an unresolved discussion from 2024-12-05. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1871683556)

@Nana-EC - There's an unresolved discussion from 2024-12-07. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1874239500)

@hendrikebbers - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1880109594)

@hendrikebbers - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1880110121)

@hendrikebbers - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1880116514)

@IvanKavaldzhiev - There's an unresolved discussion from 2024-12-11. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1880248819)

@Nana-EC - There's an unresolved discussion from 2024-12-16. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887352802)

@Nana-EC - There's an unresolved discussion from 2024-12-16. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887376974)

@Nana-EC - There's an unresolved discussion from 2024-12-16. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887377013)

@Nana-EC - There's an unresolved discussion from 2024-12-16. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887377068)

@Nana-EC - There's an unresolved discussion from 2024-12-16. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887377141)

@Nana-EC - There's an unresolved discussion from 2024-12-16. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887377509)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887902519)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887910059)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887916220)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887930294)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887933398)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887936357)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887938511)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887940098)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887942726)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887945863)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887947528)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887948070)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887949600)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887950086)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887954131)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887954647)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887957616)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887958553)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887958806)

@rbair23 - There's an unresolved discussion from 2024-12-17. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1887958999)

@Nana-EC - There's an unresolved discussion from 2024-12-18. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1889516499)

@steven-sheehy - There's an unresolved discussion from 2025-01-03. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1902212726)

@steven-sheehy - There's an unresolved discussion from 2025-01-03. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1902213176)

@steven-sheehy - There's an unresolved discussion from 2025-01-03. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1902220483)

@steven-sheehy - There's an unresolved discussion from 2025-01-03. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1902220933)

@steven-sheehy - There's an unresolved discussion from 2025-01-03. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1902223301)

@steven-sheehy - There's an unresolved discussion from 2025-01-03. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1902224158)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923163849)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923163996)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923164056)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923164115)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923164329)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923164379)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923164430)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923164568)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923164658)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923164755)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923164854)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923164903)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923164966)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923165020)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923165072)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923165175)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923165253)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923213898)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923221837)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923239976)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923240739)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923256344)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923257194)

@Nana-EC - There's an unresolved discussion from 2025-01-21. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1923257240)

@mgarbs - There's an unresolved discussion from 2025-01-23. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1926142995)

@edward-swirldslabs - There's an unresolved discussion from 2025-01-28. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1932775896)

@david-bakin-sl - There's an unresolved discussion from 2025-01-29. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1934353399)

@Nana-EC - There's an unresolved discussion from 2025-02-03. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1938808062)

@Nana-EC - There's an unresolved discussion from 2025-02-03. Could you please take a look and either resolve or provide additional context? [Link to discussion](https://github.com/hashgraph/hedera-improvement-proposal/pull/1082#discussion_r1938808158)

HIP/hip-1082.md
edward-swirldslabs: I am fine with this HIP going through as is, but think there should be a new HIP to extend capability to allow for alias rotation: 
1. The 0x0000000### alias can be rotated to active alias without additional signing. 
2. The active alias can be rotated to any ECDSA derived alias with signing from that key and verification that the alias is not used by any other account. 

This way there are no inaccessible assets caused by permanent alias associations. 

UX problems should be solved at the level of UX, not here.  Tracking possible keys whose alias can be activated and tracking asset associations to ECDSA keys so the user doesn't throw away a key with active assets associated is a UX responsibility.

edward-swirldslabs: This conversation is resolvable from my perspective.

PR #1084, opened 2024-11-20
Zero Cost EthereumTransaction on Success

Nana-EC:
**Description**:
On many EVM chains execution costs are fully captured in gas. Thus submitting transactions to execution nodes bears no
cost to the relay node. On Hedera it is necessary to charge a fee for `EthereumTransaction` HAPI submissions (just like
all other HAPI transactions).

This HIP suggests the network logic not charge successful `EthereumTransaction` submissions, but instead only charge the
regular HAPI fee for badly formed transactions.

**Related issue(s)**:

Fixes #1083 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-1084.md
+++ b/HIP/hip-1084.md
@@ -10,7 +10,7 @@ needs-council-approval: Yes
 status: Draft
 created: 2024-11-20
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/1083
-updated: 2024-12-06
+updated: 2024-13-06
 requires: 410
 ---
 
@@ -32,14 +32,15 @@ network.
 
 ## Rationale
 
-On Ethereum, all execution costs are paid for by the 'sender' who created, signed, and submitted the transaction. Hedera
-follows a similar approach for all `ContractCallTransaction`s but also charges an additional fee for each
-`EthereumTransaction` submitted via JSON-RPC relays. This fee serves as a defense against a malicious JSON-RPC relay that
-might send malformed transactions to the network, where there would be no one to pay for the processing of those transactions.
-This fee is unnecessary for successful transactions since the intrinsic gas fee covers the base costs (gossip & consensus).
-Therefore, this HIP proposes that the fee for submitting `EthereumTransactions` to the network should only be charged if the
-JSON-RPC relay submits a transaction that fails due diligence checks, such that the EOA cannot be charged for the overhead
-of gossip and consensus.
+On Ethereum, all execution costs are paid for by the 'sender' who created, signed, and submitted the transaction.
+Hedera follows a similar approach for `ContractCallTransaction`s. However, Hedera charges an additional fee for
+each `EthereumTransaction` submitted via JSON-RPC relays. This fee serves as a defense against a malicious
+JSON-RPC relay that might send malformed transactions to the network, where there would be no one to pay for
+the processing of those transactions.
+This fee is unnecessary for successful transactions since the intrinsic gas fee covers the base costs (gossip &
+consensus). Therefore, this HIP proposes that the fee for submitting `EthereumTransactions` to the network
+should only be charged if the JSON-RPC relay submits a transaction that fails due diligence checks, such that the
+EOA cannot be charged for the overhead of gossip and consensus.
 
 ## User stories
 
--- a/HIP/hip-1084.md
+++ b/HIP/hip-1084.md
@@ -10,7 +10,7 @@ needs-council-approval: Yes
 status: Draft
 created: 2024-11-20
 discussions-to: https://github.com/hashgraph/hedera-improvement-proposal/discussions/1083
-updated: 2024-13-06
+updated: 2024-12-17
 requires: 410
 ---
 
--- a/HIP/hip-1084.md
+++ b/HIP/hip-1084.md
@@ -16,7 +16,7 @@ requires: 410
 
 ## Abstract
 
-Updates the rules for charging fees for EthereumTransactions, so that the payer of an EthereumTransaction will only
+Updates the rules for charging fees for EthereumTransactions, so that the payer of an `EthereumTransaction` will only
 incur a fee if the transaction fails due diligence checks.
 
 ## Motivation
@@ -34,7 +34,7 @@ network.
 
 On Ethereum, all execution costs are paid for by the 'sender' who created, signed, and submitted the transaction. Hedera
 follows a similar approach for all `ContractCallTransaction`s but also charges an additional fee for each
-`EthereumTransaction` submitted via JSON-RPC relays. This fee serves as a defense against a malicious JSON-RPC relay tha
+`EthereumTransaction` submitted via JSON-RPC relays. This fee serves as a defense against a malicious JSON-RPC relay that
 might send malformed transactions to the network, where there would be no one to pay for the processing of those transactions.
 This fee is unnecessary for successful transactions since the intrinsic gas fee covers the base costs (gossip & consensus).
 Therefore, this HIP proposes that the fee for submitting `EthereumTransactions` to the network should only be charged if the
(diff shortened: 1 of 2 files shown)


rbair23:
@Neurone I agree, the text has been updated, but I think it would benefit from being even more precise.

HIP/hip-1084.md
```diff
@@ -0,0 +1,111 @@
...
+The `relay` will **only** be charged if handling of the `EthereumTransaction` fails **before** any bytecodes are executed.
+In other words, the `relay` is charged when one of the pre-requisite checks for `EthereumTransaction` fails, regardless of
+the execution status of the transaction itself once it begins smart contract creation or execution in the EVM.
+
+Examples:
+ - The `relay` is charged if the `evm tx` cannot be parsed or is otherwise malformed
+ - The `relay` is charged if the `EOA` specified in the `evm tx` refers to a non-existent account
+ - The `relay` is charged if the `evm tx` was not properly signed by the keys of the `EOA`
+ - The `relay` is charged if the `EOA` has insufficient funds to pay the intrinsic gas fee
+ - The `relay` is charged if the `nonce` check fails
```
quiet-node, line 73: This indeed highlights a list of prechecks performed by the Relay to validate transactions and reject them if any conditions are met, ensuring they are not forwarded to the network unnecessarily, thus avoiding charges. Notably, the Hedera JSON-RPC Relay already handles and rejects these scenarios. However, there are additional prechecks currently performed by the Relay that are not mentioned here:

- Transaction type  
- Gas price  
- Gas limit  
- Chain ID  
- Value  

I recommend considering these prechecks and including them in this HIP for greater clarity and completeness.

Nana-EC:
@Neurone added an explicit clarification of response codes. 
For not it's probably better to not the cases for which no charge to the relay account is made.
Listing the cases where it will may be lengthy and that list may grow over time, but the non charged cases are less likely to grow

Neurone:
Hi Nana, good. Listing the cases for which the operator is not charged is perfectly fine and clear, thanks.

HIP/hip-1084.md
```diff
@@ -0,0 +1,118 @@
...
+ - The `relay` is not charged if the contract is called but concludes with an error code 
+ - The `relay` is not charged if the contract is called and concludes without error
+
+In summary the `relay` account is not charged if the `EthereumTransaction` transaction completes in a 
+`CONTRACT_REVERT_EXECUTED` or `SUCCESS` HAPI response code, all other responses codes will result in a charge to the
+relay account.
```
IvanKavaldzhiev, line 79: If such a pre-check fails on consensus node side, how it will be made sure the relay has enough balance to be charged for the submission of a faulty transaction?

HIP/hip-1084.md
```diff
@@ -0,0 +1,118 @@
...
+- The `relay` is charged if the receiver doesn't exist, must be auto-created, and the `EOA` has insufficient funds to
+    pay for the account creation
+- The `relay` is charged if the `nonce` check fails
+
+In summary the `relay` account is not charged if the `EthereumTransaction` transaction completes in a `SUCCESS`
+HAPI response code, all other response codes will result in a charge to the `relay` account.
+
+## Backwards Compatibility
+
+No changes to transaction execution or record stream externalization is made in this HIP.
```
IvanKavaldzhiev, line 102: As far as I understood the idea, if the transaction is correct the relay won't be the payer of the transaction anymore. This would mean the signer (the `from` fieldn) and the payer will be the same account. Even if the record stream or proto structures are not changed we should make sure that eventually, the `signer=payer` part will be properly reflected in the implementation and the externalisation of the record.

Neurone: Hi @IvanKavaldzhiev I suggest not mixing the signer (from) and the payer concepts, regardless who is charged for the fees. The wrapped ethereum tx remains signed by the sender, and the "envelop" transaction still contains the payer's signature. Both accounts can be charged based on the outcome of the tx. Even if the payer is charged 0, it is still a potential payer, it set a maximum amount of hbar it allows to be spent, etc. This is the same concept now where, in case of some errors, only the payer is charged and not the sender (i.e. WRONG_NONCE: https://hashscan.io/testnet/transaction/1730913223.447111312)

IvanKavaldzhiev: Your idea is keeping the payer concept to reflect the relayer account even if we don't charge it in case of successful transaction? You have a point that if signer=payer, then we will miss track of which relayer account has sent the transaction to the consensus node.

HIP/hip-1084.md
Nana-EC: This is standard logic in todays network where precheck varies a HAPI submitters balance to process a transaction before gossiping out.
So there's no change here

HIP/hip-1084.md
Nana-EC: @Neurone is correct.
No change to that logic, separate concepts.

HAPI sender/payer and the internal tx sender are by definition always different in the relay case when using EVM tools.
The only time they'd be the same is if the EOA was running their own relay using their account as the relay operator and they submitted an Eth TX to themselves.m or if they used the SDK directly to submit an `EthereumTransaction` in the same manner as the relay

IvanKavaldzhiev: I see, I agree with your point. It was an important aspect to clear out in this discussion.

HIP/hip-1084.md
```diff
@@ -0,0 +1,118 @@
...
+## Rationale
+
+On Ethereum, all execution costs are paid for by the 'sender' who created, signed, and submitted the transaction. Hedera
+follows a similar approach for all `ContractCallTransaction`s but also charges an additional fee for each
+`EthereumTransaction` submitted via JSON-RPC relays. This fee serves as a defense against a malicious JSON-RPC relay that
+might send malformed transactions to the network, where there would be no one to pay for the processing of those transactions.
+This fee is unnecessary for successful transactions since the intrinsic gas fee covers the base costs (gossip & consensus).
+Therefore, this HIP proposes that the fee for submitting `EthereumTransactions` to the network should only be charged if the
+JSON-RPC relay submits a transaction that fails due diligence checks, such that the EOA cannot be charged for the overhead
+of gossip and consensus.
```
rbair23, line 42:
```suggestion
On Ethereum, all execution costs are paid for by the 'sender' who created, signed, and submitted the transaction.
Hedera follows a similar approach for `ContractCallTransaction`s. However, Hedera charges an additional fee for
each `EthereumTransaction` submitted via JSON-RPC relays. This fee serves as a defense against a malicious
JSON-RPC relay that might send malformed transactions to the network, where there would be no one to pay for
the processing of those transactions.

This fee is unnecessary for successful transactions since the intrinsic gas fee covers the base costs (gossip &
consensus). Therefore, this HIP proposes that the fee for submitting `EthereumTransactions` to the network
should only be charged if the JSON-RPC relay submits a transaction that fails due diligence checks, such that the
EOA cannot be charged for the overhead of gossip and consensus.
```

HIP/hip-1084.md
rbair23: I agree, thank you @quiet-node! Can you please reply to this comment with a suggested replacement (you can click the icon of a document with the + and - signs within it to offer a suggestion) and we will get it applied to the HIP text.

stoqnkpL: If a precheck is performed by the Relay, doesn't this mean that if it fails, the Relay will never send a transaction to a Hedera node, meaning the relay account can never be charged for it?

HIP/hip-1084.md
Nana-EC: Changes applied

HIP/hip-1084.md
Nana-EC: Correct. 
A relay will only submit to a CN a transaction that passes it's prechecks.
If it fails it is in the best interest of the relay to not submit to a CN otherwise it will be charged.

Nana-EC:
Heiro TSC approved. Waiting on Hedera considerations to determine when this might be pushed to Hedera mainnet

mgarbs (approved):

mgarbs (approved):

PR #1086, opened 2024-11-20
Jumbo EthereumTransactions

Nana-EC:
**Description**:
Hedera enforces a 6kb transaction limit which requires additional network complexity to support large callData
scenarios. This often means only about ~5kb is left for the transaction callData (1kb is used by signatures)
However, EVM type transactions on other EVM networks support callData sizes of up to 128kb.
This is partly because the cost of work is fully captured by `gas * gasPrice` on the network.

This HIP proposes to support `EthereumTransaction` submissions with `callData > 5KB` for increased EVM
equivalence.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


Neurone:
Should we consider increasing the fees for malformed `EthereumTransaction` accordingly with their size, i.e., from $0.0001 for 5kb up to $0.0025 for 128kb?

rbair23:
@Neurone, I agree the EthereumTransaction fee will go up based on the number of bytes in a non-linear way. That is, 12KB should cost more than 2x the price of 6KB.

Nana-EC:
> Should we consider increasing the fees for malformed `EthereumTransaction` accordingly with their size, i.e., from $0.0001 for 5kb up to $0.0025 for 128kb?

Yep, my next update will include a suggestion for a tiered progressive pricing. Great minds ....

rbair23:
@Nana-EC I think we should not include a suggestion for tiered progressive pricing. We should leave this to CoinCom. I avoided trying to get too specific because Coincom has to make that call.

HIP/hip-1086.md
```diff
@@ -0,0 +1,164 @@
...
+- Overall deployment / call latency is increased due to the need to upload files and then reference them in the
+  Ethereum transactions.
+
+In addition, JSON-RPC relays need to find a business model whereby they can charge users for these HFS transactions.
+This creates a poor developer experience. This again breaks the business model of many cross chain relay operators and
+makes access to cheap relays difficult.
+
+## User stories
+
+1. As an `EOA` submitting a transaction with large call data I expect the network to accept my transaction and charge me
```
vtronkov, line 62: This HIP refers only to the Ethereum `transactions`, right? Do you think we should specify this here?

HIP/hip-1086.md
```diff
@@ -0,0 +1,164 @@
...
+system). If any transaction received exceeds 6KB in size, then it is classified as a "jumbo" transaction. In 
+addition to the normal throttle limits, the number of bytes exceeding 6KB will be added to the 'jumbo' throttle
+bucket. If the bucket does not have capacity, then the transaction will be rejected.
+
+In addition, the price of the transaction will be increased non-linearly for transactions that exceed 6KB in size. The
+actual pricing will be determined by the Hedera Council and specified as part of the network configuration. The price
+will **not** affect the gas, but **will** affect the exchange rate between gas and USD.
+
+A JSON-RPC relay *should* calculate the HBAR cost of the transaction and verify the sender has sufficient funds before
+sending the transaction to the network to avoid due diligence failure fees.
```
quiet-node, line 87: So correct me if I'm wrong but this could basically be broken down as:  

- **Non-jumbo transactions** (≤ 6 KB) will execute normally, with the only cost to the sender (EOA) being gas fees. With honor of HIP-1084, the Relay operator incurs no additional cost, provided the `EthereumTransaction` is well-formed and does not return any response codes other than `CONTRACT_REVERT_EXECUTED` or `SUCCESS`. This is cool and makes sense!

- **Jumbo transactions** (6 KB ≤ tx ≤ 128 KB) also do not impose costs on the Relay operator. However, the sender (EOA) must cover the increased cost, which is separate from gas fees i.e. in addition to gas, an **extra fee** applies specifically to jumbo transactions.  

If that's correct, then the fee discrepancy between Ethereum and Hedera in this sense is still there, right? But instead of the cost being charged to the Relayer operator via HFS, it is now applied directly to the senders? 

Pre-HIP-1086, and also on the Ethereum network, EOAs only paid for gas. Post-HIP-1086, they now have to pay an extra fee for jumbo transactions.

Previously, the fee was tied to HFS transactions and charged to the Relay operator, which was reasonable since operating a Relay requires considering network architecture and accounting for costs in the business model. This trade-off made sense, as Relay operators aimed to maintain Ethereum compatibility.

However, now that this fee is shifted to EOAs, those migrating from Ethereum—expecting compatibility without looking to deep into the network architecture—may find it unexpected that submitting an `eth_sendRawTransaction` incurs an additional fee beyond gas. Could this become a friction point for users?

Nana-EC: Hey @quiet-node good question.
The first part is correct but not the second.
The cost should be a part of the gas costs paid by the EOA in a predictable manner.
So this would be reflected in `eth_estimateGas` calls and not a surprise to a user.

However, this HIP doesn't detail that part exactly as some thought needs to be given to it by a network itself.

These exact details are intentionally left out of the HIP.

I agree with your concerns though. We'd want it to be predictable and not a shock.

Note,there continues to be a responsibility or a best practice for developers and users to check the gas costs before submitting a transaction. Most tools do this for you so even if it is additional gas users will have a choice to submit or not.
The experience at worst should be similar to the auto creation flow when sending value to an account that doesn't yet exist.

quiet-node: Ah, I see. That makes sense. If the increased cost is included in the gas, then it’s expected. I believe the wording got me confused a little the “The price will not affect the gas” made me think the cost would act like a network fee like how the HFS fees were rather than a regular EVM gas value.

HIP/hip-1086.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Relay
+
+The JSON-RPC relay will be updated to use jumbo transactions. It will no longer use files for large callData.
+
+## Backwards Compatibility
+
+Behavior for the consensus node is fully backwards compatible. The default behavior for SDKs will change to use
+jumbo transactions, which may surprise users who are not expecting the change, and the overall cost may be higher than
+if the user had used HFS. However, the user can still use HFS if they prefer.
```
MiroslavGatsanoga, line 133: I suppose transactions above the new 128KB limit should be handled using HFS as previously?
Will the SDKs support that seamlessly or we'll let users handle that on their own?

edward-swirldslabs: The last line of the SDK section indicates the SDKs would still support the HFS medium for Eth transactions. 

> The SDK should support an option to use HFS for callData if the user prefers 
to manage the complexity themselves.

HIP/hip-1086.md
Nana-EC: I see.
I guess it's still up for debate. Preferably the gas amount stays the same and Hedera uses gas price to factor things in.

steven-sheehy (approved):

edward-swirldslabs (approved):

mgarbs (dismissed):

lukelee-sl:
This HIP does not seem to address the size of the runtime bytecode being limited to 6k for contract deployment.  More recent versions of Besu allow for setting this limit as a configuration value in the `EvmConfiguration` class.  Should this limit be set to a value greater than 6k?

Neurone:
> Should this limit be set to a value greater than 6k?

Hi @lukelee-sl I was pretty sure smart contract's size on Hedera were capped to 24 Kb as on Ethereum, I remember doing tests about this without encountering issues. Isn't the case anymore?

HIP/hip-1086.md
```diff
@@ -0,0 +1,181 @@
...
+the use of HFS for Ethereum data will be lessened. The Mirror Node should be able to handle both jumbo
+Ethereum transactions and legacy HFS-based Ethereum transactions.
+Mirror node will need to support the gas estimation of the "new" Jumbo Ethereumtransaction.
+
+### Relay
+
+The JSON-RPC relay will be updated to use jumbo transactions. It will no longer use files for large callData.
+
+A JSON-RPC relay *should* calculate the HBAR cost of the transaction and verify the sender has sufficient funds before
+sending the transaction to the network to avoid due diligence failure fees.
```
quiet-node, line 120: Since the `HBAR cost of the transaction` mentioned here appears to be associated with the sender, is it accurate to assume that it essentially refers to the total of the gas fees and the transferred value i.e. tx.value + tx.gasPrice * tx.gasLimit? It shouldn't be the HAPI fee for the the EthereumTransaction itself, right?

Neurone: Correct. The rest of the fees are already paid by the relay operator.

Side note: failures are always technically possible, because the user's balance can still change between the time the check is made and the time the tx reaches consensus on the network.

mgarbs: I am marking this as resolved

mgarbs:
> > Should this limit be set to a value greater than 6k?
> 
> Hi @lukelee-sl I was pretty sure smart contract's size on Hedera were capped to 24 Kb as on Ethereum, I remember doing tests about this without encountering issues. Isn't the case anymore?

This concern is resolved, yea?

mgarbs (approved):

PR #1095, opened 2025-01-08
HIP-991: Introduce generic custom fee limits

netopyr:
**Description**:

Moved `max_custom_fee` to `TransactionBody`
Removed `accept_all_custom_fees` flag
Some minor improvements


HIP/hip-991.md
```diff
@@ -203,35 +204,43 @@ message ConsensusUpdateTopicTransactionBody {
...
+message TransactionBody {
   [..]
   /**
-    * The maximum custom fee that the user is willing to pay for the message. This field will be ignored if `accept_all_custom_fees` is set to `true`.
+    * The maximum custom fee that the user is willing to pay for the message. If left empty, the user is willing to pay any custom fee.
+    * If used with a transaction type that does not support custom fee limits, the transaction will fail.
     */
-  repeated FixedFee max_custom_fees = 4;
-  
+  repeated CustomFeeLimit max_custom_fees = 1001;
```
ty-swirldslabs, line 231: This is the only place I see FixedFee being updated to CustomFeeLimit, and the value changing from 4 to 1001. Only making a comment as I'm unsure if intentional.

ty-swirldslabs (approved):
lgtm, I added one comment to confirm that change was wanted, but everything else looks correct

HIP/hip-991.md
netopyr: The number is part of the protobuf definition and probably irrelevant in the HIP. We use 7-65 for the different transaction types. To simplify the definition, I thought we should leave the following indexes for other transactions that we are going to add. But really, it does not matter as long as the index is unique.

steven-sheehy, line 231: Are we sure we really want to add this field to the generic `TransactionBody` instead of a common structure added to each impacted individual transaction bodies? This is the first time we're adding a new generic field to this message in 5+ years. This is also the opposite approach we settled on for block stream when discussing a field that applied to only two transaction types to not put it in the generic `TransactionResult` and to put it only in the specific transaction output.

steven-sheehy (changes requested):

HIP/hip-991.md
```diff
@@ -132,20 +132,21 @@ The HIP parameters are defined as follow:
 ### User Flows and Interaction
 
 * Users will specify the fee settings during the topic creation process through a simple interface in their Hedera client (refer to the creation of token custom fees/fixed fee for reference).
-* When submitting a message to a topic with custom fees through an application or wallet interface, users must set the maximum fee for the message. Alternatively, users can add a flag to accept all custom fees from a topic id. If the topic has no custom fees, neither is required.
+* When submitting a message to a topic with custom fees through an application or wallet interface, users can set the maximum fee for the message. If `max_custom_fee` is not set, any custom fee will be accepted by the user (up to the payer's available funds).
 * In case of user wallets, applications show the custom fees to the user before submitting the message.
 * Operators will get fee collections and distributions automatically through the custom fees just like they do in the token service currently.
+* If the specified custom fee limit or the payer's funds are insufficient to cover the custom fees, the transaction will fail.
```
xin-hedera, line 138: should we also highlight that when max custom fee is specified and any custom fee the topic intends to charge without a matching max custom fee will also fail the transaction?

HIP/hip-991.md
rbair23: We had decided in the meeting that this would go onto `TransactionBody` (the high level transaction object, not the transaction body data)

steven-sheehy (dismissed):

HIP/hip-991.md
```diff
@@ -203,35 +204,43 @@ message ConsensusUpdateTopicTransactionBody {
...
+  repeated CustomFeeLimit max_custom_fees = 1001;
+}
+
+message CustomFeeLimit {
   /**
-   * If set to true, the transaction will accept all custom fees from the topic id
+   * The account of the fee payer
    */
-  bool accept_all_custom_fees = 5;
+  AccountID account_id = 1;
```
xin-hedera, line 238: is this supposed to be the fee collector? and if so, for consistency as named in proto message `CustomFee`, should we also name it `fee_collector_account_id`?

Neeharika-Sompalli: +1

HIP/hip-991.md
netopyr: Yes, makes sense.

HIP/hip-991.md
netopyr: No, this is not for the fee collector. This field defines the account for which the custom fees are applied. This could be the payer of the transaction or any account that appears in the `TransferList` (in the case of `CryptoTransfer`).

I'd be happy to use a more specific name, but I cannot think of anything. Using `payer` in any form would make it too easy to be confused with the payer of the transaction.

xin-hedera (dismissed):
LGTM

Neeharika-Sompalli (approved):
LGTM! Thanks @netopyr

HIP/hip-991.md
```diff
@@ -318,6 +328,257 @@ To comply with this HIP, the Mirror Node provides the following features:
...
+              {
+                "amount": 1000,
+                "denominating_token_id": null
+              },
+              {
+                "amount": 500,
+                "denominating_token_id": "0.0.2000"
+              }
+            ],
+            "payer_account_id": "0.0.8"
```
steven-sheehy, line 354: This should be `account_id` to match the `CustomFeeLimit` proto.

```suggestion
            "account_id": "0.0.8"
```

steven-sheehy, line 354: At the cost of diverging slightly from the protobuf, we'd prefer to duplicate the account id into the max fees in the JSON response to make the structure flatter and simpler.

```suggestion
          {
            "account_id": "0.0.8"
            "amount": 1000,
            "denominating_token_id": null
          },
          {
            "account_id": "0.0.8"
            "amount": 500,
            "denominating_token_id": "0.0.2000"
          }
```

steven-sheehy (approved):

mgarbs:
Looks good all conversations are resolved!

mgarbs (approved):

PR #1126, opened 2025-02-17
Add HIP to allow the deletion of allowances without being associated

netopyr:
**Description**:

This PR adds the first draft of a HIP to allow the deletion of token allowances for tokens that are not associated with the account anymore. This was requested by the community in the form of a [GitHub issue](https://github.com/hiero-ledger/hiero-consensus-node/issues/17441).

### DIFF
--- a/HIP/hip-1126.md
+++ b/HIP/hip-1126.md
@@ -1,5 +1,5 @@
 ---
-hip: <HIP number (assigned by the HIP editor), usually the PR number>
+hip: 1126
 title: "Delete Allowance without Token Association"
 author: Michael Heinrichs <@netopyr>
 working-group:
--- a/HIP/hip-1126.md
+++ b/HIP/hip-1126.md
@@ -10,7 +10,7 @@ needs-council-approval: "Yes"
 status: "Draft"
 created: 2025-02-17
 discussions-to: https://github.com/hiero-ledger/hiero-consensus-node/issues/17441
-updated: 2025-02-17
+updated: 2025-02-24
 requires: 
 replaces: 
 superseded-by: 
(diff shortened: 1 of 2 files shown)


HIP/hip-1126.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Motivation
+If a user wants to remove an allowance for a token without an association, one has to associate the token, remove the allowance, and dissociate the token again.
+This is a cumbersome and expensive process with no benefits.
+A user may decide to keep the allowances instead, which wastes resources and risks hitting the limit of allowed allowances.
+
+While it makes sense to associate the token before giving a non-zero allowance, it does not make sense that the token must be associated in order for the allowance to be removed.
+
+## Rationale
+The execution of `CryptoDeleteAllowance` does not require the owner to have an association with the token.
```
MiroslavGatsanoga, line 32: Perhaps we can elaborate that in order to remove allowance we use `CryptoDeleteAllowance` for NFTs only and `CryptoApproveAllowance` with an amount of `0` for fungible tokens, because initially I thought we should use `CryptoDeleteAllowance` for both.

HIP/hip-1126.md
```diff
@@ -0,0 +1,52 @@
...
+If `CryptoApproveAllowance` contains an entry for a fungible token and an amount of `0`, the owner does not require an association with the token.
+
+## User stories
+1. As a user, I want to remove an allowance without having to associate the token with my account first.
+
+## Specification
+This change requires no changes to the HAPI or the record stream/block stream.
+
+## Backwards Compatibility
+This HIP is fully backward compatible. It does not change the behavior of any existing functionality except for allowing the deletion of allowances in previously forbidden cases.
```
MiroslavGatsanoga, line 42: A not-so-useful case that will change is when `owner` is not associated to token and has no token amount allowed.
Pre-HIP `CryptoApproveAllowance` behavior will be to return `TOKEN_NOT_ASSOCIATED_TO_ACCOUNT` while post-HIP it will succeed making the transaction a no-op instead i.e. `owner` will still remain not associated and with no token allowance.

vtronkov (approved):
Other than Miro's comments, LGTM

HIP/hip-1126.md
netopyr: Added clarification

HIP/hip-1126.md
netopyr: Yes, the behavior will change if the owner is not associated with the token.

mgarbs: @MiroslavGatsanoga is this okay to be resolved?

Neeharika-Sompalli (approved):
LGTM! Thanks @netopyr

HIP/hip-1126.md
MiroslavGatsanoga: Yes, this can be resolved now. (I am not able to do it though)

MiroslavGatsanoga (approved):

mgarbs (approved):

mgarbs (dismissed):

mgarbs (approved):

PR #1128, opened 2025-02-18
Update hip 1064

Neeharika-Sompalli:
Add updates to HIP-1064 as suggested


netopyr (approved):
LGTM - thanks @Neeharika-Sompalli

mgarbs (approved):

PR #1140, opened 2025-03-11
feat: add Hiero approval filter and UI components

EMerchant90:
- Add needs-hiero-approval field to HIP template
- Add Hiero approval radio button filter to UI
- Add Hiero approval column to HIP tables
- Extend filter.js to handle Hiero approval filtering


mgarbs (approved):
LGTM gj @EMerchant90 !

PR #1141, opened 2025-03-13
Draft hip-0000-improve-hedera-sharding

jxcwu:
Submit HIP draft, Improve Hedera Sharding via Dynamic Committee Reconfiguration


jxcwu:
Thanks for the update! I'll close this for now and look forward to contributing when the HIP is presented.

PR #1146, opened 2025-03-19
Block streams - update block timestamp definition

edward-swirldslabs:
**Description**:
This PR modifies the block stream HIP to clarify the definition of the block timestamp as the consensus timestamp on the first round in the block.

### DIFF
--- a/HIP/hip-1056.md
+++ b/HIP/hip-1056.md
@@ -513,12 +513,9 @@ message BlockHeader {
     bytes previous_block_hash = 4;
 
     /**
-     * A consensus timestamp for the start of this block.
-     * <p>
-     * This SHALL be the timestamp assigned by the hashgraph consensus
-     * algorithm to the first transaction of this block.
+     * A timestamp for this block.
      */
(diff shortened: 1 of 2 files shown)


Nana-EC (approved):

steven-sheehy (approved):

derektriley (commented):
Internal to the consensus node, at the moment the block time is the consensus time of the first event within the first round in the block. See https://github.com/hiero-ledger/hiero-consensus-node/blob/main/hedera-node/hedera-app/src/main/java/com/hedera/node/app/blocks/impl/BlockStreamManagerImpl.java#L232

Should we update the consensus node to use the consensus time of the particular round to align with this update?

edward-swirldslabs:
> Should we update the consensus node to use the consensus time of the particular round to align with this update?"

My understanding is that there are a host of protobuf changes that execution will need to adapt to so that the code follows what the HIP specifies. 

10 people have different conversations with each other at different times, it's really hard to keep everyone in sync and saying the same thing.   

This change that I am pushing now is a reflection of the last known conversation that I was part of.

PR #1151, opened 2025-03-23
NEW HIP: Improve Hedera Sharding

jxcwu:
NEW HIP: Improve Hedera Sharding via Dynamic Committee Reconfiguration

### DIFF
--- a/hip-0000-improve-hedera-sharding.md
+++ b/hip-0000-improve-hedera-sharding.md
@@ -1,0 +1,160 @@
+---
+hip: <HIP number (assigned by the HIP editor), usually the PR number>
+title: "Improve Hedera Sharding via Dynamic Committee Reconfiguration"
+author: Cong Wu, <cnacwu@gmail.com>
+working-group: <List of the technical and business stakeholders' names and/or usernames, or names and emails. Ex: John Doe <@johnDoeGithub1778>, Jane Smith <jane@email.com>>
+requested-by: <Name(s) and/or username(s), or name(s) and email(s) of the individual(s) or project(s) requesting the HIP. Ex: Acme Corp <request@acmecorp.com>>
+type: Standards Track
+category: <"Core" | "Service" | "Mirror" | "Application">
+needs-council-approval: <"Yes" | "No">
+status: Draft
+created: 2025-03-13
+discussions-to: <A URL pointing to the official discussion thread. Ex: https://github.com/hashgraph/hedera-improvement-proposal/discussions/000>
+updated: <Latest date HIP was updated, in YYYY-MM-DD format.>
+requires: <HIP number(s) this HIP depends on, if applicable. Ex: 101, 102>
+replaces: <HIP number(s) this HIP replaces, if applicable. Ex: 99>
+superseded-by: <HIP number(s) that supersede this HIP, if applicable. Ex: 104>
+---
+
+## Abstract
+
+This HIP proposes the integration of a sharding mechanism into Hedera. By partitioning the overall network into Local Committees and a coordinating Global Committee, the design aims to reduce per-node storage and communication overhead while preserving security and Byzantine Fault Tolerance. The proposal details protocols for inter-shard synchronization, efficient cross-shard transaction processing, dynamic committee reconfiguration, and the incorporation of shard state proofs. In addition, the design leverages Hedera’s mirror network to offload full historical data storage, thereby enabling Hedera to scale effectively without sacrificing performance.
+
+------
+
+## Motivation
+
+Hedera’s underlying Gossip-about-Gossip protocol and DAG-based consensus deliver excellent performance, fairness, and low latency. However, with network expansion, several challenges emerge:
+
+- **Scalability Limitations:** Every consensus node must store the ledger’s most recent snapshot (lasting a few minutes) for consensus, while mirror nodes maintain the complete transaction history. This leads to increased resource demands as the network grows.
+- **Communication Overhead:** The extensive Gossip protocol can cause significant per-node bandwidth usage as events proliferate.
+- **Security Concerns:** Static committee assignments can make targeted attacks more effective. Adversaries might exploit this by focusing on a specific shard.
+
+The introduction of sharding with dynamic committee reconfiguration and shard state proofs partitions the workload and data across multiple committees, thereby creating a scalable ecosystem that supports enterprise-level applications while leveraging mirror nodes to maintain full historical data retention.
+
+------
+
+## Rationale
+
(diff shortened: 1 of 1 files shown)


hip-0000-improve-hedera-sharding.md
```diff
@@ -0,0 +1,173 @@
...
+- DAG is naturally partitioned by time and committee ownership.
+- State changes are recorded as state deltas in the Block Stream; these can be replayed or queried directly through Block Nodes.
+
+### Cross-Shard Communication
+
+1. Intra-Shard Phase: 
+   Local Committees process transactions and produce round-based Block Streams.
+
+2. Global Coordination Phase:
+   Cross-shard transactions are handled through the Global Committee using deterministic atomic commit patterns (e.g., 2PC). Round-level consensus is captured via threshold-signed global Block Streams.
```
edward-swirldslabs, line 108: atomic commit patterns require that transactions be able to enter a state where they can be committed to or rolled back.    If all transactions can be committed to, they all execute successfully.   If one or more transactions cannot be committed to, they are all rolled back.    

I suspect this is going to require a transaction by transaction altering of semantics to define an intermediate execution state of a form similar to `(escrowed state, commit synthetic transaction, rollback synthetic transaction)`.  Based on the condition of all commit or all rollback, the respective stored synthetic transaction would execute.  

Making native crypto transfers have that form may be straight forward, escrowing the assets in a ledger escrow accounts pending the commit or rollback, but other transactions are going to be harder to put in that form, such as arbitrary smart contract execution. 

What are the proposed approaches to doing atomic commit patterns with smart contract executions?

mgarbs:
Hi @jxcwu thank you for submitting this hip. There are plans for sharding when 10k tps is reached and sustained. When that happens there will be a hip for the implementation and you are welcome to contribute to that hip at that time. Sound good?

jxcwu:
Hi @mgarbs , thanks for your following up. After internal discussions, we believe it's necessary to bring forward our HIP at this stage. You're very welcome to share your feedback or suggestions on this. You can also refer to the original text for more context [hip link](hip-0000-improve-hedera-sharding.md) and [paper link](https://cdn.prod.website-files.com/669a53c60f11ddb32e07366a/67dda22b3182a5f07ac28d2b_Discussion%20Paper%2008-2025%20(1).pdf) Looking forward to your thoughts!

hip-0000-improve-hedera-sharding.md
jxcwu: Hi @edward-swirldslabs,

Thank you for raising this, it’s an important point. For native crypto transfers, our proposal uses shard-local escrow accounts during the prepare phase. Funds are held temporarily until a commit or rollback signal is received from the Global Committee via threshold signature (TSS). If all shards approve, the transaction finalizes; if any fail, funds are safely returned.

For smart contracts, we propose a prepare-commit model where each shard pre-executes its part of the contract and records a deterministic state delta. These changes aren’t applied until a global commit is issued. If any shard cannot complete its phase, the deltas are discarded. This ensures full atomicity across shards without partial execution risks.

To support this, we’re also looking at sandboxed execution and contract hooks (e.g., prepare(), commit(), rollback()) to handle complex logic deterministically. These additions would make cross-shard smart contract execution reliable while preserving compatibility and minimizing changes to existing contract logic.

hip-0000-improve-hedera-sharding.md
```diff
@@ -0,0 +1,175 @@
...
+
+------
+
+## Motivation
+
+Despite Hedera’s high throughput through the Gossip-about-Gossip protocol, the growing size of the network introduces challenges including:
+
+- Scalability bottlenecks from requiring each consensus node to maintain the full DAG’s recent state.
+- Increasing communication overhead from Gossip spreading every event to every node.
+- Security vulnerabilities stemming from static committee structures, making individual shards more susceptible to targeted attacks.
```
rbair23, line 31: My understanding of the current conceptual design for sharding is that committees are dynamic, not static as stated here.

hip-0000-improve-hedera-sharding.md
```diff
@@ -0,0 +1,175 @@
...
+
+- Local Committees
+  - Composed of subsets of consensus nodes.
+  - Each manages its own shard’s ledger via a localized DAG.
+  - Each committee produces Block Streams (a unified record format) of their shard history.
+  - Uses shard-specific state proofs and aligns with the network-level TSS for consensus.
+
+- Global Committee
+  - Composed of rotating coordinators from Local Committees.
+  - Handles cross-shard transaction batching and finalizes inter-shard state updates.
```
rbair23, line 78: This introduces a bottleneck into the design. I had been thinking shards would communicate point to point rather than through a central communicator.

rbair23 (commented):
There are a few key design points I'd like to get clarity on, for any sharding design proposal:

1. Is the design suitable for horizontally scaling? Or will there ultimately be a bottleneck preventing further scaling? For example, if we shard for execution but not consensus, then eventually consensus will become a bottleneck.

2. Is the design susceptible to noisy neighbors? If I have an account in shard A all alone, I have the full resources of that shard at my disposal. If a noisy neighbor makes use of shard A too, they could end up competing with me for all those resources. When I decide things have gotten too noisy and I'd like to move to another shard, how do I do so? Our current design with shard ID baked into the account ID is an example of a design that does not deal with this problem. I'd like to see the noisy neighbor problem solved.

3. How does the design account for reduction of nodes per shard? Suppose 1M nodes are available, split into 1000 different shards. Then suppose half of those nodes stop participating. What happens to those shards? Our current design with baked into shard ids doesn't handle this case. I'd like to see a solution that does.

This HIP doesn't cover these details, but only focuses on the idea of a main committee and shard committees. But in that it looks very similar to the existing design concept. The primary difference that I see is that this HIP proposes electing one member of each shard committee to also be part of the global committee. And all inter-shard transactions flow through the global committee. But this forms a bottleneck for intershard transactions. If I have 1000 shards with many cross-shard transactions then I can easily overwhelm the global committee's capacity.

hip-0000-improve-hedera-sharding.md
jxcwu: Thanks for the feedback, @rbair23. 
This HIP proposes sharding both execution and consensus, so each Local Committee finalizes its own DAG, avoiding a global consensus bottleneck. Dynamic committee reconfiguration allows for handling noisy neighbors and node churn by enabling reshuffling and adaptive shard membership. Cross-shard coordination is batched and limited to necessary transactions, with scalability options for the Global Committee as load increases.

PR #1173, opened 2025-04-09
HIP 1056: Add 8 leaf merkle tree and Trace Data improvements

Nana-EC:
**Description**:
Update HIP with details of
- Update `EventDescriptor` index replacement with correct integer size
- Add identical bool flag that can be used to highlight that a state change wrote the same value for a given entity, in which case the `MapChangeValue` value can be optimized
- Update block proof description with 8 leaf logic
- Add `TraceData` block item to capture trace data, initially smart contract `EVMTraceData` only. Also update description and protobuf
- Update Smart Contract Service protobuf with optimization and simplification of output using an `EVMActivity`
- Update images with changes to block streams items and merkle trees

**Related issue(s)**:

Fixes #

**Notes for reviewer**:
Still in draft and need verification from execution and smart contract teams.
Open to early feedback. 

Goal is to 
1. reduce footprint of block streams size 
2. support filtering ability of trace data for smart contracts without breaking equivalence
3. Clarify block streams layout

**Checklist**


HIP/hip-1056.md
```diff
@@ -821,32 +826,17 @@ message ContractCallOutput {
...
 ```
 
 ```protobuf
 message CryptoTransferOutput {
     /**
-    * Custom fees assessed during a CryptoTransfer.
-    */
+     * Custom fees assessed during a CryptoTransfer.
+     */
     repeated proto.AssessedCustomFee assessed_custom_fees = 1;
```
jsync-swirlds, line 840: This is already removed.
Custom Fees are added to the TransactionResult, and removed from all `Output` messages.

HIP/hip-1056.md
```diff
@@ -1206,16 +1176,46 @@ message MapUpdateChange {
...
+         * change has occurred in the state of the map entry, and thus
+         * the state change can be considered a no-op for the purposes of
+         * state updates.
+         * <p>
+         * If this flag is set to false, it indicates that the value being
+         * written to the map is different from the value that was already
+         * present in the map for the given key. In this case, the state
+         * change represent in value represents an actual update to the map entry
+         * <p>
+         * This field is OPTIONAL. If not set, it is assumed to be false.
```
jsync-swirlds, line 1197: Perhaps worth noting, this will always be `true` if present.  Otherwise the `oneof` must be `value`, not `identical`.

Basically the oneof makes this a presence-only entry (either present and `true` or absent and `value` is set instead). In theory we could invert this (name `not_identical` and false is meaningful) to save 1 byte every time it is "set" because the `oneof` presence tracking would ensure that we get the correct result.

The proto3 "sometimes" presence tracking makes things confusing sometimes...

HIP/hip-1056.md
```diff
@@ -1316,6 +1316,25 @@ message MapChangeKey {
 }
 ```
 
+```protobuf
+message StorageSlot {
+    /**
+     * The contract storage slot counter used to correlate slots across BlockItems for a given contract in a transactions.
+     */
+    int32 index = 1;
+
```
jsync-swirlds, line 1325: Why is this needed?  Wouldn't the `key` be enough?

HIP/hip-1056.md
```diff
@@ -1367,9 +1386,9 @@ message MapChangeValue {
         proto.ScheduleList schedule_list = 8;
 
         /**
-        * An EVM storage slot value.
-        */
-        proto.SlotValue slot = 9;
+         * A set of EVM storage slot values with associated indexes and keys.
+         */
+        repeated StorageSlot storage_slot_values = 9;
```
jsync-swirlds, line 1391: It might be a little clearer what this contains if it were named `written_slot_values`.

Also, a _little_ concerned about the extra parsing this requires.  Prior to this change we could process state changes without parsing the values; we could leave the details unparsed as a `byte[]` and still manage the inserts.  With this change we now have to parse state change values which breaks the block node "upgrade separately from consensus" idea; we would now be locked to upgrading in the same window as consensus anytime the definition of any state value changed.

HIP/hip-1056.md
```diff
@@ -1590,20 +1613,45 @@ message BlockProof {
...
+
+    /**
+     * The hash of the merkle tree root for the state changes as a result of transactions in this block.
+     */
+    bytes state_changes_root_hash = 7;
+
+    /**
+     * The hash of the merkle tree root for the trace data user transactions submitted to the network in this block.
+     */
+    bytes trace_data_root_hash = 8;
```
jsync-swirlds, line 1649: Why are these added to a block proof?  We never added them before because they must always be calculated.  This is just extra hashes that _should not_ be transmitted...

Nana-EC: Meant to remove Good catch. Thanks

Nana-EC: Removed

HIP/hip-1056.md
Nana-EC: Removed this to start

HIP/hip-1056.md
Nana-EC: Might need to rename it for clarity. I weighed index vs counter.
To avoid restating the key which for map cases can't be shrunk we could employ a counter used only in output that highlighted the storage slot changes. This way parsers of the block stream can correlate the changes in the output and trace data block items with a reduction in space.

So Given a `MapUpdateKey` - contract, then a `MapUpdateValue` (key, index, value, ...) a parser can map the write values to a given read value in the `EVMTraceData.read_values` if they are not filtered out

assets/hip-1056/protobuf/stream/block_item.proto
jsync-swirlds, line 1: Note, this is missing the following important data, replacing lines 100-101
```protobuf
 *
 * ### Forward Compatibility
 * In order to maximize forward compatibility, and minimize the need to
 * coordinate deployments of different systems creating and processing
 * block streams in the future, the following rules SHALL be followed
 * for field numbering in this message.
 * - The first 15 field numbers SHALL be assigned to the fields present
 *   in the first release. Unused fields in this range SHALL remain reserved
 *   until needed for additional options that do not fit into "input" or
 *   "output" categories.
 * - Fields numbered 16 and above MUST be numbered as follows.
 *    - "input" items MUST use `odd` field numbers.
 *    - "output" items MUST use `even` field numbers.
 *
 * #### Forward Compatibility Example
 * A future update adding three new items. A "BlockTrailer" item which is
 * neither input nor output, a new "ConsensusTransom" which is an input,
 * and a new "BridgeTransform" which is an output.
 * - The "BlockTrailer" is field 11, which is removed from the `reserved` list.
 * - The "ConsensusTransom" is an input, so it is field `17` (the first unused
 *   `odd` field greater than or equal to 16).
 * - The "BridgeTransform" is an output, so it is field `16` (the first unused
 *   even field greater than or equal to 16).
 *
 * #### Initial Field assignment to "input", "output", and "other" categories.
 * - Inputs
 *    - `event_header`
 *    - `round_header`
 *    - `event_transaction`
 * - Outputs
 *    - `block_header`
 *    - `transaction_result`
 *    - `transaction_output`
 *    - `state_changes`
 * - Any subtree (depending on what was filtered).
 *   This item details it's path in the tree.
 *    - `filtered_item_hash`
 * - Neither input nor output (and not part of the "proof" merkle tree)
 *    - `block_proof`
 *    - `record_file`
 */
message BlockItem {
    // Reserved for future items that are neither "input" nor "output".
    reserved 11,12,13,14,15;

```

HIP/hip-1056.md
Nana-EC: Fair point.
I'm inclined to think it's a necessary trade off if we want to reduce the footprint and support filtering of the trace data but let me keep thinking on it.
Let me know if you see alternative improvements

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
```diff
@@ -129,33 +103,119 @@ message ContractSystemDeleteOutput {}
...
+      * A quantity of "gas" used.<br/>
+      * This represents the resource units expended to execute this
+      * contract call, and correlates to transaction costs.
+      */
+     uint64 gas_used = 4;
+
+     /**
+      * Any Log events produced by this contract call.
+      */
+     repeated EVMTransactionLog logs = 5;
```
jsync-swirlds, line 175: These should absolutely be in the trace data, in my opinion.
Logs are debugging information, not transaction results (as I understand them).

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
```diff
@@ -129,33 +103,119 @@ message ContractSystemDeleteOutput {}
...
+      * This SHALL be encoded in RLP bytecode format.
+      */
+     bytes contractCallResult = 2;
+
+     /**
+      * Any error message produced by the contract call.
+      * <p>
+      * This SHALL be unset if the contract call succeeded.
+      */
+     string errorMessage = 3;
```
jsync-swirlds, line 163: Is there a limit on the length here?

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
```diff
@@ -129,33 +103,119 @@ message ContractSystemDeleteOutput {}
...
+ * Details maps to the log eth_getTransactionReceipt response without repeating info already available in the input transaction
+ */
+message EVMTransactionOutput {
+     /**
+      * A bloom filter produced by this contract call.<br/>
+      * Ethereum uses this bloom filter to search for call results in the
+      * Ethereum block history. High false positive rates make the bloom
+      * filters quite limited value.
+      */
+     bytes bloom = 1;
```
jsync-swirlds, line 149: I wonder if this might be better in trace data, if it's low value...

assets/hip-1056/protobuf/stream/block_item.proto
Nana-EC: Aww man, consensus node really made a lot changes that didn't fin their way back here.
Thanks for pointing this out. Will have to do a diff of all files to see

HIP/hip-1056.md
jsync-swirlds: My concern here is we're making a big change to remove a single varint, and adding a varint back in.
This basically removes any reduction in size from using the repeated list...

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: In normal coding practice yes, but on the EVM logs are considered storage and used as such. 
Filtering them out of the block stream isn't an option without losing EVM equivalence for every contract execution.
A BN/MN that was queried for a transaction in a block and was missing logs storage would be of little value to developers and DApps.
A close parallel is probably HCS Topic messages. They aren't stored in state on CN but they must be in the stream and a block streams parser that returned a topic but had no messages available would bot be useful and would break anybody who was previously streaming them

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: Good consideration, it might be @xin-hedera any thoughts here?

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: Good question. I think likely 100 bytes.
@lukelee-sl could you confirm?

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
```diff
@@ -129,33 +103,119 @@ message ContractSystemDeleteOutput {}
...
-    * This field SHALL contain all of the data produced by the contract
-    * create transaction as well as basic accounting results.
-    */
-    proto.ContractFunctionResult contract_function_result = 2;
-
+     * An ethereum transaction hash value.
+     * <p>
+     * This SHALL be a keccak256 hash of the ethereumData that represents the RLP encoded transaction.
+     */
+    bytes ethereum_hash = 2;
```
Nana-EC, line 210: Note, this in under consideration for removal. Just need to confirm there's no edge cases of recalculation resulting in different values

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
jsync-swirlds: Ye gads.  They use logs as permanent storage?  That's kind of nuts.

I suspect we _could_ still include in trace data (since that still reaches mirror node, it won't be filtered out of the block stream).  If the data in "logs" produced today has to be available 10 years from now, however, then we have to keep it here...

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
jsync-swirlds: Worth, perhaps, noting also that being in trace data doesn't mean it disappears, it's just a candidate for possible removal from the long-term storage in the future.  For example, any trace data over a year old might be trimmed for efficiency on _most_ block nodes (though some must keep it all forever to ensure the chain is complete).

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
```diff
@@ -129,33 +103,119 @@ message ContractSystemDeleteOutput {}
...
+message EVMTransactionLog {
+     /**
+      * The address of the contract emitting the log
+      */
+     string address = 1;
+
+     /**
+      * The sequence index of the log event in this contract execution
+      */
+     int32 log_index = 2;
```
Nana-EC, line 125: Need to correct the comment to be in block.
Also need to confirm if CN also has this available and if this matches the order it's written in the block stream. If so then this can be removed.
If the order of write is not matching the index in the block then this should be maintained

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: Agreed, although the impact of filtering is that any client reading a filtered stream will have missing data.
So a BN or MN that services API based on filtered data may not be able to provide users with the same expected data as today.
Hence, items that are essential should stay in output but items that say only support explorer fields additional traceability are safe to be in Trace Data section

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: Yep, very interesting right.
My understanding is it does need to be available.
For example it's common for an explorer or indexer to start up today and query logs for a contract from genesis to get the full details of activity on the contract. Not being able to query from genesis would prevent the support of some DApps.

However, I get what you're noting and I'm thinking through it. Will also confirm with SC team

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
xin-hedera: right now in mirrornode log_index is the index of a log in its EVM transaction. Is this something we try to fix?

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
xin-hedera: `bloom` is a field in out contract results REST API, and mirrornode also aggregates `bloom` filters in a block / record file to produce the `bloom` filter for a block, which is needed for the relay. mirrornode itself doesn't utilize the bloom filter to speed up queries.

So the question is if this is moved to EVM trace data, and trimmed, a new mirrornode will lose the info, will that break DApps?

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
```diff
@@ -129,33 +103,119 @@ message ContractSystemDeleteOutput {}
...
+/**
+ * The slots read during the contract execution. These slots may or may not have been overwritten.
+ * The associated writes and contract will be found in the state changes MapUpdateChange KVs
+ * Q: Should we add the ContractId in here for simplicity?
+ */
+message SlotRead {
+     /**
+      * The contract storage slot counter used to correlate slots across BlockItems for a given contract in a transactions.
+      */
+     int32 index = 1;
```
xin-hedera, line 187: how would the index work?

My understanding is the index here needs to match the index in a `StorageSlot` in state changes, and then the value read for the identified slot is here and the value written to is in the corresponding `StorageSlot`.

What about storage slots only read during EVM execution?

HIP/hip-1056.md
xin-hedera: I don't get why changing the key from `SlotKey` to a `ContractID` can save space. The same `bytes key` is still there, just moved from `SlotKey` to `StorageSlot`.

I don't fully understand the impact of changing the value to a completely new proto message `StorageSlot` instead of `SlotValue` from services state. Wouldn't that either require services to use `StorageSlot` in state or make it impossible to reconstruct state?

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: > My understanding is the index here needs to match the index in a StorageSlot in state changes, and then the value read for the identified slot is here and the value written to is in the corresponding StorageSlot.

Correct assumption.
So the hope is to replace the need for restating the storage slot address here. 
So that when a BSs parses sees a Contract-SlotAddress-Index item in the MapUpdateChange in the output then when they parse the TraceData here they can correlate this SlotRead to that item easily.
The index is a counter only for output and not in state. 
I need to confirm if the counter is across the single transaction or block. Also need to confirm there's an order already available to CN during processing or if this will be arbitrary at output time.

Nana-EC: > What about storage slots only read during EVM execution?

Hmmm, it should still be possible to have have the entry as the ContractID would still be known.
However, the key would be off so maybe index here should be a `oneof` for `index` or `key`.
If there's a partner in Changes then it's an index, if it's only read then it's the key.
Good catch

xin-hedera: yeah, agree we need a oneof and `index` is set if it's a slot written to, otherwise a value of type with both contract id and slot key need to be set in the oneof (a slot only read). We do need the contract id since there's no other place to get the info.

A size optimization we can consider, if it's a new storage slot, we can skip its slot read in the trace data, because the value is always an empty byte array. The saving is a `SlotRead` message with index set.

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
jsync-swirlds: TraceData won't be trimmed from the live stream.  A mirror node can always read it.
The only time TraceData _might_ be filtered is in long-term archives (for example only, perhaps after a year or more), or for clients that don't _want_ that data.

Moving something to the TraceData is more to make it easy for clients to filter it all out at once if they don't want that data; it should have no negative impact on any mirror node.

HIP/hip-1056.md
```diff
@@ -821,72 +826,27 @@ message ContractCallOutput {
...
-     * Contract bytecode after deployment
-     */
-    bytes runtime_bytecode = 3;
- 
-    /**
-     * The inter contract interaction details. This represents the internal EVM message frames.
+     * A receipt for an Ethereum Transaction execution
      */
-    repeated proto.ContractAction contract_actions = 4;
+    EVMTransactionReceipt evm_transaction_receipt = 1;
```
jsync-swirlds, line 831: Where is this message defined `EVMTransactionReceipt`?  I cannot seem to find it anywhere.

HIP/hip-1056.md
```diff
@@ -1206,16 +1166,46 @@ message MapUpdateChange {
...
+         * This field is OPTIONAL. If not set, it is assumed to be false.
+         * <p>
+         * Note: This field is useful for optimization purposes, allowing
+         * clients to determine whether an update was necessary or if it
+         * was a no-op. It can help reduce unnecessary state updates and
+         * improve performance in scenarios where many updates are being
+         * made to the same map entry without any actual change in value.
+         * <p>
+         */
+        bool identical = 2;
```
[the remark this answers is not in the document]
Nana-EC: To be clear you're suggesting that by having it outiosde parsers are made aware that when they get into the details of the MapChangeValue thaey may encounter null/ommitted values but metdata like pointers may still eb valid?
So in your case when `MapUpdateChange.identical` is true, then `MapUpdateChange.MapChangeValue. pending_airdrop_value` may be empty but `MapUpdateChange.MapChangeValue. previous_airdrop` and `MapUpdateChange.MapChangeValue.next_airdrop` may have valid values.

This helps and confirms another case where I wanted to make sure we could capture pointer changes on `StorageSlot` without ignoring the whole value message.

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: Checking the [execution apis spec schema](https://github.com/ethereum/execution-apis/blob/main/src/schemas/receipt.yaml#L47) it seems to be required and anyone that queries a filtered source would be unable to support a compliant `eth_getTransactionReceipt` object

HIP/hip-1056.md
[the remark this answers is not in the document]
xin-hedera: yeah agree. That's possible per the doc for `MapChangeValue` that only the changed fields are set, fyi


```proto
    /**
     * A value in a virtual map.
     * <p>
     * This value MUST correctly represent the state of the map entry
     * _after_ the asserted update.<br/>
     * This value MAY be reduced to only transmit fields that differ
     * from the prior state.<br/>
     * This field is REQUIRED.
     */
    MapChangeValue value = 2;
```

However I believe consensus node sends the full state value.

HIP/hip-1056.md
xin-hedera: believe it's named `EVMTransactionOutput` in the definition, so need to fix the reference here or the name in the definition.

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
```diff
@@ -129,33 +103,119 @@ message ContractSystemDeleteOutput {}
...
+/**
+ * EVM transaction execution log storage details
+ * Details maps to the log object in eth_getTransactionReceipt response without repeating info already available in the input transaction
+ */
+message EVMTransactionLog {
+     /**
+      * The address of the contract emitting the log
+      */
+     string address = 1;
+
```
xin-hedera, line 121: there is a `bytes bloom` field in the existing `ContractLoginfo` proto message, do we intend to remove it and what's the impact? mirrornode rest api shows each log's bloom filter in various endpoints, does relay depend on it?

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: Maybe the underlying question is what data can be filtered out but does not break the availability of core data needed to describe network activity.
With new APIs exposing Heiro data it's easier but with API's on network or already well established in web3 it may heavily impact operator and users access to information if everyone filters out because they can.
Especially without a way to easily query for "who has this missing data?"

HIP/hip-1056.md
Nana-EC: > I don't get why changing the key from SlotKey to a ContractID can save space. The same bytes key is still there, just moved from SlotKey to StorageSlot.

It's a minor saving but if `n` slots of a contract are modified then `n` `MapUpdateChange` messages are needed since the `MapChangeKey` is a `Contract-Key` combo.
In this consideration only 1 `MapUpdateChange` is needed with repeated `MapChangeValue` with no repetitions of `ContractID`

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
jsync-swirlds: Tier 1 block nodes will always have to store full history, but access will be throttled.  If you need _guaranteed_ access, the safest choice is always to run your own block node, or have a commercial contract with a provider that guarantees full unfiltered blocks. Lacking those, the next best option is to find a block node that's trustworthy, or go to a tier 1 block node if all else fails.

HIP/hip-1056.md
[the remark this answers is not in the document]
Nana-EC: @jsync-swirlds I think this is also what you were hinting to by if we change to a repeated list of slots for a single contract then CN would actually have to output the whole list of KVs vs just some.
So the restricting aspect is that to make the actually update logic easy we have to output the whole state at the given level of consideration, even if some components of it actually didn't change? In which case it's best for the MapChangeKey to be as low as possible?
I thought we already had the optimization or exception in place to emit only what was changed when needed.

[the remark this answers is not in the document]
jsync-swirlds: #### TL;DR
Yes, my question about the repeated was based on an assumption that you were proposing to change the in-state representation. Even without that there are concerns, however.

#### Discussion
So far we've defined every map update to exactly match the actual proto defined in state. We absolutely _can_ change that for a sufficient space savings; we considered doing so for Account because certain accounts are repeated for every transaction and are very large entities in state (though that kind of sputtered out in favor of trusting compression to handle it).
The remaining issue here is that (1) this change may not save any space versus the prior structure, and (2) if this change does save space, it's a small savings that probably doesn't justify the added complexity and processing cost.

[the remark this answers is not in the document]
Nana-EC: I'm agreed here.
Although I'm still curious of why a SlotKey vs just a key was needed in the slots are considered within the context of a contract and the serial nature of the BlockStream.

I see 2 options

1. (my preferred) Revert and updated SlotKey with a counter property that can be used in the context of a Contract to identify a slot instead of using the costly 256 bit key
2. Take this proposal but only if we can ensure CN only emits the changed KVs and we don't care for the ContractId being repeated for every n number of storage slot changes

HIP/hip-1056.md
jsync-swirlds: It adds index, however, plus a length for the repeated field.  The result is that for a small change set (1 or 2 slots) we may _increase_ size. For a large change set, the reduction is still small (depends a lot on the values in index), perhaps 1 or 2 bytes for each 10 entries in the list (it's hard to predict exactly).  After compression it's hard to know if there's any savings at all (only testing could show one way or the other).

Important to note, ContractID is only 1-5 bytes (depending on value), it's not large.

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: > We do need the contract id since there's no other place to get the info.

Yeah within the context of other items the counter might have to be moved into actual state in which case it exists within the context of a contract, which means yeah a ContractId needs to be repeated again

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: I believe the bloom is the same across the transaction, so each Log doesn't need a restate of the bloom right? So it's stated one for the transaction

HIP/hip-1056.md
Nana-EC: Yeah, I was concerned `EVMTransactionOutput` might give the implication there was an associated `EVMTransaction` HAPI type. So i was exploring "Receipt" instead of "Output" as that's what the evm execution APIs schema uses so the language carries over.

HIP/hip-1056.md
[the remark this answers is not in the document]
Nana-EC: Thanks for the context on the lookups.
Note the mostly 0's is not true. It's highly dependent on the contract and what data types are used. Maps are very common in smart contracts and they have no 0's to pad.
Also SC devs are highly incentivized in practice to pack storage slots to optimize gas, so we shouldn't rely on lots of padded 0's always.

Yes the removal of padding for keys should be in this proposal, let me check that i made it explicit.
I made the SlotRead use a oneof for counter vs key. We could use that counter in SlotValue instead if that's applicable - depends on the state and EVM logic. 🤔

HIP/hip-1056.md
Nana-EC: This may be flattened based on MN input

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: Hmm, the EVM defined it as across block so i think so.
Ses https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getfilterchanges

Some examples
- https://etherscan.io/tx/0xcf833ccd3f7ef7b4e39966b72a3d1a1fcdd6c0e339241ca03102ac3bacf47468#eventlog
- https://etherscan.io/tx/0x295425b3d5a1c7344a8fdde9e8004035897509d5bae953a88d588f0b7049e547#eventlog
- https://etherscan.io/tx/0x6a5dc82b3df0db293459bf13a6689d7709b75648a0662a99000c83974889ec69#eventlog
- https://etherscan.io/tx/0x19757f097b0c89024441184ed29b2246fb86fd7528ec9de7c38e2a626d1c9ee1#eventlog

HIP/hip-1056.md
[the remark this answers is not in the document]
jsync-swirlds: Thanks for the clarification.
I meant to say that _when_ padded with `0`s we could save that storage.  I did not mean to imply it was always true.

EVM compatible storage is hard to optimize, unfortunately.  Hopefully we can find some options that work well with the very different structure of Hiero ledgers.

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: Correct, but is it safe to presume a MN or client is always reading from a tier 1 or has access to it?
As you've noted it's true that any blockItem could be filtered out - so where do we draw the line

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
xin-hedera: dang, alright, so we need to fix it

https://github.com/hiero-ledger/hiero-mirror-node/blob/bc7cd8d3b0f261001cdd9f4c2735cd1381c78b07/hedera-mirror-importer/src/main/java/com/hedera/mirror/importer/domain/ContractResultServiceImpl.java#L279-L290

HIP/hip-1056.md
jsync-swirlds: We've used `ThingOutput` for all the other outputs; I don't think it's particularly confusing.  We could call it `EvmActivityOutput` if that works better and avoid connecting to either the `transaction` name or introducing inconsistency.

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: We'll have to recheck the yellow paper also. There's definitely a block level and transaction level logsBloom.
This may hint to a calculable field since we don't need both, but i'm i'm unclear that we need it per log.

HIP/hip-1056.md
Nana-EC: I wanted to avoid an inference that there is an X HAPI transaction of type "EVMTransaction" or "EVMActivity" since every other `XOutput` has an X HAPI transaction type.
Plus this message is under the 3 types of EVM transactions `ContractCallOutput`, `ContractCreateOutput` and `EthereumOutput`

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: Maybe in BSs we can set a limit. Need to confirm input as I believe these are often constrained to the HAPI response code

Nana-EC (commented):
Self review capturing some feedback

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: So the block filter at the transaction level contains all the logs in its transaction.
The bloom filter at the block level contains all the logs in the block.
They are presented to clients at request tie for easy searching via the indexer. 
It seems in RSs we aded the block at an individual log level and MN was aggregating it to the transaction and then block level.
Given the bloom filter logic MN and other indexers can create the transaction and block level bloom filter when parsing the block by starting with a 2048-bit array of zeros

1. Hash each address and topic in a log with keccak256. For each hash
2. Take the first 3 pairs of the hash e.g. from a topic `0xaabbccddeeff00112233445566778899aabbccddeeff00112233445566778899 ` it would be `aa`, `bb` and `cc`
3. Extract the lowest three 11-bit indices from each pair
4. Using the decimal of the 11 bit set the location in the bloom filter 

Repeat this for all logs in a transaction for the transaction bloom and all the logs in the block for the block logsBloom. Return the hex representation for EVM JSON RPC API responses

With this I can remove bloom from the Outputs and save quite a bit of space.
Record streams was adding 2048 bytes for every log to a block! Ethereum does it for every transaction and once for the block.
With this approach we'd save all of that. The impact is the processing work for MN and other clients, but beyond the keccak hashing it should be simple

HIP/hip-1056.md
```diff
@@ -1198,24 +1153,51 @@ message SingletonUpdateChange {
...
+      * This field is OPTIONAL. If not set, it is assumed to be false.
+      * <p>
+      * Note: This field is useful for optimization purposes, allowing
+      * clients to determine whether an update was necessary or if it
+      * was a no-op. It can help reduce unnecessary state updates and
+      * improve performance in scenarios where many updates are being
+      * made to the same map entry without any actual change in value.
+      * <p>
+      */
+    bool identical = 2;
```
xin-hedera, line 1180: nit, since this is not final, and the fields are ordered alphabetically, we should also reassigned the filed number so this becomes 1, and `key` becomes 2.

*update*: the field numbers in the proto file are correct

HIP/hip-1056.md
```diff
@@ -1707,6 +1693,77 @@ message RecordFileItem {
...
+          /**
+          * The storage value in this slot, may be left-padded with zeros to form a 256-bit word.
+          */
+          bytes read_value = 3;
+     }
+     
+     /**
+      * The contract assocaited with the storage slots this slot belongs to.
+      */
+     ContractID contractID = 1;
```
xin-hedera, line 1733: snake case here and all other places

```suggestion
     ContractID contract_id = 1;
```

HIP/hip-1056.md
```diff
@@ -1707,6 +1693,77 @@ message RecordFileItem {
...
+
+    /**
+     * The inter contract interaction details. This represents the internal EVM message frames.
+     */
+    repeated ContractAction contract_actions = 2;    
+ 
+    /**
+     * Contract slot values that were read during the execution of the EVM transaction. Associated written values will be in state changes
+     */
+    ContractSlotReads contract_slot_reads = 3;
```
xin-hedera, line 1759:
```suggestion
    repeated ContractSlotReads contract_slot_reads = 3;
```

HIP/hip-1056.md
```diff
@@ -1707,6 +1693,77 @@ message RecordFileItem {
...
+
+/**
+ * EVM tranaction execution trace details
+ * Details maps to the variable needed for debugging executions
+ */
+message EVMTraceData { 
+    /**
+     * Contract init and runtime bytecodes for child contracts that aren't in tranactions or state changes.
+     */
+    repeated ContractBytecode child_contract_bytecodes = 1;
```
xin-hedera, line 1749: should we create a new proto message with child contract id and its initcode only?

The reason is runtime bytecode is in statechanges and repeating it here is duplication.

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
```diff
@@ -129,33 +103,190 @@ message ContractSystemDeleteOutput {}
...
+/**
+ * EVM transaction execution log storage details
+ * Details maps to the log object in eth_getTransactionReceipt response without repeating info already available in the input transaction
+ * Log bloom logic is removed as it may be calculated by block parser for both transactions and block level
+ */
+message EVMTransactionLog {
+     /**
+      * The contract emitting the log. ContractID vs 20 bye address is used to preserve space
+      */
+     ContractID contractID = 1;
```
xin-hedera, line 200: can remove contract_id, trace data and evm transaction (contract create / call / eth tx) has 1:1 mapping, so all logs in a EVMTraceData should come from the same contract id (can someone confirm?). we removed the contract id in the output (though the removal in EVMTransactionResult), we should also remove it here.

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
```diff
@@ -129,33 +103,190 @@ message ContractSystemDeleteOutput {}
...
+ * 4. ethereum_hash is removed as it may be calculated by block parsers and indexers
+ */
+message EVMTraceData { 
     /**
-    * The inter contract interaction details. This represents the internal EVM message frames.
-    */
-    repeated ContractAction contract_actions = 5;
+     * The init bytecode components for child contracts created as part of the contract execution.
+     */
+     repeated ContractInitByteCode child_init_bytecodes = 1;
```
xin-hedera, line 269: sorry about the back and forth.

this shouldn't be `repeated` and we shouldn't call it out as child's info. Reason is we should fill the initcode when needed in its own contract create transaction, not the top level transaction. So what we want to do (correct me if I'm wrong) in the simplest form is

- have a `BytesValue initcode` field
- don't fill it if it's a top level contract create transaction
- fill it for child contract create transactions

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: I don't think that's the case. Contract `A` can call `B` and `C` and `B` and `C` can emit events so we'd need to disambiguate which contract it's coming from.
@lukelee-sl and @AlfredoG87?

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
Nana-EC: This is good, appreciate the focused considerations.
I think we have similar considerations.

This is intended only for child contract creations since the top level contract will have it's `init_bytecode` either in the transactionbody input or in a fileID which block stream parser can access. 
FileID should already be captured for `ContractCreate`/`EthereumTransaction` for top level transactions.
Parsers of the block stream will either need to parse the input or query the FileID contents at the time of this transaction. If there's an edge case I can remove the child reference but i'm hoping there isn't.

For contracts created as part of the execution which result in child ContractCreate transaction their init code will not be the same as the parent and it's unclear which part of the parent's `init_bytecode` are the childs.

What I'm suggestion is capturing the actual bytes (without repetition) that make up a child contract.
In most cases you would only have a population of `ContractInitByteCode.deploy_bytecode` and in some cases you'll also have `ContractInitByteCode.metadata_bytecode`.
A client can combine the items along with runtime bytecode to get a contracts full init_bytecode.

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
xin-hedera: yeah, you are right. data in our db says so:

https://testnet.mirrornode.hedera.com/api/v1/contracts/results/logs?timestamp=1740795413.512890162

two logs for this ethereum transaction, for contract id 0.0.5508307 and 0.0.5505184

HIP/hip-1056.md
Nana-EC: The runtime bytecode is only populated in the case that it is missing from state. I added a note in the full protobuf to highlight that such a case could happen.
There should be no time where both at specified for the duplication reason you noted

HIP/hip-1056.md
Nana-EC: Good note I was thinking of this too. Thanks

HIP/hip-1056.md
Nana-EC: Details are in proto but i added a quick note here

HIP/hip-1056.md
Nana-EC: Added

HIP/hip-1056.md
Nana-EC: Fixed

HIP/hip-1056.md
Nana-EC: Fixed

Nana-EC:
Will rebase off of https://github.com/hiero-ledger/hiero-improvement-proposals/pull/1056 shortly

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
xin-hedera: sure. I think my main suggestion is as follows:

The field shouldn't be repeated with `child` in the name, because one of the important outcomes of child transactions (especially for EVM transactions) in record stream record is, in general, child transaction's output is externalized in its own record, not the parents. e.g., we deprecated `ContractFunctionResult.createdContractIDs`.

So `ContractInitByteCode` should be a singular field and the data is only populated for child contract create transactions.

In addition, we can also remove `ContractInitByteCode.ContractID`.

Nana-EC: Updated

HIP/hip-1056.md
```diff
@@ -1490,18 +1473,22 @@ message FilteredItemHash {
 #### Block Proof
 
 The block proof carries the trust of the consensus network to the contents of the block. A block can conceptually be
-represented as a binary merkle tree with 4 sub-trees at depth 2 which create the block proof and tree when hashed to
+represented as a binary merkle tree with 8 sub-trees at depth 2 which create the block proof and tree when hashed to
 obtain a merkle root.
 
 ![Root Hashes](../assets/hip-1056/root_hashes.svg)
```
xin-hedera, line 1479: the svg file name here and the actual one in assets don't match, `_` here v.s. `-` in the actual filename

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
```diff
@@ -129,33 +89,186 @@ message ContractSystemDeleteOutput {}
...
+ *   Inner contract creations will have ContractCreate transations whose input and state changes should also expose this
+ *   Q: Are cases of FileID usage covered in the child cases. Can we get creative and refer to the top level FileID and maybe even byte indexes to avoid replication?
+ * 3. ContractLoginfo is not reused but shrunk into EVMTransactionLog.
+ * 4. ethereum_hash is removed as it may be calculated by block parsers and indexers
+ */
+message EVMTraceData { 
+     /**
+      * The init bytecode components for child contracts created as part of the contract execution.
+      */
+     ContractInitByteCode init_bytecodes = 1;
```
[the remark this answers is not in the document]
Nana-EC: Thanks, updated

PR #1184, opened 2025-04-23
Update hip-1086 with Batch transactions limitations

Ferparishuertas:
HIP Updated with Batch transactions limitations

### DIFF
--- a/HIP/hip-1086.md
+++ b/HIP/hip-1086.md
@@ -90,7 +90,7 @@ This HIP only permits the `EthereumTransaction` to be "jumbo" transactions. Any 
 be rejected unless specifically allowed by a future HIP.
 
 This HIP only applies to single transactions. Jumbo Ethereumtransactions will not be allowed to be executed on Batch transactions,  
-as the 6kb still applies to the batching fucntionality.
+as the 6kb still applies to the batching functionality.
 
 ### Consensus
 
--- a/HIP/hip-1086.md
+++ b/HIP/hip-1086.md
@@ -90,7 +90,7 @@ This HIP only permits the `EthereumTransaction` to be "jumbo" transactions. Any 
 be rejected unless specifically allowed by a future HIP.
 
 This HIP only applies to single transactions. Jumbo Ethereumtransactions will not be allowed to be executed on Batch transactions,  
-as the 6kb still applies to the batching functionality.
+as the standard transaction size limit still applies to the batching functionality.
 
 ### Consensus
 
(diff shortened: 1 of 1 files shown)


HIP/hip-1086.md
```diff
@@ -89,6 +89,9 @@ The pricing will be determined by the usual gas price logic in the EVM world.
 This HIP only permits the `EthereumTransaction` to be "jumbo" transactions. Any other type of transactions exceeding 6KB should 
 be rejected unless specifically allowed by a future HIP.
 
+This HIP only applies to single transactions. Jumbo Ethereumtransactions will not be allowed to be executed on Batch transactions,
```
Nana-EC, line 92: We could add a link so someone that comes to this HIP first can find more details on Jumbo Transactions if needed.
```suggestion
This HIP only applies to single transactions. [Jumbo Ethereumtransactions](https://github.com/hiero-ledger/hiero-improvement-proposals/pull/1086) will not be allowed to be executed on Batch transactions,  
```

HIP/hip-1086.md
```diff
@@ -89,6 +89,9 @@ The pricing will be determined by the usual gas price logic in the EVM world.
 This HIP only permits the `EthereumTransaction` to be "jumbo" transactions. Any other type of transactions exceeding 6KB should 
 be rejected unless specifically allowed by a future HIP.
 
+This HIP only applies to single transactions. Jumbo Ethereumtransactions will not be allowed to be executed on Batch transactions,  
+as the 6kb still applies to the batching functionality.
```
Nana-EC, line 93: We can word it out instead of a number in case Hiero ever changes the limit.
```suggestion
as the standard transaction size limit still applies to the batching functionality.
```

Nana-EC (commented):
2 suggestions for wording

HIP/hip-1086.md
Ferparishuertas: Questions

- why are we linking Jumbo to Jumbo? Your suggestion is to link to the same doc :) 
- If we words out the 6kk, which i thought to remove too, we need to remove the other mentions on the HIP

Nana-EC: whoops. I mixed up HIPs. I thought this was the Batch Transactions HIP being update.
No-op 😆

Nana-EC: Yes, the use of Hedera a configured numbers in many HIPs is something we might want to revisit.
Hiero will have configs and maybe default values. But networks like Hedera that deploy Hiero can set what they like.
We just want to be clear going forward in new HIPs or currently not completed ones that there is a config and the value is set at the network level and can change according to that networks rules.

PR #1190, opened 2025-05-07
Improvements to the Wallet Connect implementation for dApps on Hedera

mgarbs:
Description:
Updates the header to match new format
Updates the technical section to match the proposed implementation
Adds text to the reference implementation

### DIFF
--- a/HIP/hip-wallet-connect.md
+++ b/HIP/hip-wallet-connect.md
@@ -1,0 +1,181 @@
+---
+hip: TBD
+title: Improvements to the Wallet Connect implementation for dApps on Hedera
+author: Greg Scullard
+working-group: TBD
+type: Standards Track
+category: Application
+needs-hiero-review: No
+needs-hedera-approval: No
+status: Draft
+created: 2025-05-07
+discussions-to:
+requires: [820](https://github.com/hiero-ledger/hiero-improvement-proposals/blob/main/HIP/hip-820.md)
+replaces: [820](https://github.com/hiero-ledger/hiero-improvement-proposals/blob/main/HIP/hip-820.md)
+updated:
+---
+
+## Abstract
+
+This HIP proposes changes to the specification introduced in [HIP-820](https://raw.githubusercontent.com/hiero-ledger/hiero-improvement-proposals/refs/heads/main/HIP/hip-820.md) for wallets and dApps on the Hedera platform to connect using the [**Wallet Connect 2.0**](https://docs.walletconnect.com/) protocol.
+
+It introduces improved methods to facilitate native Hedera network operations such that the SDK's capabilities for handling nodes that are temporarily unavailable are used to their full extent.
+
+## Motivation
+
+Wallet connect enables dApps to work with a number of different wallets without implementing specific integration to each wallet they wish to support.
+
+The first implementation however sometimes enforced that a dApp submits a `TransactionBody` and not a full SDK `Transaction`, resulting in unnecessary failures to submit transactions to the network and degrading user experience.
+
+## Rationale
+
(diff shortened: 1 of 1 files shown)


gregscullard:
> There is a security consideration of sending a list of valid transactions to a wallet and an end user to sign only once. In a naive approach, a malicious dApp could send a disguised list of transactions to a wallet that are not the same. For example, a payload could be sent to a wallet with two transactions encapsulated in a Transaction object. The first transaction could be a transfer of 1 HBAR to a valid account, and the second transaction could be assigning a delelgated spend of 1 million HBAR to a malicious account. In a naive implementation, the wallet would prompt the user to sign both transactions as 1 and the malicious dapp would submit the malicious transaction.
> 
> For this method, it is therefore a trusted wallet's duty, burden, and responsibility to verify the transaction bodies are identical in intent before allowing a user to sign. This fortunately is handled by the @hashgraph/sdk javascript SDK.
> 
> To strengthen security assurances, while providing the desired functionality and user experience proposed by this HIP, I propose we add a test in this repository or the sdk respository to verify any given SDK version throws an unrecoverable error when trying to deserialize a transaction object in which the transaction bodies do **not** match.

There was an assumption that the SDK does check on the transaction bodies for an exact match (except node account id) which would be necessarily different for each when a serialized Transaction is de-serialized.

This doesn't appear to be the case however, and the SDK team will investigate introducing such a check.

Note that the current SDK doesn't allow manipulation of transaction lists, it would take a rogue SDK to enable this capability.

gregscullard:
> There was an assumption that the SDK does check on the transaction bodies for an exact match (except node account id) which would be necessarily different for each when a serialized Transaction is de-serialized.
> 
> This doesn't appear to be the case however, and the SDK team will investigate introducing such a check.
> 
> Note that the current SDK doesn't allow manipulation of transaction lists, it would take a rogue SDK to enable this capability.

After checking with the JS SDK team, such a check is in fact in place, see here:
https://github.com/hiero-ledger/hiero-sdk-js/blob/main/src/transaction/Transaction.js#L471-L487
This is executed whenever a transaction is de-serialized with `fromBytes`

gregscullard:
> I just need to raise my hand here and point out that this conversation itself should be noted as a red flag on the design. "Assumption" in how the SDK behaves. I've raised my objections to this design artifact of the SDKs from the beginning. A few small accidental changes in SDK internals could result in an exploit that could be overlooked for a number of versions. I am now at peace because I've said my piece (again).

@bugbytesinc Thanks and appreciate your concern, note that the assumption has been replaced with a verification that the check is in place in the SDK.

An alternative might be for the SDK to accept a serialized transaction for a singular node (or no node at all) and construct a list of transactions for random nodes, sign and return the signatures.
This would alleviate the need for the check which as you say, could fail or be inadvertently removed in the future.

It may be then, that we can either:

1. Change the `hedera_signTransaction` method to accept an optional "multinode" parameter, it would default to `false` to be backwards compatible and if set to `true`, the wallet would construct an array of transactions, however the response in this event the parameter is set to `true` would have to be different since it would need to include a SignedTransaction and public key.

2. Convey through this HIP that not having a `node_account_id` signifies to the wallet that it should create an array of transactions, sign them. This introduces the same need for a different response to an existing method which could be problematic to existing implementation.

Choosing 1 or 2 could remove the need for the new method proposed in this HIP altogether.

Note that `hedera_signAndExecuteTransaction` which was part of HIP 820 accepts a transactionList which could equally be open to exploit as you describe.

Reccetech:
@gregscullard What is the path forward on this HIP? Move forward? Stale? Withdrawn?
Thanks!

Reccetech (approved):
Approved.

PR #1193, opened 2025-05-08
Added file hip-0000.md - Records to block streams cutover

Mark-Swirlds:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


Mark-Swirlds:
Updated HIP w/ PR number and proper DCO.

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Rationale
+The transition from record streams to block streams requires careful planning to ensure network continuity, data integrity, and backward compatibility. This HIP proposes a clean cutover approach with the following key design decisions:
+
+### Bucket Structure and File Format
+
+The new block stream files will follow a revised path structure that includes network, shard, realm, and node ID components:
+
+```
+block/{network}-{YYYY-MM-DDTHH:mm}/{realm}/{shard}/{nodeID}/0000000000000000000000000000000000000.blk.gz
```
Neurone, line 44: I suggest the following:

1) Include some information in the file name as well as using a subfolder structure to store the files. This will make it easier to support multiple folder structures in the future, including a completely flat structure, while maintaining coherence and avoiding collisions.

2) Use an ID (i.e. chainID = 295) instead of a label (i.e. 'mainnet') to identify the network.

3) Move the date/time further down the structure and only use the date as a subfolder name.

4) Remove leading zeros in the block number.

In particular, I propose the following structure:

`blocks/{network ID}/{realm ID}/{shard ID}/{node ID}/YYYYMMDD/{network ID}_{realm ID}_{shard ID}_{YYYYMMDD}_{block number}.blk.zstd`

For example:

`blocks/295/0/0/10/20250901/295_0_0_20250901_83406349.blk.zstd`

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
...
+
+```
+block/mainnet/0/0/10/000000000000000000000000000083406349.blk.gz
+```
+
+- The network value will identify the network (mainnet, testnet, etc.)  
+- The shard and realm values will be included in the directory structure  
+- The nodeID will identify the consensus node that produced the block  
+- Block files will use a 36-digit zero-padded block number format  
+- Files will use gzip compression and the `.blk.gz` extension
```
Neurone, line 146: We mentioned above we are going to use zstd (`.blk.zstd`), rather than gzip.

HIP/hip-1193.md
jsync-swirlds: It should be noted here that this file structure is still very much _temporary_.
The long-term approach is that _Block Nodes_ will store blocks, and access via buckets will end. This temporary structure is intended to require minimal changes to existing clients during this interim period (so that clients can focus more resources on switching fully to block nodes as the data source).

Block Nodes (at least the Hiero implementation) use a more detailed and complex structure for local storage that makes automated access much easier; but that's an internal detail. Users of block streams will use the API to request individual blocks or a stream of blocks (with several associated options).

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
...
+Consensus nodes will implement the new path structure, for example:
+
+```
+block/mainnet/0/0/10/000000000000000000000000000083406349.blk.gz
+```
+
+- The network value will identify the network (mainnet, testnet, etc.)  
+- The shard and realm values will be included in the directory structure  
+- The nodeID will identify the consensus node that produced the block  
+- Block files will use a 36-digit zero-padded block number format
```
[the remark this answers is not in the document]
jsync-swirlds: The 36-digit length is an implementation choice in consensus node (it will never be more than 16 non-zero digits, in actuality).
The reason to choose a fixed-length value is that sorting works better and it is much easier for the software to pre-calculate reliably so that we do not need to list directories (which is abnormally expensive in cloud buckets).

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
...
+- Eases mirror node operations, enabling mirror nodes to update mirror node version before block stream cutover and have seamless service continuity.   
+- Reduces mirror node costs by not requiring periodically polling for block files.
+
+### Node Requirements
+
+Consensus nodes will:
+
+- Stop producing record and event streams at the designated cutover point  
+- Generate marker files to signal the end of record streams  
+- Begin producing block streams with proper running hash continuity
```
jsync-swirlds, line 73: As it's not technically a running hash on both sides (it's more complex than that) perhaps describe a from-to pair?
```suggestion
- Begin producing block streams with proper continuity from record stream hash to block hash.
```

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
...
+4. As a consensus node operator, I want to know exactly what changes are needed to support block streams so that I can prepare for the upgrade.
+
+5. As a mirror node operator I need to know what version of the mirror node software I need to be running and have sufficient time to upgrade to that version before the cutover  
+     
+6. As a network user, I want the transition to be seamless with no impact on transaction processing or data availability.
+
+7. As a consumer of mirror node data and associated metrics, I want the transition to be seamless with no impact on how I consume mirror node data.
+
+## Specification
+i### 1\. Cutover Point Definition
```
jsync-swirlds, line 99: typo
```suggestion
### 1\. Cutover Point Definition
```

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
...
+
+The transition from record streams to block streams will occur during a Hedera mainnet update  boundary, with the following requirements:
+
+- Occurs after a Hedera Freeze Upgrade transaction during Hedera mainnet update window  
+- Current round comes to consensus  
+- The final record stream will be completed and a marker file will be generated  
+- The first block will contain zero transactions to ensure proper formatting without impacting network usage  
+- Block continuity will be maintained by carrying forward:  
+  - The correct block number (incremented by one from the last record)  
+  - The running hash values \- cryptographic continuity of the Hedera blockchain.
```
jsync-swirlds, line 109: Hash description might be a little clearer as follows:
```suggestion
  - The last running hash value &mdash; cryptographic continuity of the Hedera blockchain.
```

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
...
+- Ensure all signatures are collected and files are properly finalized  
+- Produce final record file  
+- Stop producing both record and event streams at the designated cutover point  
+- Generate marker files to indicate the end of record and event streams
+
+#### 2.2 Block Stream Initialization
+
+Consensus nodes will:
+
+- Begin construction of the first block with zero transactions
```
jsync-swirlds, line 128: Small suggestion
```suggestion
- Construct the first block with zero transactions
```

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
...
+
+#### 2.2 Block Stream Initialization
+
+Consensus nodes will:
+
+- Begin construction of the first block with zero transactions  
+- Apply the correct running hash and block number from the final record  
+- Upload to new block path structure for uploaders
+
+NOTE:  It is required all event stream format changes are completed prior to the cutover
```
jsync-swirlds, line 132: Clarity suggestion
```suggestion
NOTE: All event stream format changes must be completed and active prior to the cutover
```

HIP/hip-1193.md
jsync-swirlds: One small suggestion: we can remove the extra characters in the date/time format to make it a _little_ shorter and more obviously intended for automation:
```suggestion
block/{network}-{YYYYMMDD_HHmm}/{realm}/{shard}/{nodeID}/0000000000000000000000000000000000000.blk.gz
```

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
...
+
+This structure provides:
+
+- Clear separation from existing record streams  
+- Support for sharding  
+- Node identification to identify source of block  
+- Block number-based naming for sequential processing  
+- ISO timestamp format of last network reset (optional field for resettable networks)
+
+Block files will use the `.blk.zstd` extension and Zstandard (Zstd) compression to optimize storage while maintaining reasonable processing performance.
```
jsync-swirlds, line 55: Shouldn't refer to ZStandard here, as the transitional files will still be gzip.
```suggestion
Block files will use the `.blk.gz` extension and GZip compression to maintain better compatibility with current record stream processing.
```

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
...
+block/{network}-{YYYY-MM-DDTHH:mm}/{realm}/{shard}/{nodeID}/0000000000000000000000000000000000000.blk.gz
+```
+
+This structure provides:
+
+- Clear separation from existing record streams  
+- Support for sharding  
+- Node identification to identify source of block  
+- Block number-based naming for sequential processing  
+- ISO timestamp format of last network reset (optional field for resettable networks)
```
jsync-swirlds, line 53: Minor clarification
```suggestion
- ISO 8601 date and time format values of last network reset (optional field for resettable networks)
```

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
...
+- Upload to new block path structure for uploaders
+
+NOTE:  It is required all event stream format changes are completed prior to the cutover
+
+#### 2.3 Path Structure Implementation
+
+Consensus nodes will implement the new path structure, for example:
+
+```
+block/mainnet/0/0/10/000000000000000000000000000083406349.blk.gz
```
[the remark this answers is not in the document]
jsync-swirlds: The date and time are optional and used for resettable networks (which mainnet is not).

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
...
+None to note
+
+## Open Issues
+Block Stream Protobuf Definitions: The block stream protobuf definitions are still under development. Several open items remain.
+
+Block Deduplication Strategy: A decision is needed on whether to implement deduplication of blocks from multiple nodes before block nodes are fully operational, or to accept the additional storage costs during the transition period.
+
+Timestamp to Block Number Mapping: A solution is needed for mirror nodes to map between timestamps and block numbers, particularly for startup and historical data access.
+## References
+A collection of URLs used as references throughout the HIP.
```
jsync-swirlds, line 263: Might want to remove this boilerplate text.

jsync-swirlds (commented):
A few mild suggestions.

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
...
+- Support the new block path structure  
+- Implement block proof verification using Threshold Signature Scheme (TSS)  
+- Support retry mechanisms for failed block ingestion
+
+## User stories
+1. As a mirror node operator, I want a clear signal when to transition from record stream processing to block stream processing, so that I don't miss any transactions during the cutover.  
+     
+2. As a mirror node operator, I want to understand the new block file format and paths so that I can adjust my infrastructure to process block streams efficiently.  
+     
+3. As a developer building on Hedera, I want to understand how the transition affects data availability and processing so that my applications continue to function correctly.
```
nadineloepfe, line 88: does this user story take into account developers or projects that currently ingest data via S3 and record files, rather than using the mirror nodes?

If so, they will likely need to rework their architecture and it would be helpful to provide guidance or migration resources to support the transition.

Mark-Swirlds: Thank you for the comment Nadine.  Happy to adjust the user story and document the impact appropriately. Do you know of devs or projects directly ingesting the files?  If yes can you share their use case and why they are using records directly.

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
...
+3. As a developer building on Hedera, I want to understand how the transition affects data availability and processing so that my applications continue to function correctly.  
+     
+4. As a consensus node operator, I want to know exactly what changes are needed to support block streams so that I can prepare for the upgrade.
+
+5. As a mirror node operator I need to know what version of the mirror node software I need to be running and have sufficient time to upgrade to that version before the cutover  
+     
+6. As a network user, I want the transition to be seamless with no impact on transaction processing or data availability.
+
+7. As a consumer of mirror node data and associated metrics, I want the transition to be seamless with no impact on how I consume mirror node data.
+
```
Nana-EC: I think there's a missing user story on the CN action for the market file.
Maybe something like
```suggestion
8. As a consensus node operator, I want to notify stream consumers with an end of record streams marker to signify the last record stream file during an upgrade.
```

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
...
+#### 2.1 Record and Event Stream Termination
+
+Consensus nodes will:
+
+- Perform Freeze operation and stop accepting transactions  
+- Freeze round comes to consensus  
+- Ensure all signatures are collected and files are properly finalized  
+- Produce final record file  
+- Stop producing both record and event streams at the designated cutover point  
+- Generate marker files to indicate the end of record and event streams
```
Nana-EC, line 121: What informs a CN to create the marker file on a given upgrade boundary?
Is it a property in a freeze transaction or a a config in a previous upgrade that notes the next upgrade will mark the end of RSs?

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
...
+```
+block/mainnet/0/0/10/000000000000000000000000000083406349.blk.gz
+```
+
+- The network value will identify the network (mainnet, testnet, etc.)  
+- The shard and realm values will be included in the directory structure  
+- The nodeID will identify the consensus node that produced the block  
+- Block files will use a 36-digit zero-padded block number format  
+- Files will use gzip compression and the `.blk.gz` extension
+
```
Nana-EC, line 146: Q: Should there be a 2.4 which highlights CN will now utilize TSS in the block stream generation?
The current thoughts around RSs to BSs cutover is that not only will the stream format change but so will the logic around singing and consensus.
Should it be doubly noted in this HIP so expectations are clear of the before and after?

HIP/hip-1193.md
```diff
@@ -0,0 +1,267 @@
+---
+hip: 1193 
+title: Record Stream to Block Stream Cutover
+author: Mark Blackman <mark@hashgraph.com>
+working-group: Steven Sheehy <@steven-sheehy>,Joseph Sinclair <Joseph Sinclair<@jsync-swirlds>, Dan Alvizu <dan.alvizu@swirldslabs.com>, Jasper Potts <jasper.potts@swirldslabs.com>, Keith Kowal <keith.kowal@swirldslabs.com>
```
Nana-EC:
```suggestion
working-group: Steven Sheehy <@steven-sheehy>, Joseph Sinclair<@jsync-swirlds>, Dan Alvizu <dan.alvizu@swirldslabs.com>, Jasper Potts <jasper.potts@swirldslabs.com>, Keith Kowal <keith.kowal@swirldslabs.com>
```

HIP/hip-1193.md
jsync-swirlds: Config in the _new_ release that triggers creating the marker file on restart (during migration processing), I believe.

HIP/hip-1193.md
jsync-swirlds: I would advocate that TSS should _predate_ cutover by some period of time.
The last thing we want is to have both TSS and cutover on the same release; that is geometrically more risk than each being separate.

jsync-swirlds: Note, it turns out TSS and cutover must be simultaneous due to interactions between the two.

HIP/hip-1193.md
jsync-swirlds: Current thinking is that a reduced scope option for ingesting data via S3 may be retained. This option will have only one copy of the data (because the block stream will have a comprehensive proof in every block) and likely a limited time duration (perhaps a month?) before block files expire.

PR #1195, opened 2025-05-13
Hiero hooks

tinker-michaelj:
**Description**:
 - Add first Hiero hooks HIP.

### DIFF
--- a/HIP/hip-1172.md
+++ b/HIP/hip-1172.md
@@ -1,0 +1,834 @@
+---
+hip: 1172
+title: Hiero hooks and an application to allowances
+author: Michael Tinker <@tinker-michaelj>
+working-group: Richard Bair <@rbair23>, Jasper Potts <@jasperpotts>, Atul Mahamuni <atul@hashgraph.com>,  Matthew DeLorenzo <@littletarzan>, Edward Wertz <@edward-swirldslabs>, Steven Sheehy <@steven-sheehy>
+requested-by: Hashgraph
+type: Standards Track
+category: Service
+needs-council-approval: Yes
+status: Draft
+created: 2025-02-19
+discussions-to: https://github.com/hiero-ledger/hiero-improvement-proposals/discussions/1172
+updated: 2025-05-13
+-------------------
+
+## Abstract
+
+We propose **hooks**, programmable Hiero extension points that let users customize the behavior of their entities.
+In principle, hooks could be programmed in any language, but we begin with **EVM hooks**. Users program EVM hooks 
+by writing contracts in a language like Solidity that compiles to EVM bytecode. EVM hooks are either **pure**
+(using neither storage nor external contracts); or **lambdas** (like code running in a cloud's event-driven compute
+offering, which may access a database to use state or call other external services). Users can install multiple
+hooks at different **indexes** on the same entity, until they reach a configured limit.
+
+As a first Hiero extension point, we propose **allowance hooks**. Users can install these hooks on their accounts.
+A Hiero API (HAPI) `CryptoTransfer` transaction can then reference a hook allowance just as it does an ERC-style
+allowance defined in [HIP-376](https://hips.hedera.com/hip/hip-376). The network uses the hook by calling its EVM bytecode at a specific function 
+signature, passing in the details of the transfers proposed in the transaction. If the hook returns `true`, the network 
+continues executing the transfer; otherwise the network rejects the transfer. Installing a hook on an account is 
+analogous to adding a function to a smart contract; the hook executes with the account's privileges when calling 
+Hedera system contracts, just as a smart contract's functions do.
+
+Unlike smart contracts, which must encapsulate trust guarantees for multiple parties, lambdas belong to a single
+owner who can directly update their storage via a `LambdaSStore` transaction. This streamlined design supports
+fast, low-cost adjustments to a lambda with less overhead than a typical `ConsensusSubmitMessage`; and far less
+than a `ContractCall`.
+
+## Motivation
+
+Hedera users often want to customize native entities instead of migrating their decentralized applications (dApps) to
+purely EVM-based smart contracts. Consider the following examples:
+- [HIP-18: Custom Hedera Token Service Fees](https://hips.hedera.com/hip/hip-18) introduced custom fee
+payments for HTS transfers.
+- [HIP-904: Frictionless Airdrops](https://hips.hedera.com/hip/hip-904) enabled more permissive token association
+policies.
+- [HIP-991: Permissionless revenue-generating Topic Ids for Topic Operators](https://hips.hedera.com/hip/hip-991)
+proposed fee-based access control for message submissions to topics.
+
+Hooks provide a more general solution to the problem of users needing custom business logic for their entities. For
+example, a token issuer might need to enforce rules on every transaction involving their token for regulatory or
+business reasons. A **transfer hook** installable on token types could enforce these rules without requiring the
+issuer to take a full design proposal through the HIP process, _or_ lose the performance and simplicity of the native
+APIs by moving everything into a smart contract.
+
+That is, by avoiding protocol-level changes for every important customization, hooks can greatly streamline innovation
+on a Hiero network while maintaining the performance and integrity of the native services.
+
+## Specification
+
+First we specify how a Hiero network will charge, throttle, and execute EVM hooks. (Non-EVM hook programming models 
+would need their own specifications.)
+
+The protobuf API for hooks in general, and EVM hooks in particular, follows in later sections.
+
+### Charging
+
+A primary concern for EVM hooks is deciding what account pays for the EVM gas upfront before executing the hook. We
+propose two charging patterns that should accommodate most use cases. The **installer** of an EVM hook chooses one of
+the patterns at the time they install it.
+1. `CALLER_PAYS` - The payer of the transaction that triggers the hook pays for its upfront gas cost. They receive the
+normal refund for unused gas.
+2. `CALLER_PAYS_ON_REVERT` - The payer of the triggering transaction again pays for the upfront gas cost, but receives
+a _full refund_ if the hook succeeds (in sense appropriate to the extension point). On success, a designated account
+that signed the lambda's installation pays for the gas actually consumed.
+
+Regardless of the charging pattern, a triggering transaction can always set an explicit gas limit for an EVM hook's
+execution, just as with a contract call. If no explicit limit is set on the transaction, the networks checks if the
+EVM was installed with a default gas limit. If there is no default limit for the hook, the network uses a global
+default, for example `hooks.evm.defaultGasLimit=25_000`.
+
+We propose the same gas price for EVM hook execution as for other contract operations. However, unlike contract calls,
+which are charged purely in gas, hook executions are already "gated" by the fee of their triggering transaction.
+So it makes sense to waive the intrinsic gas cost of their execution. We propose adding two more properties,
+
+```
+hooks.evm.pureIntrinsicGasCost=0
+hooks.evm.lambdaIntrinsicGasCost=0
+```
+
+### Throttling
+
+We propose EVM hooks be subject to the same gas throttle as top-level contract calls. Specifically, when an EVM hook
+executes, its initial EVM sender address is the payer of the referencing transaction. If this payer is throttle exempt,
+no throttles are applied. Otherwise, if the network is at capacity for gas usage, EVM hook execution will be throttled
+on that basis and the triggering transaction fail with a status of `EVM_HOOK_EXECUTION_THROTTLED`.
+
+### EVM hook execution
+
+The first message frame for the EVM transaction that executes an EVM hook will be a logical `DELEGATECALL` from the
+**hook system contract address `0x16d`** to the hook's implementing `ContractID`, sent by the payer of the upfront gas
+cost for the hook's execution.
+
+As a concrete example, suppose account `0.0.A` installs an EVM lambda hook for the `ACCOUNT_ALLOWANCE_HOOK` extension
+point at index `1`. The hook's implementing contract is `0.0.H` with EVM address `0xab...cd`. Now `0.0.B` with EVM
+address `0x01...23` sends a `CryptoTransfer` transaction that references the hook `0.0.A.1` with gas limit `100_000`.
+
+The network will construct an initial EVM message frame with,
+1. `sender` address `0x01...23`;
+2. `receiver` address `0x16d`;
+3. `contract` address `0xab...cd` (hence the source of the executing bytecode);
+4. Storage of the `0.0.A.1` lambda EVM hook; and,
+5. Gas remaining of `100_000` (no intrinsic gas cost).
+
+Since this hook is a lambda, it can then proceed with arbitrary EVM behaviors, including calls to external oracles,
+`SLOAD` and `SSTORE` operations with its storage, and so on.
+
+If the `0.0.A.1` hook is a pure EVM hook, it executes in an extremely restricted mode. Not only is the initial frame
+marked **static**, prohibiting all state-changing operations; but the network also disables the `PREVRANDAO`, `SLOAD`,
+and `CALL` opcodes. That is, a pure hook cannot do anything but apply a pure function to its input data.
+
+Although users will doubtless uncover a very wide range of use cases for EVM hooks, we expect the single most common
+type of hook contract to implement a single external method that reverts when not executed by the network _as a hook_.
+That is,
+
+```solidity
+contract HookContract {
+    function hookFunction(address installer) external payable {
+        // Revert if we are not executing a DELEGATECALL from the hook system contract
+        require(address(this) == 0x16d, "Contract can only be called as a hook");
+        // Continue executing as a hook on behalf of the installer address
+    }
+}
+```
+
+Because a hook is conceptually a function of the installing account, when it calls a Hedera system contract, the system 
+contract will see that account's address as the sender. If a hook calls another contract, it is valid for that contract 
+to call back to the hook at address `0x16d`.
+
+### Core HAPI protobufs
+
+A hook's extension point is one of an enumeration which now includes only the account allowance hook,
+```protobuf
+/***
+ * The Hiero extension points that accept a hook.
+ */
+enum HookExtensionPoint {
+  /**
+   * Used to customize an account's allowances during a CryptoTransfer transaction.
+   */
+  ACCOUNT_ALLOWANCE_HOOK = 0;
+}
+```
+
+Users install hooks by setting a `HookInstall` field on an appropriate create or update transaction. This message is,
+
+```protobuf
+/***
+ * How to install a hook.
+ */
+message HookInstall {
+  /**
+   * The extension point for the hook.
+   */
+  HookExtensionPoint extension_point = 1;
+
+  /**
+   * The entity index to install the hook at.
+   */
+  uint64 index = 2;
+
+  /**
+   * The hook implementation.
+   */
+  oneof hook {
+    /**
+     * A hook programmed in EVM bytecode that does not require access to state
+     * or interactions with external contracts.
+     */
+    PureEvmHook pure_evm_hook = 3;
+    /**
+     * A hook programmed in EVM bytecode that may access state or interact with
+     * external contracts.
+     */
+    LambdaEvmHook lambda_evm_hook = 4;
+  }
+
+  /**
+   * If set, a key that that can be used to remove or replace the hook; or (if
+   * applicable, as with a lambda EVM hook) perform transactions that customize
+   * the hook.
+   */
+  proto.Key admin_key = 5;
+
+  /**
+   * The charging pattern for the hook.
+   */
+  oneof charging_spec {
+    /**
+     * The payer of the transaction calling the hook pays for its execution.
+     */
+    CallerPays caller_pays = 6;
+    /**
+     * The payer of the transaction calling the hook pays for its execution if
+     * the hook fails; otherwise, a specified payer pays.
+     */
+    CallerPaysOnFailure caller_pays_on_failure = 7;
+  }
+}
+```
+
+The `PureEvmHook` and `LambdaEvmHook` messages share a common `EvmHookSpec` message that specifies the source of the
+hook's EVM bytecode and a default gas limit for its execution. The `LambdaEvmHook` message also includes the initial
+storage slots for a lambda hook, if desired.
+
+```protobuf
+/**
+ * Definition of a lambda EVM hook.
+ */
+message PureEvmHook {
+  /**
+   * The specification for the hook.
+   */
+  EvmHookSpec spec = 1;
+}
+
+/**
+ * Definition of a lambda EVM hook.
+ */
+message LambdaEvmHook {
+  /**
+   * The specification for the hook.
+   */
+  EvmHookSpec spec = 1;
+
+  /**
+   * Initial storage contents for the lambda, if any.
+   */
+  repeated LambdaStorageSlot storage_slots = 2;
+}
+
+/**
+ * Shared specifications for an EVM hook. May be used for any extension point.
+ */
+message EvmHookSpec {
+  /**
+   * The source of the EVM bytecode for the hook.
+   */
+  oneof bytecode_source {
+    /**
+     * The id of a contract that implements the extension point API with EVM bytecode.
+     */
+    ContractID contract_id = 1;
+  }
+
+  /**
+   * If present, the default gas limit to use when executing the EVM hook.
+   */
+  google.protobuf.UInt64Value default_gas_limit = 2;
+}
+
+/**
+ * A slot in the storage of a lambda EVM hook.
+ */
+message LambdaStorageSlot {
+  /**
+   * The 32-byte key of the slot; leading zeros may be omitted.
+   */
+  bytes key = 1;
+
+  /**
+   * If the slot is present and non-zero, the 32-byte value of the slot;
+   * leaving this field empty or setting it to binary zeros in an update
+   * removes the slot.
+   */
+  bytes value = 2;
+}
+```
+
+The indexes of newly installed hooks will appear in the legacy `TransactionReceipt` if records streams are enabled,
+
+```protobuf
+
+message TransactionReceipt {
+  // ...
+
+  /**
+   * In the receipt of a successful create or update transaction for an entity that supports hooks,
+   * the indexes of any newly installed hooks.
+   */
+  repeated uint64 installed_hook_indexes = 16;
+}
+```
+
+Once a hook is installed to an entity, a transaction generally references it by index relative to an implicit owner.
+The details of the call are specified based on its type; for example, EVM hook calls are specified by an `EvmHookCall`
+message that gives optional call data and gas limit.
+
+If the called hook does not match the given call specification, the network will fail the transaction with
+`BAD_HOOK_REQUEST`. If there is no hook installed at the specified index, the network will fail the transaction
+with `HOOK_NOT_FOUND`.
+
+```protobuf
+/**
+ * Specifies a call to a hook from within a transaction where
+ * the hook owner is implied by the point of use. (For example,
+ * it would never make sense to try to use an account allowance
+ * hook for account 0.0.X inside an AccountAmount for account
+ * 0.0.Y; hence we only need to give the index of which of
+ * 0.0.Y's hooks we want to call.)
+ */
+message HookCall {
+  /**
+   * The index of the hook to call.
+   */
+  uint64 index = 1;
+
+  /**
+   * Specifies details of the call.
+   */
+  oneof call_spec {
+    /**
+     * Specification of how to call an EVM hook.
+     */
+    EvmHookCall evm_hook_call = 2;
+  }
+}
+
+/**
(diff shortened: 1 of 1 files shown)


HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+As a first Hiero extension point, we propose **allowance hooks**. Users can install these hooks on their accounts.
+A Hiero API (HAPI) `CryptoTransfer` transaction can then reference a hook allowance just as it does an ERC-style
+allowance defined in [HIP-376](https://hips.hedera.com/hip/hip-376). The network uses the hook by calling its EVM bytecode at a specific function 
+signature, passing in the details of the transfers proposed in the transaction. If the hook returns `true`, the network 
+continues executing the transfer; otherwise the network rejects the transfer. Installing a hook on an account is 
+analogous to adding a function to a smart contract; the hook executes with the account's privileges when calling 
+Hedera system contracts, just as a smart contract's functions do.
+
+Unlike smart contracts, which must encapsulate trust guarantees for multiple parties, lambdas belong to a single
+owner who can directly update their storage via a `LambdaSStore` transaction. This streamlined design supports
```
Neurone, line 34: What is a `LambdaSStore` transaction?

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+offering, which may access a database to use state or call other external services). Users can install multiple
+hooks at different **indexes** on the same entity, until they reach a configured limit.
+
+As a first Hiero extension point, we propose **allowance hooks**. Users can install these hooks on their accounts.
+A Hiero API (HAPI) `CryptoTransfer` transaction can then reference a hook allowance just as it does an ERC-style
+allowance defined in [HIP-376](https://hips.hedera.com/hip/hip-376). The network uses the hook by calling its EVM bytecode at a specific function 
+signature, passing in the details of the transfers proposed in the transaction. If the hook returns `true`, the network 
+continues executing the transfer; otherwise the network rejects the transfer. Installing a hook on an account is 
+analogous to adding a function to a smart contract; the hook executes with the account's privileges when calling 
+Hedera system contracts, just as a smart contract's functions do.
```
Neurone, line 31: > the hook executes with the account's privileges when calling Hedera system contracts, just as a smart contract's functions do.

I think we should make the privileges model clearer. My suggestion is to make the hook run with the least privileges possible, like any other smart contract on the network.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+  proto.Key admin_key = 5;
+
+  /**
+   * The charging pattern for the hook.
+   */
+  oneof charging_spec {
+    /**
+     * The payer of the transaction calling the hook pays for its execution.
+     */
+    CallerPays caller_pays = 6;
```
Neurone, line 201: The HIP does not define the `CallerPays` type. In addition, instead of defining a new type, I suggest to use `bool` for `caller_pays`.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+
+We propose EVM hooks be subject to the same gas throttle as top-level contract calls. Specifically, when an EVM hook
+executes, its initial EVM sender address is the payer of the referencing transaction. If this payer is throttle exempt,
+no throttles are applied. Otherwise, if the network is at capacity for gas usage, EVM hook execution will be throttled
+on that basis and the triggering transaction fail with a status of `EVM_HOOK_EXECUTION_THROTTLED`.
+
+### EVM hook execution
+
+The first message frame for the EVM transaction that executes an EVM hook will be a logical `DELEGATECALL` from the
+**hook system contract address `0x16d`** to the hook's implementing `ContractID`, sent by the payer of the upfront gas
```
Neurone, line 100: > The first message frame for the EVM transaction that executes an EVM hook will be a logical `DELEGATECALL` from the **hook system contract address `0x16d`** to the hook's implementing `ContractID`

This sounds risky because it means custom code can run in the context of the system contract. Considering the hooks for the moments are lambdas or pure functions, shouldn't executing these calls as normal CALL be safer? What is the advantage of a delegate call here?

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+cost for the hook's execution.
```
Neurone, line 101: > sent by the payer of the upfront gas cost for the hook's execution.

I think _payer_ is misleading here because in some cases it can only be defined later after the transaction is executed (i.e. if CALLER_PAYS_ON_REVERT is enabled).

To make the privileges model clearer, I suggest executing all calls as if they were executed by separate entities (the hook system contract, the hook, the transaction sender, etc.) rather than assuming the privileges of the system contract or the hook installer account.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+    function hookFunction(address installer) external payable {
+        // Revert if we are not executing a DELEGATECALL from the hook system contract
+        require(address(this) == 0x16d, "Contract can only be called as a hook");
+        // Continue executing as a hook on behalf of the installer address
+    }
+}
+```
+
+Because a hook is conceptually a function of the installing account, when it calls a Hedera system contract, the system 
+contract will see that account's address as the sender. If a hook calls another contract, it is valid for that contract
```
Neurone, line 136: > Because a hook is conceptually a function of the installing account, when it calls a Hedera system contract, the system contract will see that account's address as the sender.

It seems very risky to execute with the account privilege something triggered by any external sources.
I suggest to consider the hook as a normal smart contract, using their own privileges completely separated by the caller/installer's ones. This way, a system contract will see the sender as the hook's address and not as the hook installer address.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+        // Revert if we are not executing a DELEGATECALL from the hook system contract
+        require(address(this) == 0x16d, "Contract can only be called as a hook");
+        // Continue executing as a hook on behalf of the installer address
+    }
+}
+```
+
+Because a hook is conceptually a function of the installing account, when it calls a Hedera system contract, the system 
+contract will see that account's address as the sender. If a hook calls another contract, it is valid for that contract 
+to call back to the hook at address `0x16d`.
```
Neurone, line 137: > If a hook calls another contract, it is valid for that contract to call back to the hook at address `0x16d`.

1. If the hook is always executed via a `DELEGATECALL`, the called contract will see the hook system contract as the caller, rather than the hook itself. Is this intended/correct? Regardless, I still suggest dropping the use of the `DELEGATECALL` :D.
2. What exactly does it mean that the sub-contract can call the HSC? How can this be done?
3. Which interface can I use to do this, considering that the hook system contract is a global service and, as a smart contract, I don't know exactly which hook is currently calling me?

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+ * it would never make sense to try to use an account allowance
+ * hook for account 0.0.X inside an AccountAmount for account
+ * 0.0.Y; hence we only need to give the index of which of
+ * 0.0.Y's hooks we want to call.)
+ */
+message HookCall {
+  /**
+   * The index of the hook to call.
+   */
+  uint64 index = 1;
```
Neurone, line 315: Instead of assuming and computing the value of the `AccountID` on a case-by-case basis, I suggest using the `HookID` that contains it. This would make things simpler for developers by delegating the definition of the `AccountID` to the SDK.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+ * The id of an entity that has installed a hook.
+ */
+message HookInstallerId {
+  oneof installer_id {
+    /**
+     * An account installing a hook.
+     */
+    AccountID account_id = 1;
+  }
+}
```
Neurone, line 374: Shouldn't we just use the `AccountId` and get rid of the custom `HookInstallerId` message?

Something like this:

```proto
message HookId {
  /**
   * The id of the hook's installer.
   */
  AccountID installer_id = 1;
  /**
   * A unique identifier for the hook given the installer.
   */
  uint64 index = 2;
}
```

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+message HookInstall {
+  /**
+   * The extension point for the hook.
+   */
+  HookExtensionPoint extension_point = 1;
+
+  /**
+   * The entity index to install the hook at.
+   */
+  uint64 index = 2;
```
Neurone, line 169: I think the HIP should clarify how the hook lifecycle is managed.
1. Can I overwrite a hook?
2. To update a hook, should I use the same messages that I used to install it, but specify an existing HookID?
3. What happens when I remove a hook?
4. Does the hook remain in place for historical reasons?
5. Is the hook index now free for me to use with another hook?
6. Who is going to update the prev/next hook index values?

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+    /// @param installer The address of the installing account for which the hook is being executed
+    /// @param proposedTransfers The proposed transfers
+    /// @param args The extra arguments
+    /// @return true If the proposed transfers are allowed, false or revert otherwise
+    function allow(
+       address installer,
+       ProposedTransfers memory proposedTransfers,
+       bytes memory args
+    ) external payable returns (bool);
+}
```
Neurone, line 659: I'm not sure the ABI is coherent with the current proto definition.

If we don't modify the proto messages by moving the `HookCall` outside of the single transfer operation, it seems to me that the hook will not receive data for *all transfers*, but only for the transfer in which the hook is involved.

My understanding is that, in the case of multiple transfers, the hook will be executed multiple times.
Is that correct? If so, the ABI interface and the examples below should be updated to reflect this.

Otherwise, we should clarify what happens when I define different hooks for different transfers in the same transaction. Would I receive them all? Will the hook system contract group all transfers related to the same hook?

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+    bytes32 passcodeHash;
+
+    /// Allow the proposed transfers if and only if the args are the
+    /// ABI encoding of the current one-time use passcode in storage.
+    ///
+    /// NOTE: this lambda's behavior does not depend on the installer address,
+    /// only the contents of the installed lambda's 0x00 storage slot
+    function allow(
+        address installer,
+        IHieroTransferAllowance.ProposedTransfers memory,
```
Neurone, line 689: In addition to the [above comment](https://github.com/hiero-ledger/hiero-improvement-proposals/pull/1195/files#r2091248347) about the list of transfers, it should be clearly stated that you always need to check the proposed transfer/transfers before carrying out any other checks (as we do in the second example).

As an aside, the variable name for the `ProposedTransfer` type is missing in the example — I suggest updating the example.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+}
+
+/**
+ * Specifies details of a call to an EVM hook.
+ */
+message EvmHookCall {
+  /**
+   * Extra call data to pass.
+   */
+  bytes evm_call_data = 1;
```
Neurone, line 335: Looking at the example below in the HIP description:

```proto
[...] evm_call_data: "These violent delights have violent ends" [...]
```

it seems to me that calling this parameter `evm_call_data` can be misleading.

My understanding is the **real** call data for the hook instance will be created by the hook system contract and the value in this parameter will be attached to that.

Is this correct?

If so, I suggest to make this clear in the Protobuf definition, maybe using a name like `evm_extra_call_data`, `evm_additional_call_data`, etc.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+
+Although users will doubtless uncover a very wide range of use cases for EVM hooks, we expect the single most common
+type of hook contract to implement a single external method that reverts when not executed by the network _as a hook_.
+That is,
+
+```solidity
+contract HookContract {
+    function hookFunction(address installer) external payable {
+        // Revert if we are not executing a DELEGATECALL from the hook system contract
+        require(address(this) == 0x16d, "Contract can only be called as a hook");
```
Neurone, line 129: > require(address(this) == 0x16d, "Contract can only be called as a hook");

I'm not sure what real value this check adds.

1. Can I call the hook directly via another smart contract?
2. And even if that is the case, considering the hook just returns true or false without being able of modifying the state, what is the point of not allowing its execution?

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+    }
+}
+```
+
+## Backwards Compatibility
+
+This HIP adds a net new feature to the protocol. Any account that does not install a hook will see
+identical behavior in all circumstances.
+
+## Security Implications
```
Neurone, line 793: In my opinion, in this paragraph it is necessary to clearly state and document all privileges that the hook or hook system contract will have during execution, how external contract will see them, etc.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+-------------------
+
+## Abstract
+
+We propose **hooks**, programmable Hiero extension points that let users customize the behavior of their entities.
+In principle, hooks could be programmed in any language, but we begin with **EVM hooks**. Users program EVM hooks 
+by writing contracts in a language like Solidity that compiles to EVM bytecode. EVM hooks are either **pure**
+(using neither storage nor external contracts); or **lambdas** (like code running in a cloud's event-driven compute
+offering, which may access a database to use state or call other external services). Users can install multiple
+hooks at different **indexes** on the same entity, until they reach a configured limit.
```
Neurone, line 23: > until they reach a configured limit.

This limit is mentioned here, but the HIP is missing its declaration, suggested initial value and specifications about its lifecycle.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+  /**
+   * If set, an explicit gas limit to use.
+   */
+  google.protobuf.UInt64Value gas_limit = 3;
+}
+```
+
+### Core system protobufs
+
+Once a hook is installed, it has an id in the network state.
```
Neurone, line 346: It seems the hooks are going to have an account ID.

If so, I think the HIP should also define how I can query the consensus and mirror nodes to get details about them.

Alternatively, the HIP should clearly state or reference another future HIP containing those details.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+}
+```
+
+The `PureEvmHook` and `LambdaEvmHook` messages share a common `EvmHookSpec` message that specifies the source of the
+hook's EVM bytecode and a default gas limit for its execution. The `LambdaEvmHook` message also includes the initial
+storage slots for a lambda hook, if desired.
+
+```protobuf
+/**
+ * Definition of a lambda EVM hook.
```
Neurone, line 217: This should be _pure_ EVM hook.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+created: 2025-02-19
+discussions-to: https://github.com/hiero-ledger/hiero-improvement-proposals/discussions/1172
+updated: 2025-05-13
+-------------------
+
+## Abstract
+
+We propose **hooks**, programmable Hiero extension points that let users customize the behavior of their entities.
+In principle, hooks could be programmed in any language, but we begin with **EVM hooks**. Users program EVM hooks 
+by writing contracts in a language like Solidity that compiles to EVM bytecode. EVM hooks are either **pure**
```
Neurone, line 20: What would happen if an EVM hook attempted to access EVM-specific parameters that are not applicable in a cryptoTransfer context, such as `msg.value`, `tx.origin`, `block.number`, etc.?

I think the HIP should have a dedicated section that clarifies the differences between a smart contract published inside HSCS and as a hook.

Neurone:
I suggest clarifying somewhere in the HIP whether a hook can hold or transfer funds.

kpachhai:
Had a quick look through. I think this will open up lots of use cases that weren't possible before. I had a couple of questions:
- How many hooks can an account have installed at once? Talking about the absolute limit here.
- Are hooks composable - can one hook call another? I am assuming yes but just wanted to make sure since the HIP didn't mention it.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+
+```
+hooks.evm.pureIntrinsicGasCost=0
+hooks.evm.lambdaIntrinsicGasCost=0
+```
+
+### Throttling
+
+We propose EVM hooks be subject to the same gas throttle as top-level contract calls. Specifically, when an EVM hook
+executes, its initial EVM sender address is the payer of the referencing transaction. If this payer is throttle exempt,
```
MilanWR, line 93: In what case is a payer throttle exempt?

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+no throttles are applied. Otherwise, if the network is at capacity for gas usage, EVM hook execution will be throttled
+on that basis and the triggering transaction fail with a status of `EVM_HOOK_EXECUTION_THROTTLED`.
```
gregscullard, line 95: Why introduce a new error code ? Exceeding throttles results in a `busy` message already, we should remain consistent here.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+address `0x01...23` sends a `CryptoTransfer` transaction that references the hook `0.0.A.1` with gas limit `100_000`.
+
+The network will construct an initial EVM message frame with,
+1. `sender` address `0x01...23`;
+2. `receiver` address `0x16d`;
+3. `contract` address `0xab...cd` (hence the source of the executing bytecode);
+4. Storage of the `0.0.A.1` lambda EVM hook; and,
+5. Gas remaining of `100_000` (no intrinsic gas cost).
+
+Since this hook is a lambda, it can then proceed with arbitrary EVM behaviors, including calls to external oracles,
```
gregscullard, line 114: Contracts can't call to external oracles, this should be removed.

HIP/hip-1195.md
gregscullard: Can I insert a hook between two existing hooks, or do I need to delete and re-create ?
Delete and re-create potentially breaks any compliance features I have in place where a transfer transaction may happen which should trigger a hook but doesn't because said hook is temporarily deleted.

If hooks are used for compliance purposes, it may be necessary to be able to temporarily disable transfers to (from) the account/entity while hooks are being updated (one by one) to ensure compliance isn't temporarily "bypassable"

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+    struct Transfers {
+        /// The HBAR transfers
+        IHederaTokenService.TransferList hbar;
+        /// The HTS token transfers
+        IHederaTokenService.TokenTransferList[] tokens;
+    }
+
+    /// Combines the full proposed transfers for a Hiero transaction,
+    /// including both its direct transfers and the implied HIP-18
+    /// custom fee transfers.
```
gregscullard, line 639: Will this also include transfers resulting from transaction fees to node, treasury, rewards, etc...

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+    struct ProposedTransfers {
+        /// The transaction's direct transfers
+        Transfers direct;
+        /// The transaction's assessed custom fees
+        Transfers customFee;
+    }
```
Neurone, line 645: In discussions with THF and other DAs, we found that many use cases could benefit from knowledge of the **memo field** and **network fees** associated to the transaction when deciding whether to accept a transaction.

I suggest renaming this structure to something more generic, such as `TransactionContext', and including both the above pieces of information. Or just to create a new struct and add that struct to the `allow` function parameters.

HIP/hip-1195.md
tinker-michaelj: Added section on privileges model [here](https://github.com/tinker-michaelj/hiero-improvement-proposals/blob/main/HIP/hip-1195.md#privileges-of-the-installing-entity), emphasizing we intend and affirm a model analogous to Ethereum account abstraction, where the executing code acts on behalf of the installer.

HIP/hip-1195.md
tinker-michaelj: This is to give maximum emphasis to human readers that the author intends the contract for use as a hook.

I agree that since the contract's bytecode only has access to a lambda's storage _when executed as a hook_, this `require()` does not add any security benefit. Just readability.

HIP/hip-1195.md
tinker-michaelj: > I think the HIP should clarify how the hook lifecycle is managed.
> 
> 1. Can I overwrite a hook?
> 2. To update a hook, should I use the same messages that I used to install it, but specify an existing HookID?
> 3. What happens when I remove a hook?
> 4. Does the hook remain in place for historical reasons?
> 5. Is the hook index now free for me to use with another hook?
> 6. Who is going to update the prev/next hook index values?

1. You cannot overwrite a hook, clarified [here](https://github.com/tinker-michaelj/hiero-improvement-proposals/blob/main/HIP/hip-1195.md#privileges-of-the-installing-entity).
2. The only "update" support we propose is for managing a lambda's storage via `LambdaSStore`.
3-4. It is marked deleted in state, and no longer acrues rent for your account; its storage footprint will be purged from state via a deterministic "garbage collection" process that is part of the L1 protocol.
5. Yes, now you can reuse the index.
6. Clarified in [this section](https://github.com/tinker-michaelj/hiero-improvement-proposals/blob/main/HIP/hip-1195.md#core-system-protobufs) that the `prev`/`next` pointers are an implementation detail the protocol manages and uses for efficient archival.

HIP/hip-1195.md
tinker-michaelj: Notice that the `AccountID` of the hook installer must already be in the [`AccountAmount`](https://github.com/hiero-ledger/hiero-consensus-node/blob/main/hapi/hedera-protobuf-java-api/src/main/proto/services/basic_types.proto#L444) or [`NftTransfer`](https://github.com/hiero-ledger/hiero-consensus-node/blob/main/hapi/hedera-protobuf-java-api/src/main/proto/services/basic_types.proto#L501) message.

The same will hold for all hooks; they will always be executed in a context that explicitly references the installing entity.

HIP/hip-1195.md
tinker-michaelj: Added [this section](https://github.com/tinker-michaelj/hiero-improvement-proposals/blob/main/HIP/hip-1195.md#mirror-node-and-block-explorer-support) with a suggested mirror node API for hook metadata.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+  /**
+   * The id of the hook's installer.
+   */
+  HookInstallerId installer_id = 1;
+
+  /**
+   * The call details.
+   */
+  HookCall call = 2;
+}
```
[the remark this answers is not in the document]
tinker-michaelj: Plausible, but we anticipate the `HookExecution` message will again offer a useful extension point for future HIPs.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+   * senderAccountID that must succeed for the transaction to occur.
+   */
+  HookCall sender_allowance_hook_call = 5;
+
+  /**
+   * If set, a call to a hook of type `ACCOUNT_ALLOWANCE_HOOK` installed on
+   * receiverAccountID that must succeed for the transaction to occur.
+   */
+  HookCall receiver_allowance_hook_call = 6;
+}
```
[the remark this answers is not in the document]
tinker-michaelj: This would lose the critical context on _which_ adjustment is being authorized by the call.

HIP/hip-1195.md
tinker-michaelj: No, the `HookContext` struct given to the hook (just added, see [here](https://github.com/tinker-michaelj/hiero-improvement-proposals/blob/main/HIP/hip-1195.md#mirror-node-and-block-explorer-support)) will have the fee information.

The `ProposedTransfers` are the balance adjustments/NFT ownership changes implied by the `CryptoTransfer` itself.

HIP/hip-1195.md
tinker-michaelj: Covered in more detail in [this](https://github.com/tinker-michaelj/hiero-improvement-proposals/blob/main/HIP/hip-1195.md#the-evm-environment-for-hooks) new section.

HIP/hip-1195.md
tinker-michaelj: Removed per Leemon's comments in TSC.

HIP/hip-1195.md
tinker-michaelj: This is just a _logical_ `DELEGATECALL`, in the sense that the hook's bytecode created at some contract id `0.0.C` executes with address `0x16d`. Tried to clarify further [here](https://github.com/tinker-michaelj/hiero-improvement-proposals/blob/main/HIP/hip-1195.md#the-evm-environment-for-hooks).

HIP/hip-1195.md
tinker-michaelj: Removed the `CALLER_PAYS[_ON_REVERT]` per TSC discussion.

HIP/hip-1195.md
tinker-michaelj: A valid concern; as emphasized [here](https://github.com/tinker-michaelj/hiero-improvement-proposals/blob/main/HIP/hip-1195.md#the-evm-environment-for-hooks) and discussed extensively in last two TSC calls, we expect hooks to be introduced through application HIPs that go through extensive community auditing and ecosystem adoption.

HIP/hip-1195.md
tinker-michaelj: The `0x16d` address is not a global service, it's just the "universal" address of an executing hook. Added section [here](https://github.com/tinker-michaelj/hiero-improvement-proposals/blob/main/HIP/hip-1195.md#the-evm-environment-for-hooks) to try to clarify.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+   * If non-zero, the index of the hook preceding this one in the owner's
+   * doubly-linked list of hook.
+   */
+  uint64 previous_index = 9;
+
+  /**
+   * If non-zero, the index of the hook following this one in the owner's
+   * doubly-linked list of hooks.
+   */
+  uint64 next_index = 10;
```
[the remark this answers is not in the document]
tinker-michaelj: The protocol manages these transparently. They support iterating through an account's installed hooks when doing archival in the future. Added comment in [this section](https://github.com/tinker-michaelj/hiero-improvement-proposals/blob/main/HIP/hip-1195.md#the-evm-environment-for-hooks).

HIP/hip-1195.md
tinker-michaelj: In the context of this example, the installing account just holds the one "scavenger quest" NFT prize and nothing else. It approves any transfer as long as the extended arguments have the secret passcode.

HIP/hip-1195.md
```diff
@@ -0,0 +1,834 @@
...
+marked **static**, prohibiting all state-changing operations; but the network also disables the `PREVRANDAO`, `SLOAD`,
+and `CALL` opcodes. That is, a pure hook cannot do anything but apply a pure function to its input data.
+
+Although users will doubtless uncover a very wide range of use cases for EVM hooks, we expect the single most common
+type of hook contract to implement a single external method that reverts when not executed by the network _as a hook_.
+That is,
+
+```solidity
+contract HookContract {
+    function hookFunction(address installer) external payable {
```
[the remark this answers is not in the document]
tinker-michaelj: Clarified [here](https://github.com/tinker-michaelj/hiero-improvement-proposals/blob/main/HIP/hip-1195.md#0x16d-contract-address) that besides the universal `HookContext` parameter, each extension point would have its own additional arguments.

HIP/hip-1195.md
tinker-michaelj: Delineated the two major differences [here](https://github.com/tinker-michaelj/hiero-improvement-proposals/blob/main/HIP/hip-1195.md#0x16d-contract-address); everything else is the same as a contract transaction executed from top-level HAPI.

tinker-michaelj:
> I suggest clarifying somewhere in the HIP whether a hook can hold or transfer funds.

Emphasized [here](https://github.com/tinker-michaelj/hiero-improvement-proposals/blob/main/HIP/hip-1195.md#privileges-of-the-installing-entity) that a hook, as an extension of an entity, can do anything that entity would be able to do through a Hiero system contract. (And if the entity is an account, can transfer its balance.)

HIP/hip-1195.md
tinker-michaelj: Switched to use `CONSENSUS_GAS_EXHAUSTED`.

HIP/hip-1195.md
tinker-michaelj: Fixed.

HIP/hip-1195.md
tinker-michaelj: Fixed.

HIP/hip-1195.md
Neurone: Thanks. Just to be sure: if any of the hook authorization fails, the entire transaction fails, right?

HIP/hip-1195.md
tinker-michaelj: Removed this limit and noted usage simply scales with rent.

HIP/hip-1195.md
tinker-michaelj: Yes, exactly!

mgarbs (approved):

PR #1199, opened 2025-05-20
Support block node filter

mgarbs:
## Summary
This PR adds "Block Node" as a filter option in the HIP website's category filter dropdown. This allows users to easily filter HIPs that are specifically related to Block Node functionality.

## Motivation
Block Node has recently been added as a category in HIPs like HIP-1056, but there was no matching filter option in the UI. This made it difficult for users to specifically find HIPs related to Block Node functionality.

## Changes
- Added "Block Node" option to the type-filter dropdown in the `hipstable.md` include file
- Made sure it correctly corresponds to the "block node" category tag used in HIP frontmatter
- Updated the filter.js logic to handle this new category correctly

## Testing
- Verified that HIPs with the "Block Node" category (like HIP-1056) appear when filtering by this category
- Confirmed that the filter works in combination with other filters
- Ensured that multiple category selection works properly

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,10 +8,10 @@ source 'https://rubygems.org'
 #
 # This will help ensure the proper Jekyll version is running.
 # Happy Jekylling!
-gem 'jekyll', '~> 4.3'
+gem 'jekyll', '~> 3.9.0'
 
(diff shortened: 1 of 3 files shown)


sergmetelin (approved):
LGTM

PR #1200, opened 2025-05-20
The hinTS threshold signature scheme

tinker-michaelj:
### Abstract 
Hiero networks sign blocks in the V6 record stream by each node in the network publishing an RSA signature on the hash of every block it produces as specified in [HIP-415](https://hips.hedera.com/hip/hip-415). The network's **aggregate signature** on block `N` is any combination of valid RSA **partial signatures**  from nodes holding at least 1/3 of the network stake.

This scheme is simple but inefficient. Verifiers must track the node RSA keys published in the network state, which is fairly costly. Assuming a roughly uniform stake distribution, the aggregate signature size and verification work needed both grow linearly with the number of nodes in the network. And verifying RSA signatures on EVM chains is impractical, since there 
is no native precompile support.

We propose Hiero networks adopt the **hinTS** threshold signature scheme (TSS) from [1] in tandem with adopting the block stream proposed in [HIP-1056](https://github.com/hiero-ledger/hiero-improvement-proposals/pull/1056). The hinTS scheme uses a single BLS aggregate signature that can be cheaply verified by an EVM smart contract. It achieves this by pairing each signature with a zk-SNARK proving the signature is a valid aggregation of partial signatures from BLS keys in the network address book for nodes with a threshold amount of weight.

Verifiers will verify Hiero hinTS signatures under a sequence of BLS **verification keys** specific to each network, corresponding to the chain of address books that network has adopted. Starting with the genesis address book, each time the network adopts a new address book, it will publish a recursive zk-SNARK in the block stream that proves the new address book's verification key belongs to the network's **chain of trust**.

### References
1. Garg, S., Jain, A., Mukherjee, P., Sinha, R., Wang, M., & Zhang, Y. (2023). *hinTS: Threshold Signatures with Silent
Setup*. Cryptology ePrint Archive, Paper 2023/567. Retrieved from [https://eprint.iacr.org/2023/567](https://eprint.iacr.org/2023/567)

### DIFF
--- a/HIP/hip-1183.md
+++ b/HIP/hip-1183.md
@@ -1,0 +1,478 @@
+---
+hip: 1183
+title: The hinTS threshold signature scheme 
+author: Michael Tinker <@tinker-michaelj>, Rohit Sinha <@rsinha> 
+working-group: Edward Wertz <@edward-swirldslabs>, Neeha Sompalli <@Neeharika-Sompalli>
+requested-by: Hashgraph
+type: Standards Track
+category: Service
+needs-hedera-review: Yes
+hedera-review-date: 
+hedera-approval-status: 
+needs-hiero-approval: Yes
+status: Draft
+created: 2025-04-15
+discussions-to: https://github.com/hiero-ledger/hiero-improvement-proposals/discussions/1181
+updated: 2025-05-20
+requires: 
+replaces: 
+superseded-by: 
+---
+
+## Abstract
+Hiero networks sign blocks in the V6 record stream by each node in the network publishing an 
+RSA signature on the hash of every block it produces as specified in [HIP-415](https://hips.hedera.com/hip/hip-415). 
+The network's **aggregate signature** on block `N` is any combination of valid RSA **partial signatures** 
+from nodes holding at least 1/3 of the network stake.
+
+This scheme is simple but inefficient. Verifiers must track the node RSA keys published in the 
+network state, which is fairly costly. Assuming a roughly uniform stake distribution, the 
+aggregate signature size and verification work needed both grow linearly with the number of 
+nodes in the network. And verifying RSA signatures on EVM chains is impractical, since there 
+is no native precompile support.
+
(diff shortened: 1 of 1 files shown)


HIP/hip-1200.md
```diff
@@ -0,0 +1,383 @@
...
+The appeal of the hinTS scheme is described extensively in [1], so we will not repeat each
+reason to adopt it in this HIP. The main benefit we do want to emphasize, other than
+efficiency, is that hinTS lets nodes be assigned **exact weights**, even at the resolution
+of 64-bit stake weights.
+
+This is quite attractive because for an aBFT proof-of-stake blockchain, the ideal 
+**access structure** needed to form a network signature is the agreement of any set of 
+nodes holding at least 1/3 of the total consensus weight. In a Hiero network, consensus 
+weights are the whole number of HBARs staked to each node, and hence require 64-bit 
+precision since each network has a fixed supply of 50B HBAR.
```
rbair23: Technically Hedera has a fixed supply of 50B, but another Hiero network could have a different supply?

jsync-swirlds: Agreed. This (conflating Hedera and Hiero) is something we probably need to be a lot more careful about.
Every Hiero Network will have its own native token.  HBAR is specific to Hedera.

Each Hiero network has a fixed supply of 2^63-1 of the smallest unit of the native token. For Hedera _specifically_ that is tinybar (10^-8 HBAR), which limits the maximum _theoretical_ HBAR supply to no more than ~92B tokens; Hedera chose to set the supply, therefore, at 50B.
What that native token unit is, and hence the precision of the native token, as well as its supply, is a decision for each network to make at genesis.

tinker-michaelj: Updated to clarify Hiero networks have weights with _up to_ 64-bit precision, hence an exact-weight threshold scheme is particularly attractive here. 💯

mgarbs (approved):

PR #1208, opened 2025-06-02
HIP-1208: Consensus Service Precompiled Contract for Smart Contract Service

walter-hernandez:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


kpachhai:
Putting a link to some questions I asked on the discussion post regarding this HIP: https://github.com/hiero-ledger/hiero-improvement-proposals/discussions/1207#discussioncomment-13343498

HIP/hip-1208.md
```diff
@@ -0,0 +1,78 @@
...
+Allowing smart contracts to interact natively with HCS topics enables applications to maintain tamper-proof records, trigger contract execution based on consensus-driven data, and create verifiable off-chain attestations. For example:
+- A smart contract can listen to HCS topics and execute predefined actions based on received messages.
+- HCS topics can serve as permissioned registries, where contracts validate participants' eligibility before executing transactions.
+- Multiple smart contracts on Hedera and external applications (off-graph) can read and write from the same topics, enabling many-to-many communication patterns. For example:
+  - External applications can post messages to HCS topics that smart contracts can react to.
+  - Smart contracts can emit events to topics that external applications listen to.
+- The Hedera mainnet provides the trusted consensus layer that ensures all participants within the same topics see the same sequence of events.
+
+## User stories
+- As a developer, I want my smart contract to read data from HCS topics to trigger predefined on-chain actions in a verifiable manner.
```
mgarbs: What would the specification look like for reading topics? How do you envision this working?

walter-hernandez: The HIP has been updated to focus on the writing to topics, leaving the reading from topics as needing another completely different HIP

HIP/hip-1208.md
```diff
@@ -1,78 +1,151 @@
...
 ## Rationale
-Allowing smart contracts to interact natively with HCS topics enables applications to maintain tamper-proof records, trigger contract execution based on consensus-driven data, and create verifiable off-chain attestations. For example:
-- A smart contract can listen to HCS topics and execute predefined actions based on received messages.
-- HCS topics can serve as permissioned registries, where contracts validate participants' eligibility before executing transactions.
-- Multiple smart contracts on Hedera and external applications (off-graph) can read and write from the same topics, enabling many-to-many communication patterns. For example:
-  - External applications can post messages to HCS topics that smart contracts can react to.
-  - Smart contracts can emit events to topics that external applications listen to.
-- The Hedera mainnet provides the trusted consensus layer that ensures all participants within the same topics see the same sequence of events.
+
+Allowing smart contracts to natively write to HCs topics enables applications to maintain tamper-proof records and create verifiable off-chain attestations. For example:
```
ty-swirldslabs: Small typo, "HCs topics" to "HCS topics"

walter-hernandez:
HIP updated to reflect authorship of contributors

HIP/hip-1208.md
```diff
@@ -1,78 +1,151 @@
...
-To maintain security and prevent unauthorized smart contract interactions with HCS topics:
-- Contracts must be explicitly granted permissions to interact with specific HCS topics, ensuring controlled access.
-- Contracts will not be able to delete or modify existing HCS topics. This ensures that the integrity of the topics is maintained and that the smart contract cannot interfere with the normal operation of the HCS system.
-- The creation of HCS topics will still be done through the existing HCS APIs, and the smart contract will only be able to interact with topics that have been created beforehand.
-- Messages between smart contracts and HCS will be cryptographically signed and verified, preventing tampering.
-- The proposed architecture will avoid storing private keys within smart contracts to be able to interact with topics, ensuring security best practices.
+
+To maintain security and prevent unauthorized interaction with HCS topics:
+
+- **Permission Control**: If a topic has a `submitKey`, a contract can only write to it if the contract's own ID(e.g., 0.0.1234) is part of that `submitKey`. Authorization is tied to the contract itself, not the user calling the contract.
```
ty-swirldslabs: This may be my misunderstanding, but would the contract id actually be the public key not the account id, so that the signature of the PK on the contract would be there when the message is submitted through it?

HIP/hip-1208.md
```diff
@@ -1,78 +1,151 @@
...
+- **Message Size and Content**: The precompile will enforce HCS message size limits(e.g. 1024 bytes).
+- **Topic Integrity**: Contracts will not be able to delete or modify existing HCS topics. The creation of HCS topics will still be done through existing HCS APIs via Hiero SDK.
+- **Fee Management**: The `maxHCSTxFeeBudget` must be budgeted carefully. If set too low for the message size, the child transaction will fail post-consensus, even if the parent transaction succeeded.
+- **Asynchronous Confirmation**: A `SUCCESS` response from the precompile only confirms the message was staged. Finality must be confirmed by finding the child transaction record off-chain.
```
ty-swirldslabs: "finding the child transaction record off-chain" - Can we define how you would do this, do you mean querying a mirror node?

HIP/hip-1208.md
```diff
@@ -1,78 +1,151 @@
...
+- **Interaction with Custom Topic Fees (HIP-991)**: The introduction of custom fees on HCS topics has significant security implications for this precompile, especially when used with automated systems like EVM Hooks.
+  - **Risk of Unexpected Charges**: A malicious actor could install a hook on their account that calls this precompile to write to a topic with an exorbitant custom fee. An unsuspecting user interacting with that account (e.g., via a simple `CryptoTransfer`) could be exposed to this fee without their consent.
+  - **Required Network Safeguard**: To prevent this, the network protocol must enforce that a child transaction cannot assess a custom fee against a payer unless that payer has explicitly budgeted for it in the parent transaction's `max_custom_fees` field. In the absence of this explicit consent in the parent transaction, the child transaction must fail, protecting the user from non-consensual charges.
```
ty-swirldslabs: This would change the current implementation of how custom fees works on topics, where if this variable is not present, it always is a success. 

Should this change be located to just this HSCS precompile or is this a suggestion of changing how topics work at the foundational level on how it interacts with the max_custom_fees variable?

HIP/hip-1208.md
```diff
@@ -1,78 +1,151 @@
...
+  - **hcsTransactionFee vs. Custom Fees**: It must be understood that the `maxHCSTxFeeBudget` parameter in the `submitMessage` function is a budget for network/node fees only. It does not authorize the payment of operator-defined custom fees.
+  - **Safe Use with Hooks**: For the hook use case to be safe and predictable, the hook installer should ensure the target HCS topic has the installer's key on its `FeeExemptKeyList`(as defined in HIP-991). This allows the hook to write to the topic without incurring custom fees for the user who triggers the hook, as the custom fee is waived for the exempt key.
```
ty-swirldslabs: great callout!

HIP/hip-1208.md
```diff
@@ -1,78 +1,151 @@
...
 
 ## How to Teach This
 
+Documentation will emphasize the asynchronous nature of the precompile. The recommended workflow is:
+
+1. **Call Precompile**: A developer calls `IHCS(PRECOMPILE_ADDRESS).submitMessage(topic, message, maxHCSTxFeeBudget)` from their contract and verifies the returned `responseCode` is `SUCCESS`.
```
ty-swirldslabs: Where would `max_custom_fees` variable go in this function call?

HIP/hip-1208.md
```diff
@@ -1,78 +1,151 @@
...
 ## Rejected Ideas
 
+- **Smart Contract-Based HCS Message Reading**: The ability for contracts to directly read HCS messages poses significant challenges regarding EVM determinism and efficient access to the historical state. While highly desirable, this complex feature requires a separate, dedicated proposal and is considered out of scope for this HIP, which focuses on the "write" capability.
+
 ## Open Issues
 
+- The precise gas cost for the precompile call needs to be finalized based on implementation benchmarking.
+- Whether or not to refund the unused `maxHCSTxFeeBudget` back to the user needs to be finalized.
+- Add in Topic (no custom fees) create, update, delete.
+- Add in read Topic info from the state.
```
ty-swirldslabs: I would add, How do we implement the `max_custom_fees` variable functionality as defined in this HIP, since it is a departure of current topic functionality

ty-swirldslabs (commented):
This is a great update! I've added some questions and comments for some clarity

HIP/hip-1208.md
walter-hernandez: There are two options for validating the contracts permission to read a topic:

1. Expand the functionality of `submitKey` to use the contract ID as a public key, with the HCS pre-compiled validating that the contract ID is part of `submitKey` 
2. Expand the functionality of `TopicCreateTransaction()` and `TopicUpdateTransaction()` methods during topic creating, allowing to set the contract ID(s) as an additional parameter for a private topic. So, when a contract tries to read from a topic, the HCS precompiled can validate that the contract ID of this smart contract is allowed to interact with the topic. 

We could re-write this part to clarify how this suggestion is to expand authentication as another option to `submitKey` or during topic creation, with the main purpose of avoiding to have a private key stored in the smart contract, which would make the topic vulnerable to exploitation by third parties

Which one is more feasible to implement?

HIP/hip-1208.md
walter-hernandez: Yes, correct. We will update to indicate the use of a mirror node to confirm the finality

HIP/hip-1208.md
ty-swirldslabs: I think we don't have to do anything extra if we do option 1

My understanding would be the function call going through the contract function to the topic is essentially a signature, so when we connect the precompile to allow the contract to post to topics, it'll have the contract sig, and if the submit key is the public key of that contract, it should just work. 

This should also allow users to setup topics with threshold keys with 1/n threshold, letting them have other keys to submit in addition to the contract if needed.

HIP/hip-1208.md
walter-hernandez: We can remove this part about `max_custom_fees` because it is likely to be outside the scope of this HIP. With the custom fixed fees of topics functionality (https://hips.hedera.com/hip/hip-991), it should be enough to limit the fees that someone interacting with a smart contract reading a topic would be subject to

HIP/hip-1208.md
walter-hernandez: We will remove this part about `max_custom_fees` because it is likely to be outside the scope of this HIP

HIP/hip-1208.md
walter-hernandez: Yes, correct. We add the address of the contract as its public key just like adding a regular user's public key to `submitKey`. So, when a smart contract calls a topic function, the contract's address/ID serves as its identity/authentication.

I will add this to the HIP for clarity

kpachhai: The act of the contract calling the precompile serves as its "signature" or authorization. Just to clarify, we don't need to expand or change any functionality. This is already supported by Hedera's native Key structure. A Hedera Key can be a standard public key, but it can also be set directly to a contractID (e.g., 0.0.12345). This also works perfectly with threshold keys, allowing for flexible permissions like a "1-of-N" key that includes the contract ID and other admin keys.

HIP/hip-1208.md
kpachhai: The proposal that an unset `max_custom_fees` budget defaults to zero is a deliberate design choice for user safety, and it is scoped exclusively to child transactions generated by this HSCS precompile.

We are NOT proposing a foundational change to how custom fees work on all native transactions.

The reason for this "safe anti-pattern" is the unique risk profile of the EVM especially with hooks. For example, a user making a simple token transfer has no way of knowing that a hook might trigger an action with a high custom fee. This localized safety measure will protect users in the smart contract environment from unexpected and non-consensual charges. This sort of risk doesn't exist for direct, native transactions.

HIP/hip-1208.md
kpachhai: The `max_custom_fees` variable is not a parameter in the Solidity function call.

It cannot be a function parameter because that would create a security hole. A malicious smart contract could simply ignore a user's input and substitute a much higher fee limit.

Instead, the `max_custom_fees` budget must be set on the parent `ContractCallTransaction` itself when it is created at the SDK or JSON-RPC relay level. Doing this will mean that the user's fee limit is cryptographically signed as part of their top-level transaction and cannot be tampered with by the smart contract code they are calling. 

The JSON-RPC relay and wallet are responsible for discovering the potential fee and getting the user's informed consent before setting this value. We can discuss this further if this is the correct way to go but this was my thinking at least.

HIP/hip-1208.md
kpachhai: I see it working like this: 
1. Parse Parent Transaction: When a node receives the parent `ContractCallTransaction`, it must parse and store the `max_custom_fees` budget provided by the user.
2. Tag Child Transaction: As the precompile executes and stages a child transaction (like `ConsensusSubmitMessage`), that child must be internally "tagged" with the parent's fee budget context.
3. Enforce Rule Before Charging: When processing the child transaction, the node will compare the topic's required custom fee against the tagged budget. It will apply the new rule for this context so if no budget was tagged, it is treated as zero.
4. Fail on Violation: If the required fee exceeds the budget, the child transaction fails with an appropriate error code (e.g., `CUSTOM_FEE_NOT_AUTHORIZED`), and the user is protected.

I think it makes sense to add this to the proposal. Let me do that to make it clearer.

HIP/hip-1208.md
walter-hernandez: Corrected

HIP/hip-1208.md
walter-hernandez: Corrected

HIP/hip-1208.md
walter-hernandez: Updated for clarity

HIP/hip-1208.md
walter-hernandez: @ty-swirldslabs The possible attack that could exist with the custom fees of topics + the HCS precompile would be that an attacker has a hook on their account that writes to thei custom fees topic. Then, when the attacker receives a transfer to the their account, the hook triggers writing to their custom fees topic, which would mean that the user sending the transfer to the attacker now has to pay the extra fees from the hook writing to the custom fees topic

@kpachhai for 2, what if instead of tagging, the child transaction inherits the parent's fee budget? This could handle a child transaction spawning its own child transactions

@ty-swirldslabs @kpachhai What about these options to simplify the implementation (removing tagging, budget tracking, etc.):

1. to have a hard rule for child transactions from the HCS precompile to be automatically exempt from custom fees of topics.
2. a contract ID must be set on `FeeExemptKeyList` (as well as `submitKey`) to be able to interact with a topic. 

Independent of either choice, a developer can implement whatever fee model makes sense for their use case at the smart contract level, not topic level, in this particular case of the smart contract interacting behind the scenes with a topic. The fees would be visible in the contract code and approved by the user when interacting with the smart contract.  What do you think?

HIP/hip-1208.md
jsync-swirlds: It should be [Contract ID](https://github.com/hiero-ledger/hiero-consensus-node/blob/main/hapi/hedera-protobuf-java-api/src/main/proto/services/basic_types.proto#L271) which is the [Key](https://github.com/hiero-ledger/hiero-consensus-node/blob/main/hapi/hedera-protobuf-java-api/src/main/proto/services/basic_types.proto#L909) for a contract.
Contracts do not have private keys available and cannot actually sign transactions cryptographically.

walter-hernandez: I think we agree in proceeding with Contract ID for `submitKey`. So, should we resolve this conversation?

HIP/hip-1208.md
walter-hernandez: Closing this as resolved

ty-swirldslabs (commented):
These updates LGTM

HIP/hip-1208.md
```diff
@@ -0,0 +1,189 @@
...
+- **deleteTopic**: This function is authorized by the topic’s `adminKey`. It can delete a topic regardless of whether that topic has an existing custom fee schedule.
+- **getTopicInfo**: This is a synchronous read-only call that returns the topic’s metadata. It does not generate a child transaction. It returns the comprehensive `TopicInfo` struct, providing read-only access to all topic properties, including the full custom fee schedule. This allows smart contracts to become “fee-aware”.
+- **submitMessage**: This function can submit messages to any topic, including those with custom fees. It also supports automatic message chunking. If the message payload is larger than the ~6KiB single transaction limit, the precompile will automatically calculate the number of chunks required, validate the `maxHCSTxFeeBudget` against the total cost for all chunks, and stage N child transactions automatically.
```
ty-swirldslabs: We should update this 6KiB to 1KiB, which is the max message length of 1 message on a topic, to make sure the chunking of the bytes is accurate for HCS messages.

jsync-swirlds: It might be better to specify the network property `consensus.message.maxBytesAllowed` as it's configurable and might be different for different Hiero networks.
The default is _currently_ 1024 bytes.

walter-hernandez: @ty-swirldslabs @jsync-swirlds Updated the PR with the network property `consensus.message.maxBytesAllowed`

PR #1229, opened 2025-06-27
chore: Update Contract Service output and trace specification

tinker-michaelj:
**Description**:
 - Updates the spec for the normalized and deduplicated Smart Contract trace and output data.
 - The V6 record stream is proven to be recoverable from these block stream items; please see the PRs closing the sub-issues of [this](https://github.com/hiero-ledger/hiero-consensus-node/issues/19331) issue. 
   * For 100% fidelity in recovering the legacy `ContractFunctionResult#contract_nonces` field, a stream consumer does need to maintain the current nonce of each contract account; but the _goal_ for this `ContractFunctionResult#contract_nonces` field was only to keep stream consumers up-to-date on nonces; and this is trivially achieved using the block stream `StateChanges` now.

### DIFF
--- a/HIP/hip-1056.md
+++ b/HIP/hip-1056.md
@@ -3,7 +3,7 @@ hip: 1056
 title: Block Streams
 author: >-
   Jasper Potts <@jasperpotts>, Richard Bair <@rbair23>, Nana Essilfie-Conduah <@Nana-EC>,
-  Mark Blackman <mark@swirldslabs.com>, Edward Wertz<@edward-swirldslabs>
+  Mark Blackman <mark@swirldslabs.com>, Edward Wertz<@edward-swirldslabs>, Michael Tinker <@tinker-michaelj> 
 working-group: >-
   Jasper Potts <@jasperpotts>, Richard Bair <@rbair23>, Nana Essilfie-Conduah <@Nana-EC>,
   Mark Blackman <mark@swirldslabs.com>, Leemon Baird, Joseph Sinclair <@jsync-swirlds>, Nick Poorman <@nickpoorman>,
@@ -488,11 +488,12 @@ information. Another service could filter on only accounts they are interested i
 stream are represented by one or more `FilteredItemHash`’s which provide the missing hashes required to validate the
 filtered stream with the same block proof as the full stream.
 
-> The parent-child relationship of some transactions is preserved in the Block Streams and will see the appropriate
+The parent-child relationship of some transactions is preserved in the Block Streams and will see the appropriate
 `BlockItem`s for each level. For example a `CryptoTransfer` transaction to a non existing EVM address results in the
-auto creation of a new Account with the said EVM address. In the Block Streams the `CryptoTransfer` transaction will be described by its `EventTransaction`, `TransactionResult`, `TransactionOutput` and `StateChange` `BlockItem`s. The
-`CryptoCreate` transaction will also be described by its own `EventTransaction`, `TransactionResult`, `TransactionOutput
-and `StateChange` `BlockItem`s. In this way a transactions input, output and impacts on the network are clearly described.
+auto creation of a new Account with the said EVM address. In the Block Streams the `CryptoTransfer` transaction will
+be described by its `EventTransaction`, `TransactionResult`, `TransactionOutput` and `StateChange` `BlockItem`s. The
+`CryptoCreate` transaction will also be described by its own `EventTransaction`, `TransactionResult`, `TransactionOutput`
+and `StateChange` `BlockItem`s. In this way a transactions input, output and impacts on the networkare clearly described.
 
 #### BlockHeader
 
@@ -820,20 +821,28 @@ message TransactionOutput {
 ```
 
 ```protobuf
-message ContractCallOutput {
+message CallContractOutput {
+    // Preview block streams used this field for sidecars
+    reserved 1;
+
     /**
-     * Result details for an EVM transaction execution
+     * An EVM transaction result. Contains just the information not already
+     * externalized in the matching transaction body.
      */
-    EVMTransactionResult evm_transaction_result = 1;
+    proto.EvmTransactionResult evm_transaction_result = 2;
 }
 ```
 
 ```protobuf
-message ContractCreateOutput {
+message CreateContractOutput {
+    // Preview block streams used this field for sidecars
+    reserved 1;
+
     /**
-     * Result details for an EVM transaction execution
+     * An EVM transaction result. Contains just the information not already
+     * externalized in the matching transaction body.
      */
-    EVMTransactionResult evm_transaction_result = 1;
+    proto.EvmTransactionResult evm_transaction_result = 2;
 }
 ```
 
@@ -848,10 +857,33 @@ message CreateAccountOutput {
 
 ```protobuf
 message EthereumOutput {
+   // Preview block streams used these fields for sidecars and verbose
+   // contract function results
+   reserved 2,3,4;
+
     /**
-     * Result details for an EVM transaction execution
+     * If the associated Ethereum transaction required hydrating its
+     * call data from a file, the Keccak256 hash of the transaction.
+     * <p>
+     * If the Ethereum transaction was inline, this field SHALL NOT be set.
      */
-    EVMTransactionResult evm_transaction_result = 1;
+    bytes ethereum_hash = 1;
+
+    oneof txn_result {
+        /**
+         * An EVM transaction result for an Ethereum transaction executed as a
+         * call. Contains just the information not already externalized in the
+         * matching transaction body.
+         */
+        proto.EvmTransactionResult evm_call_transaction_result = 5;
+
+        /**
+         * An EVM transaction result for an Ethereum transaction executed as a
+         * create. Contains just the information not already externalized in the
+         * matching transaction body.
+         */
+        proto.EvmTransactionResult evm_create_transaction_result = 6;
+    }
 }
 ```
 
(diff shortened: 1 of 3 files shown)


HIP/hip-1056.md
```diff
@@ -820,20 +821,28 @@ message TransactionOutput {
 ```
 
 ```protobuf
-message ContractCallOutput {
+message CallContractOutput {
+    // Preview block streams used this field for sidecars
+    reserved 1;
```
jsync-swirlds: The preview was expressly not API complete.  We should be OK reusing this field number.
We can probably apply this to all of the "preview reuse" items in this PR.

HIP/hip-1056.md
```diff
@@ -1720,116 +1752,177 @@ however every byte of information is stored fore er and needed in the proof of t
...
+message SlotRead {
+    oneof identifier {
+        /**
+        * If this slot was also written, the index of the written key in the
+        * ContractSlotUsage#slot_writes list above.
+        */
+        int32 index = 1;
+
+        /**
+        * If this slot was only read, its key; may be left-padded with zeros to 32 bytes.
```
jsync-swirlds: We would prefer not to left-pad the keys or values if we can avoid that.
I get possibly doing so during the preview period, but (ideally) I think we should probably remove any left-padding before the final cut-over to block streams.

HIP/hip-1056.md
```diff
@@ -1720,116 +1752,177 @@ however every byte of information is stored fore er and needed in the proof of t
...
     /**
-     * Any error message produced by the contract call.
+     * The logged data.
      */
-    string full_error_message = 4;
+    bytes data = 2;
 
     /**
-     * Any Log events produced by this contract call.
+     * The log's topics; may be left-padded with zeros to form 256-bit words.
```
jsync-swirlds: Similar to above, left padding is less than ideal; hopefully we can avoid doing so long-term.

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
```diff
@@ -15,24 +15,87 @@ syntax = "proto3";
...
+     */
+    uint64 gas_used = 5;
+
+    /**
+     * If not already externalized in a transaction body, the context of the
+     * internal call producing this result.
+     */
+    InternalCallContext internal_call_context = 6;
+}
+// <<<END new types from services/contract_types.proto>>>
```
jsync-swirlds: Are these actually in `contract_types.proto` and merely mirrored here for clarity, or is the comment indicating something else?

jsync-swirlds (approved):
Looks good, just some minor items to consider.

assets/hip-1056/protobuf/stream/trace/smart_contract_service.proto
```diff
@@ -0,0 +1,196 @@
...
+     * The storage slots that were written to in the EVM transaction.
+     * <p>
+     * The written value can be unambiguously derived from the block items
+     * in this EVM transaction's Hiero transactional unit. In particular,
+     * 1. If a following EVM trace modifies the same slot, the value written
+     * in this EVM transaction was whatever value the following trace read.
+     * 2. If the slot was not modified in a following EVM trace, the value
+     * is whatever value was committed to the Merkle state for the slot.
+     */
+    repeated bytes written_slot_keys = 2;
```
xin-hedera: The removal of slots written was based on the understanding that the information is already in the state changes of the corresponding top level transaction.

The `MapUpdateChange.identical` flag was also added so blockstream consumers can filter out slot `MapUpdateChange` due to `previous_key` /  `next_key` change in `SlotValue`  while there's no value written to the slot during EVM execution.

Is that not enough to reconstruct equivalent record stream `ContractStateChange`?

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
```diff
@@ -41,234 +104,83 @@ message ContractCallOutput {
...
+   // contract function results
+   reserved 2,3,4;
+
+    /**
+     * If the associated Ethereum transaction required hydrating its
+     * call data from a file, the Keccak256 hash of the transaction.
+     * <p>
+     * If the Ethereum transaction was inline, this field SHALL NOT be set.
+     */
+    bytes ethereum_hash = 1;
```
xin-hedera: `ethereum_hash` was removed to save space and blockstream consumers are supposed to do the necessary work to calculate it on their own. In case of call data in a file, consumers should know the file id, and then pull the content and calculate the Keccak256 hash.

Are there any corner cases so we have to add it back?

assets/hip-1056/protobuf/stream/trace/smart_contract_service.proto
```diff
@@ -0,0 +1,196 @@
...
+/**
+ * Information about the initcode of a contract whose creation was attempted
+ * during the EVM transaction.
+ */
+message ContractInitcode {
+    oneof code {
+        /**
+         * The initcode used for a failed top-level creation.
+         */
+        bytes failed_initcode = 2;
```
xin-hedera: for a top-level creation, if it has failed, can we not get the failed initcode directly from `ContractCreateTransactionBody.initcodeSource` (ofc it's either contained in the body as bytes or in a separate file)?

assets/hip-1056/protobuf/stream/trace/smart_contract_service.proto
```diff
@@ -0,0 +1,196 @@
...
+ * 2. Contract actions executed during the transaction (i.e., the call trace)
+ * 3. Contract slot usages, including reads and writes.
+ * 4. Events logged during the transaction.
+ * 5. Full error message, if any, produced by the contract call.
+ */
+message EVMTraceData {
+    /**
+     * The initcode of child contracts created during execution.
+     */
+    repeated ContractInitcode initcodes = 1;
```
xin-hedera: is `repeated` necessary because only the top level transaction will have statechanges? If so, we need to make sure that the information in `EVMTraceData` either only exist for the top level transaction, or we have a way to map them to corresponding child transaction's `ContractFunctionResult` / `ContractStateChange`.

assets/hip-1056/protobuf/stream/trace/smart_contract_service.proto
```diff
@@ -0,0 +1,196 @@
...
+     * Log events produced during the transaction, by contract id.
+     */
+    repeated EvmTransactionLog logs = 4;
+
+    /**
+     * Additional details of any error message in the EVM transaction result.
+     * <p>
+     * This SHALL be unset if the contract call succeeded.
+     */
+    string error_details = 5;
```
xin-hedera, line 210: assume `EvmTraceData` only exists for the top level transaction, is it a problem to externalize a child transaction's `error_details`?

HIP/hip-1056.md
Nana-EC: Agreed. We can break preview block stream output as the spec is not final and not utilized in mainnet.
That was the intent for all block stream changes until this HIP is complete and the network cuts over

HIP/hip-1056.md
```diff
@@ -848,10 +857,33 @@ message CreateAccountOutput {
...
+   // Preview block streams used these fields for sidecars and verbose
+   // contract function results
+   reserved 2,3,4;
+
     /**
-     * Result details for an EVM transaction execution
+     * If the associated Ethereum transaction required hydrating its
+     * call data from a file, the Keccak256 hash of the transaction.
+     * <p>
+     * If the Ethereum transaction was inline, this field SHALL NOT be set.
```
Nana-EC: Q: In the case of hydration the File has the transaction which can be used to calculate the hash right?
If so do we need to state the hash?
As part of the execution of the transaction a CN would have to read the File Contents anyway so there's no extra work to obtain the transaction details.

I guess the challenge is for a block stream reader that does not maintain state. 
Is that what this is trying to optimize for?

HIP/hip-1056.md
```diff
@@ -848,10 +857,33 @@ message CreateAccountOutput {
...
      */
-    EVMTransactionResult evm_transaction_result = 1;
+    bytes ethereum_hash = 1;
+
+    oneof txn_result {
```
Nana-EC: Q: Why is a `oneof` required?
The input transaction type would signify a Call or Create. 
Even in the case of an `EthereumTransaction` a call or create can be inferred by the presence of `tx.to`

Nana-EC (changes requested):
Good considerations.
Some questions and clarifications.
@xin-hedera covered my other concerns so interested in the responses to his comments.

In general, preview block streams spec can be broken so no need to preserve the numbers. Also where data can be calculated based on transaction details or retrieved from state we shouldn't be specifying it in the streams.
It's a fine balance though so there may be some cases where duplication/restating is required.

HIP/hip-1056.md
xin-hedera: from consumer perspective, would prefer the explicit result type in the `oneof`. It doesn't increase data over the wire and consumers don't have to parse the transaction in order to just figure out if it's a call or create.

steven-sheehy: We should avoid shortened, non-English words like `txn`

HIP/hip-1056.md
tinker-michaelj: Updated to reuse fields from preview block streams as applicable.

HIP/hip-1056.md
tinker-michaelj: Changed comments to emphasize that the serialized value will have its minimal byte representation (no leading zeros); but a consumer can left-pad to get a 32-byte word.

HIP/hip-1056.md
tinker-michaelj: As above.

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
tinker-michaelj: Moved to _services/contract_types.proto_ for clarity.

assets/hip-1056/protobuf/stream/trace/smart_contract_service.proto
tinker-michaelj: This looks slightly different in latest protobufs, but as emphasized in new comments, this field is for the less common case that a Hiero transactional unit includes multiple EVM transactions (in which case we have to specify which EVM transaction caused which slots to be written).

assets/hip-1056/protobuf/stream/output/smart_contract_service.proto
tinker-michaelj: Removed given your note that stream consumers should be expected to know the contents of all user-space files during ingest.

assets/hip-1056/protobuf/stream/trace/smart_contract_service.proto
tinker-michaelj: Removed this.

assets/hip-1056/protobuf/stream/trace/smart_contract_service.proto
tinker-michaelj: `repeated` was a typo, fixed; each child transaction can have its own `TraceData`.

assets/hip-1056/protobuf/stream/trace/smart_contract_service.proto
tinker-michaelj: As above.

HIP/hip-1056.md
tinker-michaelj: Removed `ethereum_hash`.

HIP/hip-1056.md
tinker-michaelj: Fixed.

HIP/hip-1056.md
```diff
@@ -503,16 +503,18 @@ in. Block items removed from a filtered stream are represented by one or more
...
+a `CryptoTransfer` transaction to a non existing EVM address results in the
+auto creation of a new Account with the said EVM address. In the Block Streams
+the `CryptoTransfer` transaction will be described by its `EventTransaction`,
+`TransactionResult`, and `TransactionOutput` and `BlockItem`s. The `CryptoCreate`
+transaction will also be described by its own `EventTransaction`,
+`TransactionResult`, and `TransactionOutput and `BlockItem`s. However, because
+both logical transactions are in the **same Hiero transactional unit**, only
+the parent `CryptoTransfer` will have a `StateChanges` item; and that item will
+include all changes for the entire transactional unit. In this way a transactions
+input, output and impacts on the network are clearly described.
```
Nana-EC, line 517: Can we add clarification on what this looks like in the Batch transaction case?
For instance are the state changes at the Batch level or at the individual parent transactions like this one. Is there any other change to the expected order of the BlockItems in that case?

HIP/hip-1056.md
```diff
@@ -874,20 +875,22 @@ message TransactionOutput {
 ```
 
 ```protobuf
-message ContractCallOutput {
+message CallContractOutput {
```
Nana-EC, line 894: Q: Why the change?
I thought at this level we had a convention of `ServiceName-Action-Object` where applicable.
Where as Service method names would be of the order you're proposing.

HIP/hip-1056.md
```diff
@@ -1790,118 +1804,192 @@ A `TraceData` block item is added to contain all information that could be cosni
...
-message EVMTransactionLog {
-     /**
-      * The contract emitting the log. ContractID vs 20 bye address is used to preserve space
-      */
-     ContractID contract_id = 1;
+message EvmTraceData {
+    /**
+     * The initcode executed for an internal contract creation.
+     */
+    ExecutedInitcode executed_initcode = 1;
```
Nana-EC: nit: I missed this before merging but should we sort this properly into
```proto
/**
 * EVM trace data, including:
 * 1. Contract actions executed during the transaction (i.e., the call trace)
 * 2. Contract slot usages, including reads and writes.
 * 3. Full error message, if any, produced by the contract call.
 * 4. Initcode used for any internal contract creations.
 * 5. Events logged during the transaction.
 */
message EvmTraceData {
    /**
     * All contract actions executed during the transaction.
     */
    repeated proto.ContractAction contract_actions = 1;

    /**
     * Contract slot usages in the transaction, by contract id.
     */
    repeated ContractSlotUsage contract_slot_usages = 2;

    /**
     * Additional details of any error message in the EVM transaction result.
     * <p>
     * This SHALL be unset if the contract call succeeded.
     */
    string error_details = 3;

    /**
     * The initcode executed for an internal contract creation.
     */
    ExecutedInitcode executed_initcode = 4;

    /**
     * Log events produced during the transaction, by contract id.
     */
    repeated EvmTransactionLog logs = 5;
}
```

HIP/hip-1056.md
```diff
@@ -1790,118 +1804,192 @@ A `TraceData` block item is added to contain all information that could be cosni
...
      */
-    google.protobuf.BytesValue deploy_bytecode = 1;
+    proto.ContractID contract_id = 1;
+
+    oneof written_keys {
+        /**
+         * If true, the list of written slot keys is contained in the non-identical
+         * state changes in the Hiero transactional unit of this EVM transaction.
+         */
+        bool written_keys_are_non_identical_state_changes = 2;
```
Nana-EC: Do we think this is needed or could it not be inferred but the rules of the spec?
If `written_slot_keys` is not empty then slot writes are fully described here, if empty then they are in state changes. This avoids the extra bool needing to be written.
Is such a savings valuable to consider or is the explicitness required?

assets/hip-1056/protobuf/services/contract_types.proto
```diff
@@ -0,0 +1,80 @@
...
+ * document are to be interpreted as described in
+ * [RFC2119](https://www.ietf.org/rfc/rfc2119) and clarified in
+ * [RFC8174](https://www.ietf.org/rfc/rfc8174).
+ */
+syntax = "proto3";
+
+package proto;
+
+// SPDX-License-Identifier: Apache-2.0
+option java_package = "com.hederahashgraph.api.proto.java";
```
Nana-EC, line 17: Since this is a new proto file is it worth putting it under the `org.hiero` package?
Maybe not at this time since it may make build and considerations more difficult.
All proto will eventually have to be moved under `org.hiero`, wondering if there's room to consider it for the block streams if the overhead isn't a lot.

assets/hip-1056/protobuf/services/contract_types.proto
```diff
@@ -0,0 +1,80 @@
...
+// <<<pbj.java_package = "com.hedera.hapi.node.contract">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+import "services/basic_types.proto";
+import "google/protobuf/wrappers.proto";
+
+/**
+ * Context of an internal call in an EVM transaction that is not otherwise externalized.<br/>
```
Nana-EC, line 25: Q: By internal do you mean Hiero internal, for instance this is a child Hiero transaction and not an internal EVM contract to contract call.
We should disambiguate between the two if possible

assets/hip-1056/protobuf/services/contract_types.proto
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/**
+ * Results of executing a EVM transaction.<br/>
+ */
+message EvmTransactionResult {
+    /**
+     * The Hedera id of the caller.<br/>
+     */
+    AccountID sender_id = 1;
```
Nana-EC, line 54: Q: Do `sender_id` and `contract_id` need to be here, don't we already have them on input?
Could they be in `InternalCallContext` assuming they are only necessary for an internal call.
In all regular EVM calls both here items are in the transaction body explicitly or RLP encoded

assets/hip-1056/protobuf/stream/trace/smart_contract_service.proto
```diff
@@ -0,0 +1,210 @@
+/**
+ * # Smart Contract Service
```
Nana-EC: I don't think the file name implied the contents, it's not a service definition.

Nana-EC (commented):
Nice updates. Thanks for the validations and modifications.
Some questions and suggestions to confirm we've got this final optimization right

assets/hip-1056/protobuf/services/contract_types.proto
jsync-swirlds: It's actually extremely hard to move _any_ proto files once published.
If we want block streams under `org.hiero` (and I agree that is a good choice), then we need to do so _now_, not after the specification becomes final and the package is baked into SDKs and dApps.

HIP/hip-1056.md
tinker-michaelj: Hmm seems all the the output protos [here](https://github.com/hiero-ledger/hiero-improvement-proposals/tree/main/assets/hip-1056/protobuf/stream/output) have a `<Verb><Entity>Output` naming pattern (`CreateTokenOutput`, etc). Not sure it matters...should we have a separate PR to switch to `<Entity><Verb>Output`?

HIP/hip-1056.md
tinker-michaelj: PBJ and `protoc`-generated Java objects don't let you distinguish between "empty" and "omitted" `repeated` lists. 

In any case we definitely don't need to worry about the occasional extra bit in the stream. 😁

assets/hip-1056/protobuf/services/contract_types.proto
tinker-michaelj: Hmm I don't know what the goal is for packages; @jsync-swirlds if you want to recommend a package for this, happy to update.

assets/hip-1056/protobuf/services/contract_types.proto
tinker-michaelj: We _could_ shift the responsibility to stream consumer of resolving aliases to `0.0.X` ids, but this is just a vanishing detail in the big picture, and not worth it imo.

HIP/hip-1056.md
tinker-michaelj: Added a paragraph (below) emphasizing that all batch state changes are associated to the `AtomicBatch`, not to the logical transactions within it.

assets/hip-1056/protobuf/services/contract_types.proto
tinker-michaelj: Added a sentence emphasizing this is an internal message call in an EVM transaction, _not_ a logical transaction in a Hiero transactional unit.

HIP/hip-1056.md
tinker-michaelj: Re-ordered alphabetically.

assets/hip-1056/protobuf/stream/trace/smart_contract_service.proto
tinker-michaelj: Switchd to `Smart Contract Service Trace Data`

assets/hip-1056/protobuf/services/contract_types.proto
jsync-swirlds: For existing _public API_ files we have to keep what is there for now because any change has a very large "impact radius".

For _new_ files, the following pattern is _recommended_ (with limited exceptions).
```proto
package org.hiero.system.service.subpackage;

// the following two are not required if consensus changes the PBJ build options.
option java_package = "org.hiero.system.service.subpackage.legacy";
// This next line should match the main package when that value is not `proto`.
// <<<pbj.java_package = "org.hiero.system.service.subpackage">>> This comment is special code for setting PBJ Compiler java package
```

That said, this is an _existing public API file_ and we, therefore need to leave it as-is for now (future project, and HIP, may migrate to the `org.hiero` namespace, but that affects all SDKs and isn't a small change).

HIP/hip-1056.md
tinker-michaelj: _Update_ - actually when I replied above I forgot I had introduced the `WrittenSlotKeys` message; removed this `oneof`.

mgarbs (approved):

HIP/hip-1056.md
Nana-EC: Yeah, a separate PR makes sense. Let's just make sure execution code uses the `<Entity><Verb>Output` initially so that we minimize disruption for downstream integrators

assets/hip-1056/protobuf/services/contract_types.proto
Nana-EC: Okay.
Anyone with state can resolve it and others would have to query a state hosting service like MN.
Once it's in the proto we can't take it out so we should only keep it if needed

Nana-EC (approved):
LG.
Unblocking, other items can be addressed in follow up PRs.
Let's just make sure we get the proto items right soon so we reduce breaking changes to downstreams consumers

xin-hedera (approved):
LGTM

PR #1238, opened 2025-07-11
HIP-1238: Quiescence

lpetrovic05:
**Description**:
This HIP proposes a new feature for Hiero networks, quiescence. The Quiescence feature introduces a mechanism to pause event creation in Hiero networks during periods of no transaction activity, optimizing resource usage in low-volume networks.

**Related issue(s)**:
N/A

**Notes for reviewer**:
N/A

**Checklist**

### DIFF
--- a/HIP/hip-1238.md
+++ b/HIP/hip-1238.md
@@ -10,10 +10,10 @@ needs-hiero-review: Yes
 needs-hedera-approval: Yes
 hedera-review-date:
 hedera-approval-status:
-status: Review
+status: Approved
 created: 2025-07-11
 discussions-to: https://github.com/hiero-ledger/hiero-consensus-node/pull/20069
-updated: 2025-07-15
+updated: 2025-09-02
 ---
 
 ## Abstract


HIP/hip-1238.md
```diff
@@ -0,0 +1,211 @@
...
+## Reference Implementation
+
+No reference implementation exists for this HIP at the moment.
+
+## Rejected Ideas
+
+### TCT alternative
+
+One of the rejected ideas was a substitute for [Rule 3](#rule-3-target-consensus-timestamp-tct). Instead of event
+creation keeping track of TCT, the idea was to submit a no-op transaction to the network at this time. The main benefit
```
poulok:
```suggestion
One of the rejected ideas was a substitute for [Rule 3](#rule-3-target-consensus-timestamp-tct). Instead of keeping track of TCT, the idea was to submit a no-op transaction to the network at this time. The main benefit
```

HIP/hip-1238.md
```diff
@@ -0,0 +1,210 @@
...
+quiesce, or will break quiescence if already quiesced.
+
+#### Quiescing
+
+Once all the conditions are met, we can quiesce. This is done by simply stopping event creation. The consensus module
+will stay in this state until one of the following conditions is met:
+
+- A transaction is submitted to the node that needs to be put into an event
+- A node sends us an event with transactions that need to reach consensus and become part of a fully signed block
+- `wallClockTime` + `tctDuration` is less than the next target consensus timestamp (TCT)
```
jsync-swirlds: Do you mean _greater than_ here?
```suggestion
- `wallClockTime` + `tctDuration` is greater than the next target consensus timestamp (TCT)
```

The wall clock time + TCT duration _should_ always be less than the next TCT while quiescent.

HIP/hip-1238.md
```diff
@@ -0,0 +1,210 @@
...
+
+The only impact on Hiero Mirror node is that blocks will not flow constantly. If the network is quiesced, the mirror
+node will not receive blocks until the network breaks quiescence. The mirror node will need to distinguish between
+a failure and a quiesced network, since they can appear similar.
+
+### Impact on SDK
+
+No impacts on Heiro SDKs are expected.
+
+## Backwards Compatibility
```
jsync-swirlds, line 162: Should we add a section for impacts on block node?
We could discuss how a block node must detect quiescence and allow blocks to not exist for a long period, while also making some small effort to ensure that all block nodes detect the end of quiescence in a timely manner, even if the block node does not directly receive a new `publishBlockStream` request.

HIP/hip-1238.md
```diff
@@ -0,0 +1,210 @@
...
+There are no impacts to backwards compatibility.
+
+## Security Implications
+
+There are no security implications of this HIP. The worst case scenario is that the network does not quiesce when it
+should, which leads to increased resource usage. This is not a security issue, but rather an efficiency issue.
+
+## How to Teach This
+
+There is nothing to teach.
```
jsync-swirlds: I disagree, quiescence is somewhat complex and we should have a plan for how the concepts and nuances can be taught to others.

poulok (approved):

HIP/hip-1238.md
poulok: I think it depends on the audience this section is meant for. If it includes people wanting to implement their own consensus node, then perhaps. But the technical section above does that IMO (without getting into the implementation details). I don't think this needs to be taught to Hedera users.

lpetrovic05: Yeah, I also understand this as how it should be taught to end users. Node operators need to be aware of this, but I'm not sure this belongs here. Wdyt @jsync-swirlds ?

HIP/hip-1238.md
lpetrovic05: should we have both mirror node and block node? Or should this be a single section, something like "downstream impacts".

jsync-swirlds: I think it's reasonable to have both.  Particularly as the Block Node is a critical element of the network (if the block nodes all go offline, the network stops) while mirror node is merely "very important" (if mirror node is offline, the network can continue, though many dApps will fail).

HIP/hip-1238.md
jsync-swirlds: I think it may well be of interest to end users, particularly for Hiero networks run privately (e.g. Spheres). The quiescence feature is, at least in some respects, a selling point, and I believe we should at least consider the need to teach this feature to people who select, use, or install an Hiero network, rather than those implementing the software.

poulok: Understood. What is there to teach though? The network will work exactly as it normally does to the end user, with the exception of less resources used which is not really an end user concern.

jsync-swirlds: With block streams, the activity of the network is pretty visible. I am no expert on how to teach or what to teach, but here are some ideas that immediately come to my mind.
* Your Hiero network may stop producing blocks for a period of time, this is normal due to a feature called "quiescence" and reduces resource requirements for your network.
* Your Hiero network is optimized to reduce energy usage and emissions. One of the features that supports this optimization is called "quiescence" where the network pauses when there is no work to be done. You may be able, depending on your usage, to further reduce your energy usage by grouping transactions together to extend this "quiescence" period.
* A Hiero network may enter a "quiescent" state to minimize resource consumption during idle periods, if any. This state reduces the cost to operate a private Hiero network and also helps to maintain a secure and private network by not generating new consensus rounds when there is no demand on the network.

I am certain that DA and others can come up with more and better suggestions than I have.

HIP/hip-1238.md
lpetrovic05: I added a section for it, but I don't really know what to put in it

HIP/hip-1238.md
lpetrovic05: I added the second one

HIP/hip-1238.md
jsync-swirlds: What is added now appears sufficient.
Block Node will need to have some mechanism to determine the difference between a consensus node publisher that just isn't sending data versus a quiescent network; but that's more a detail to be worked on in the block node based on the stated impact.

HIP/hip-1238.md
```diff
@@ -0,0 +1,219 @@
...
+## Motivation
+
+Current Hiero mainnet specifications mandate continuous event creation and gossip, even in the absence of transaction
+activity, to maintain consensus and network state. While robust for consistent network traffic, this approach is
+inefficient for networks with periodic usage where prolonged periods without transactions are common.
+
+The constant generation of events and blocks in such scenarios produces unnecessary data resulting in increased
+bandwidth, CPU, and long-term storage usage. These inefficiencies lead to increased operational costs and unnecessary
+resource utilization, misaligning with the needs of lightweight or intermittently active networks. By enabling the
+network to pause event creation when no non-ancient, non-consensus transactions are pending and fully signed blocks are
```
kpachhai: Could we describe what are non-ancient, non-consensus transactions somewhere in the HIP?

HIP/hip-1238.md
```diff
@@ -0,0 +1,219 @@
...
+5. As a network participant with non-scheduled transactions, I want a Hiero network to come out of Quiescence to process
+   my non-scheduled transactions.
+
+## Specification
+
+### Requirements
+
+- When transactions stop being submitted to the network, the network should stop producing events in a timely manner.
+  This will happen after the submitted transactions end up in a fully signed block or become stale.
+- The amount of time it takes the network to stop should be close to the C2RC time of the network.
```
kpachhai: Do we wanna mention what "C2RC" is? I assume this means "Code Complete to Release Candidate"

HIP/hip-1238.md
```diff
@@ -0,0 +1,219 @@
...
+network to pause event creation when no non-ancient, non-consensus transactions are pending and fully signed blocks are
+achieved, the Quiescence feature addresses these shortcomings.
+
+Quiescence provides a cost-effective mode of operation for Hiero networks with intermittent usage.
+
+## Rationale
+
+The rationale fleshes out the specification by describing why particular design decisions were made. It should describe
+alternate designs that were considered and related work, e.g. how the feature is supported in other ecosystems. The
+rationale should provide evidence of consensus within the community and discuss important objections or concerns raised
```
kpachhai: The Rationale section seems to be a description of what the rationale is supposed to contain(possibly a leftover from the HIP template). Can we update this section?

HIP/hip-1238.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Consensus concern
+
+We considered having the consensus module keep track of all three quiescence conditions, but this was rejected because
+they are already concerns of the execution module. Pushing these concerns into the consensus module would create several
+new touchpoints between the two modules, which would complicate the design and muddle responsibilities.
+
+## Open Issues
+
+- How will mirror nodes detect quiescence?
```
kpachhai, line 205: In the same vain, how would block nodes detect quiescence? Would they use similar logic to detect this?

jsync-swirlds: Block Nodes will detect that they stop receiving blocks, and query all other known (tier 1) block nodes to see if they also stopped receiving blocks (which has to be done to prevent gaps anyway).  If all the block nodes are receiving no blocks, the block node will assume the network is quiescent and wake less often to check if other block nodes are getting blocks.

I assume Mirror Node will do something similar (also checking multiple block nodes to determine if blocks are available anywhere).

It's heuristic, however. There is no definitive "the network is now quiescent" signal.

HIP/hip-1238.md
lpetrovic05: i reworded this to make it clearer

HIP/hip-1238.md
lpetrovic05: I removed this acronym since its not widely known, it should be easier to understand now

HIP/hip-1238.md
lpetrovic05: Yup, this one slipped though the cracks, I moved part of the motivation here and added a bit

HIP/hip-1238.md
lpetrovic05: Should we add this explanation to the HIP?

jsync-swirlds: Probably not. How any given non-consensus-node system might handle quiescence seems to be a non-normative element.

PR #1242, opened 2025-07-11
chore: Fix HIP-1215 system contract return values

tinker-michaelj:
**Description**:
 - The system contracts that create a new scheduled transaction on success should return the created schedule's address.

### DIFF
--- a/HIP/hip-1215.md
+++ b/HIP/hip-1215.md
@@ -73,15 +73,16 @@ past the `expiry`, of course).
 --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 | Hash        | Function signature                                                                                                                                                   |
 |-------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| 0x6f5bfde8  | scheduleCall(address to, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (int64 responseCode)                                            |
-| 0xeb459436  | scheduleCallWithSender(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (int64 responseCode)                  |
-| 0x2c300715  | executeCallOnSenderSignature(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (int64 responseCode)         |
-| 0x07cdefc0  | hasScheduleCapacity(uint256 expirySecond, uint256 gasLimit) view returns (bool capacity) |
+| 0x6f5bfde8  | scheduleCall(address to, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address)                                                    |
+| 0xeb459436  | scheduleCallWithSender(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address)                          |
+| 0x2c300715  | executeCallOnSenderSignature(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address)                    |
+| 0x07cdefc0  | hasScheduleCapacity(uint256 expirySecond, uint256 gasLimit) view returns (bool)                                                                                      |
 --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 ```
 
-The `scheduleCall()` variants revert with `SCHEDULE_EXPIRY_IS_BUSY` is the specified
-second is saturated. The `hasScheduleCapacity()` view function never reverts, but
+The `scheduleCall()` variants revert with reason `"SCHEDULE_EXPIRY_IS_BUSY"` if the given
+second is saturated. If they succeed, they return the address of the newly created
+scheduled transaction. The `hasScheduleCapacity()` view function never reverts, but
 returns `true` iff the given second still has capacity to schedule a contract call with
 the specified gas limit. 
 


HIP/hip-1215.md
```diff
@@ -73,15 +73,16 @@ past the `expiry`, of course).
...
-| 0x2c300715  | executeCallOnSenderSignature(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (int64 responseCode)         |
-| 0x07cdefc0  | hasScheduleCapacity(uint256 expirySecond, uint256 gasLimit) view returns (bool capacity) |
+| 0x6f5bfde8  | scheduleCall(address to, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address)                                                    |
+| 0xeb459436  | scheduleCallWithSender(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address)                          |
+| 0x2c300715  | executeCallOnSenderSignature(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address)                    |
+| 0x07cdefc0  | hasScheduleCapacity(uint256 expirySecond, uint256 gasLimit) view returns (bool)                                                                                      |
```
lukelee-sl: Should the `scheduleDeleteTransaction(address)` function be included in this update?

tinker-michaelj: Added @lukelee-sl !

lukelee-sl (dismissed):
LGTM

HIP/hip-1215.md
stoyanov-st: We might consider keeping both the `int64 responseCode` and `address scheduleAddress` in the response?

stoyanov-st: Also, does it make sense to consider `scheduleDeleteWithSender(address schedule, address sender)` and/or IHRC schedule redirect `scheduleDelete()` functions?

lukelee-sl (approved):
LGTM

HIP/hip-1215.md
```diff
@@ -73,17 +73,23 @@ past the `expiry`, of course).
...
 |-------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| 0x6f5bfde8  | scheduleCall(address to, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (int64 responseCode)                                            |
-| 0xeb459436  | scheduleCallWithSender(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (int64 responseCode)                  |
-| 0x2c300715  | executeCallOnSenderSignature(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (int64 responseCode)         |
-| 0x07cdefc0  | hasScheduleCapacity(uint256 expirySecond, uint256 gasLimit) view returns (bool capacity) |
+| 0x6f5bfde8  | scheduleCall(address to, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address)                                                    |
```
gkozyryatskyy: Hello guys! I am interested why `expiry` is `uint256` everywhere? it is `seconds` so it can be `uint32`. 
Few points about it
- `uint256` can be confusing when value is in seconds
- with `uint256` instead `uint32` we should cast it to `int` inside
- with `uint256` instead `uint32` we should think about `int` overflow inside

HIP/hip-1215.md
```diff
@@ -73,17 +73,23 @@ past the `expiry`, of course).
...
+| 0xeb459436  | scheduleCallWithSender(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address)                          |
+| 0x2c300715  | executeCallOnSenderSignature(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address)                    |
+| 0x25bd252f  | scheduleDeleteTransaction(address)                                                                                                                                   |
```
gkozyryatskyy: shouldn't `scheduleDeleteTransaction(address)` return something? Like status `(int64)`?

HIP/hip-1215.md
```diff
@@ -73,17 +73,23 @@ past the `expiry`, of course).
...
-| 0x6f5bfde8  | scheduleCall(address to, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (int64 responseCode)                                            |
-| 0xeb459436  | scheduleCallWithSender(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (int64 responseCode)                  |
-| 0x2c300715  | executeCallOnSenderSignature(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (int64 responseCode)         |
-| 0x07cdefc0  | hasScheduleCapacity(uint256 expirySecond, uint256 gasLimit) view returns (bool capacity) |
+| 0x6f5bfde8  | scheduleCall(address to, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address)                                                    |
+| 0xeb459436  | scheduleCallWithSender(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address)                          |
```
gkozyryatskyy: Currently existed `scheduleNative(address,bytes,address)` returning status and address `(int64,address)`.
Mb this function also should return `(int64,address)`?

HIP/hip-1215.md
```diff
@@ -73,17 +73,23 @@ past the `expiry`, of course).
...
-returns `true` iff the given second still has capacity to schedule a contract call with
-the specified gas limit. 
+The `scheduleCall()` variants revert with reason `"SCHEDULE_EXPIRY_IS_BUSY"` if the given
+second is saturated. If they succeed, they return the address of the newly created
+scheduled transaction.
+
+This scheduled transaction may be deleted with the `scheduleDeleteTransaction()`, which
+only reverts if it fails to perform the deletion.
+
+The `hasScheduleCapacity()` view function never reverts, but returns `true` iff the given
```
gkozyryatskyy: typo `iff`

HIP/hip-1215.md
```diff
@@ -73,17 +73,23 @@ past the `expiry`, of course).
...
-| 0x07cdefc0  | hasScheduleCapacity(uint256 expirySecond, uint256 gasLimit) view returns (bool capacity) |
+| 0x6f5bfde8  | scheduleCall(address to, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address)                                                    |
+| 0xeb459436  | scheduleCallWithSender(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address)                          |
+| 0x2c300715  | executeCallOnSenderSignature(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address)                    |
+| 0x25bd252f  | scheduleDeleteTransaction(address)                                                                                                                                   |
+| 0x07cdefc0  | hasScheduleCapacity(uint256 expirySecond, uint256 gasLimit) view returns (bool)                                                                                      |
```
gkozyryatskyy: What is the `capacity`? if this is a number mb we should return it? So the user saw not just status, but the amount of capacity left? 
In this case user can plan few scheduled on one second.

HIP/hip-1215.md
lukasz-hashgraph: I don't think it's a typo, iff stands for "if and only if"

HIP/hip-1215.md
lukasz-hashgraph: As far as I understand it's a specific consensus timestamp (in seconds), not a diff.
While int32 can hold the unix timestamp (in seconds) at the moment, it would overflow in ~12 years

tinker-michaelj: It just mirrors the Solidity type of `block.timestamp`.

HIP/hip-1215.md
tinker-michaelj: Updated!

HIP/hip-1215.md
tinker-michaelj: Done!

HIP/hip-1215.md
lukasz-hashgraph: Correction: it would overflow in 80 years :) How about uint64 for a closer match to Java's `long` that seems to be used in the `ScheduleService`?

HIP/hip-1215.md
tinker-michaelj: Hmm I hesitate a little on this one, as I don't think people should really be scheduling calls with a higher `gasLimit` than they need in the first place. 🤔

HIP/hip-1215.md
tinker-michaelj: Imo the two-`address` signature is a little out of scope, but the redirect makes a ton of sense, added!

HIP/hip-1215.md
gkozyryatskyy: - `uint32` max value 4294967295
- 4294967295 / 60 / 60 / 24 / 365 = 136.19 years
- `uint64` is x2 `java long` because `java long` is `signed`, so it still will need `BigInteger`

HIP/hip-1215.md
gkozyryatskyy: not sure if this is important and do not want to sound "boring" but you changed to `(address,int64)` and existed functions uses `(int64,address)`, the order is different =)
Just mentioning.. On your choice if change to same params order is important.

stoyanov-st (dismissed):
All good on my side
tyvm @tinker-michaelj !

HIP/hip-1215.md
lukasz-hashgraph: 55 years out of those 136 are already consumed 😄

We could use uint64 but restrict it to 63 bytes with an assertion and map to `long` in java. I guess we'd need the same if we went with uint256 anyway. Using the same type as `block.timestamp` makes sense to me, for ease of use.

HIP/hip-1215.md
```diff
@@ -57,33 +57,51 @@ interest reallocation) without relying on bots.
...
+| Hash        | Function signature                                                                                                                                       |
+|-------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
+| 0x6f5bfde8  | scheduleCall(address to, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address, int64)                                 |
+| 0xeb459436  | scheduleCallWithSender(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address, int64)       |
+| 0x2c300715  | executeCallOnSenderSignature(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (address, int64) |
+| 0x5312ea4e  | deleteSchedule(address) returns (int64)                                                                                                                  |
```
lukasz-hashgraph: This doesn't seem to be the correct hash. Shouldn't it be: `0x72d42394`?

HIP/hip-1215.md
```diff
@@ -57,33 +57,51 @@ interest reallocation) without relying on bots.
...
+| 0x07cdefc0  | hasScheduleCapacity(uint256 expirySecond, uint256 gasLimit) view returns (bool)                                                                          |
```
lukasz-hashgraph: Same here, the hash is `0xdfb4a999`

HIP/hip-1215.md
tinker-michaelj: Fixed, tyvm!

HIP/hip-1215.md
tinker-michaelj: Fixed, tyvm!

HIP/hip-1215.md
tinker-michaelj: Good point, switched to match existing pattern.

gkozyryatskyy (approved):

HIP/hip-1215.md
gkozyryatskyy: Really I think any of proposed solutions (including leave uint256) should be fine. 
This seems the last unresolved discussion in this pull.
If all fine with it, I think it is up to @tinker-michaelj to take a decision on this and we can merge for not to block next work.

HIP/hip-1215.md
```diff
@@ -57,33 +57,51 @@ interest reallocation) without relying on bots.
...
+|-------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
+| 0x6f5bfde8  | scheduleCall(address to, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (int64, address)                                 |
+| 0xeb459436  | scheduleCallWithSender(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (int64, address)       |
+| 0x2c300715  | executeCallOnSenderSignature(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (int64, address) |
+| 0x72d42394  | deleteSchedule(address) returns (int64)                                                                                                                  |
+| 0xdfb4a999  | hasScheduleCapacity(uint256 expirySecond, uint256 gasLimit) view returns (bool)                                                                          |
```
lukasz-hashgraph: Minor naming inconsistency: `expiry` vs `expirySecond`.

gkozyryatskyy: +1 to use `expirySecond` everywhere

lukasz-hashgraph: I'd actually prefer `expiry` ;)  or `expiryTimestamp`. `expirySecond` could be a bit confusing and by some might be intuitively understood as a diff (expires in n seconds from now)

gkozyryatskyy: From other hand from `expirySecond` you understand the unit and from others you do not know really if is second, ms or ns (with uint256 it, easily, can be ns).

Anyway im ok with both and agree they should be in sync

HIP/hip-1215.md
```diff
@@ -57,33 +57,51 @@ interest reallocation) without relying on bots.
...
+If the `scheduleCall()` variants succeed, they return the address of the newly created
+scheduled transaction and `22` as the status code of `SUCCESS`. A contract may attempt to
+delete an already scheduled transaction with the `deleteSchedule(address)`, receiving `22`
+iff the delete succeeds. A contract or EOA may also attempt to delete a scheduled transaction
+at address `0x00...abcd` by calling a "redirect" `deleteSchedule()` function at that address.
+```
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+| Hash        | Function signature                                                                                                                                       |
+|-------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
+| 0x8309a29b  | deleteSchedule() returns (int64)                                                                                                                         |
```
gkozyryatskyy: Seems like the `Hash` for `deleteSchedule()` should be `0xc61dea85`

HIP/hip-1215.md
```diff
@@ -57,33 +57,51 @@ interest reallocation) without relying on bots.
...
+| Hash        | Function signature                                                                                                                                       |
+|-------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
+| 0x6f5bfde8  | scheduleCall(address to, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (int64, address)                                 |
+| 0xeb459436  | scheduleCallWithSender(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (int64, address)       |
+| 0x2c300715  | executeCallOnSenderSignature(address to, address sender, uint256 expiry, uint256 gasLimit, uint64 value, bytes memory callData) returns (int64, address) |
+| 0x72d42394  | deleteSchedule(address) returns (int64)                                                                                                                  |
```
gkozyryatskyy: add `scheduleAddress` as an `address` param name?

HIP/hip-1215.md
tinker-michaelj: I think `uint64` is fine.

HIP/hip-1215.md
tinker-michaelj: Added.

HIP/hip-1215.md
tinker-michaelj: Fixed, tvym!

HIP/hip-1215.md
tinker-michaelj: Standardized to `expirySecond`.

stoyanov-st (approved):

mgarbs (approved):

PR #1247, opened 2025-07-16
Update HIP-1139 to clarify the submit key update

vtronkov:
**Description**:
Updating the HIP-1139 to clarify that even if a submit key is set to `SENTINEL`, if we have an admin-key we can update it to a valid key


mgarbs (approved):

ty-swirldslabs (commented):
LGTM

mgarbs (approved):

PR #1249, opened 2025-07-25
Create HIP for Precise Smart Contract Throttling.

rbair23:
This proposal introduces a new throttling mechanism for smart contracts on the Hiero network to replace the current
gas-per-second limit. By shifting from gas-based throttling to an operations-per-second (ops/sec) model, the network can
support significantly higher throughput for smart contract execution—potentially an order of magnitude more than the
current 15 million gas/sec default—while maintaining security and compatibility with Ethereum tools. Gas will continue
to be used for user billing, but throttling will rely on a separate, measured cost for each EVM opcode, precompile, and
system contract to better reflect actual computational demands.

### DIFF
--- a/HIP/hip-xxxx.md
+++ b/HIP/hip-xxxx.md
@@ -1,0 +1,190 @@
+# Precise Smart Contract Throttling
+
+---
+hip: XXX
+title: Precise Smart Contract Throttling
+author: Richard Bair <@rbair23>
+working-group: Richard Bair <@rbair23> Jasper Potts <@jasperpotts>, Luke Lee <@lukelee-sl>
+type: Standards Track
+category: Core
+needs-hiero-review: Yes
+needs-hedera-approval: Yes
+status: Draft
+created: 2025-06-30
+last-call-date-time:
+discussions-to:
+replaces: https://hips.hedera.com/hip/hip-185
+---
+
+# Abstract
+
+This proposal introduces a new throttling mechanism for smart contracts on the Hiero network to replace the current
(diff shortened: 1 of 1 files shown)


mgarbs (approved):

HIP/hip-1249.md
```diff
@@ -0,0 +1,199 @@
...
+
+1. Remove the existing gas/sec throttle from both frontend (ingest) and backend (consensus) layers.
+2. Introduce a new ops/sec throttle exclusively on the backend (consensus) layer.
+3. Provide network-specific configuration for the ops cost of each EVM opcode and precompile, based on hardware
+   benchmarks.
+4. For each system contract, calculate and introduce the ops cost based on hardware benchmarks.
+5. During smart contract execution:
+    - Calculate gas as before for user billing.
+    - Separately track ops units for throttling.
+6. Bill users based solely on gas consumed, eliminating the prior 80% minimum charge rule (which required billing at
```
lukasz-hashgraph: "gas consumed" is a term that's being used for something else: raw computation cost, before applying any additional charges and/or refunds. It's "gas used" that we're charging for.

Perhaps we should change it to something like this to make it clear:
>  Bill users based on Ethereum-equivalent gas used, eliminating the prior 80% minimum charge rule ...

mgarbs (approved):

HIP/hip-1249.md
mgarbs: updated!

mgarbs (approved):

PR #1259, opened 2025-08-02
Submit new HIP for Fee Collection Account

rbair23:
This proposal introduces a new network-controlled account, known as the fee collection account (default address
0.0.802), into which all fees are paid. Instead of distributing fees immediately across multiple accounts, the system
accumulates them here and performs a single daily transfer at the start of each staking period. This transfer sends the
appropriate portions to the admin fee account (0.0.98), staking reward account (0.0.800), node reward account (0.0.801),
and individual node accounts. The change simplifies transaction records, reduces block stream size, and improves overall
network performance.


HIP/hip-1259.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Protobuf Changes
+
+The singleton state is defined in a new protobuf file: `services/state/node_payments.proto`.
+
+```protobuf
+// A singleton state object that accumulates node fees for distribution.
+message NodePaymentsInfo {
+  // A map from account_number to NodePaymentInfo.
+  map<int64, NodePaymentInfo> payments = 1;
```
steven-sheehy:
```suggestion
  map<uint64, NodePaymentInfo> payments = 1;
```

steven-sheehy: The info suffix seems unnecessary and makes it harder to say when plural. Could consider dropping it unless it's some standard naming for state objects.

```suggestion
message NodePayments {
  // A map from account_number to NodePaymentInfo.
  map<int64, NodePayment> payments = 1;
```

mgarbs: Applied

mgarbs (approved):

PR #1261, opened 2025-08-08
HIP-1261: Simple fees

rbair23:
This proposal introduces a simplified and transparent fee model for Hiero networks. It reduces code complexity, enhances
network performance, and improves predictability for fee calculations. Tools like SDKs and the Mirror Node Explorer can
accurately estimate fees using the same logic as consensus nodes. The method for converting USD fees to HBAR remains
unchanged.

### DIFF
--- a/HIP/hip-1195.md
+++ b/HIP/hip-1195.md
@@ -12,7 +12,7 @@ status: Approved
 last-call-date-time: 2025-06-06T07:00:00Z
 created: 2025-02-19
 discussions-to: https://github.com/hiero-ledger/hiero-improvement-proposals/discussions/1172
-updated: 2025-09-02
+updated: 2025-09-17
 ---
 
 ## Abstract
@@ -22,8 +22,8 @@ In principle, hooks could be programmed in any language, but we begin with **EVM
 writing contracts in a language like Solidity that compiles to EVM bytecode. EVM hooks may have optimized variants in
 the future, but we propose first a full-featured **lambda** EVM hook. The name is chosen to evoke event-driven code
 running in the cloud or call other external services. For any given entity, its user can create many hooks for that
-entity with different 64-bit **hook ids**. There is no limit on the number of hook ids than an entity can use; bu
-tits storage footprint, and hence rent, will increase proportionally.
+entity with different 64-bit **hook ids**. There is no limit on the number of hook ids than an entity can use; but
+its storage footprint, and hence rent, will increase proportionally.
 
 As a first Hiero extension point, we propose **account allowance hooks**. Users can create these hooks on their
 accounts, and a Hiero API (HAPI) `CryptoTransfer` transaction can then reference an allowance hook just as it does an
@@ -822,7 +822,7 @@ message NftTransfer {
 Note that `NftTransfer` supports both sender and receiver transfer allowance hooks, since the transaction may
 need to use the receiver hook to satisfy a `receiver_sig_required=true` setting.
 
-### The transfer allowance ABI
+### The account allowance ABI
 
 The account allowance EVM hook ABI is as follows,
 
@@ -925,6 +925,29 @@ interface IHieroAccountAllowancePrePostHook {
    ) external payable returns (bool);
 }
 ```
+
+### Call order of `ACCOUNT_ALLOWANCE_HOOK`s
+
+A `CryptoTransferTransactionBody` can execute multiple hooks, subject only to the global limit on the number of child
+records for a single transaction,
+```
+consensus.handle.maxFollowingRecords=50
+```
+The `CryptoTransfer` handler will execute hook calls in the following order:
+1. All `pre_tx_allowance_hook` and `pre_post_tx_allowance_hook` calls in the HBAR `TransferList`, in the order they
+appear in the `CryptoTransferTransactionBody`.
+2. All `pre_tx_allowance_hook` and `pre_post_tx_allowance_hook` calls in the fungible `transfers` of the `tokenTransfers`
+list, in the order they appear in the `CryptoTransferTransactionBody`.
+3. All `pre_tx_sender_allowance_hook`, `pre_post_sender_tx_allowance_hook`, `pre_tx_receiver_allowance_hook`, and
+`pre_post_tx_receiver_allowance_hook` calls in the non-fungible `nftTransfers` of the `tokenTransfers` list, in the
+order they appear in the `CryptoTransferTransactionBody`. (_Note:_ if both sender and receiver hooks are present, then
+the sender hook is executed first.)
+4. All `pre_post` hooks previously executed in steps (1) to (3), _in the same order they were previously executed_;
+while the first call will be to the `allowPre(HookContext, ProposedTransfers)` method signature, the second call will
+be to the `allowPost(HookContext, ProposedTransfers)` signature.
+
+A block stream or legacy record stream client can use this well-known ordering to identify which child `ContractCall`
+to `0x16d` corresponds to which `HookCall` in a `CryptoTransferTransactionBody`.
 
 ### Examples
 
--- a/_data/draft_hips.json
+++ b/_data/draft_hips.json
@@ -122,7 +122,7 @@
     "title": "HIP-1261: Simple fees",
     "number": 1261,
     "url": "https://github.com/hiero-ledger/hiero-improvement-proposals/pull/1261",
-    "headRefOid": "4f88d072bf1c0ac75bf91119ad6a0a81b31687f9",
+    "headRefOid": "e4d5c222347e39eff3703860a6396bca57baf1e7",
     "files": {
       "edges": [
         {


HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+   * defined within the `FeeSchedule`.
+   */
+  string name = 1;
+
+  /**
+   * The count of this "extra" that is included for free. For example, 256 "Bytes"
+   * may be included for free. Each byte above 256 would be charged the fee defined
+   * on the "Bytes" extra.
+   */
+  uint32 includedCount = 2;
```
jsync-swirlds: Note, for consistency this should be `included_count`.

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+
+```json
+{
+  "network": {"multiplier": 9}
+}
+```
+
+### Services
+
+A service is a grouping of transactions and queries. These match the gRPC **services defined in the HAPI specification.
```
steven-sheehy: Rendering error on services

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+### Node
+
+The *node* fee is computed exactly the same way for all transactions. It is therefore defined in one place in the
+configuration and applicable to all transactions.
+
+```json
+{
+  "node": {
+    "base": 100000,
+    "extras": [
```
steven-sheehy, line 300: surcharge is a possible alternative name that has a direct financial meaning.

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+### Schema
+
+The schema is defined in Protobuf. The wire format is JSON.
+
+```protobuf
+/**
+ * Network fee schedule definition.<br/>
+ * This message defines how fees are charged for transactions.
+ */
+message FeeSchedule {
```
steven-sheehy, line 452: Changing the protobuf format of 0.0.111 is a breaking change for anyone using the network file directly. The current protobuf structure is public and could be used by anyone. This is not discussed anywhere.

For example, our mono-based contracts/call API would stop functioning. This would require the transition to modularized to be completed.

steven-sheehy, line 452: How will we handle the change in format for mirror node /api/v1/network/fees that partially exposes the current protobuf format?

```json
{
  "fees": [
    {
      "gas": 32,
      "transaction_type": "ContractCall"
    },
    {
      "gas": 32,
      "transaction_type": "ContractCreate"
    },
    {
      "gas": 32,
      "transaction_type": "EthereumTransaction"
    }
  ],
  "timestamp": "1753290258.948867042"
}
```

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+  exist, load token associations). This is the default if no `mode` is specified.
+
+If the provided transaction is invalid (e.g., cannot be parsed), the response includes an error. For `state` mode, if
+required state is unavailable (e.g., due to pruning), the endpoint falls back to `intrinsic` and notes this in the
+response.
+
+### Endpoint Definition
+
+- **Method**: `POST`
+- **Path**: `/api/v1/fees/estimate`
```
steven-sheehy: Should it even be a REST API? The mirror node gRPC API would be a more natural source of an API that takes as input protobuf. SDKs already supports gRPC API for subscribe topic so it would be easy to consume. This also side steps the difficulties with accepting JSON formatted transactions. And it avoid double decoding the payload which would be a memory concern with jumbo transactions. If you agree, I can take a crack at it.

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+| **AddressBookService**   | UpdateNode                 | ✅              |           |          |                |                |              |                            |                               |                             |                                |                             |                     |               |         |                |              |
+
+### Mirror Nodes
+
+Mirror nodes ingest and index data from the Hiero network, providing queryable access to historical and real-time state.
+To support fee transparency, this HIP introduces a new REST API endpoint on mirror nodes for estimating transaction
+fees. This endpoint allows users, SDKs, and tools to compute expected fees without submitting transactions to the
+network.
+
+The endpoint accepts a `Transaction` (as defined in the Hiero API protobufs) in either binary protobuf or JSON format.
```
steven-sheehy: A `Transaction` proto just defines a field with `bytes signedTransactionBytes`. The `SignedTransaction` when decoded has `bytes bodyBytes` and `SignatureMap sigMap`. It seems fine to me to accept `Transaction` proto directly as bytes, but the difficulty comes if we also want to support `Transaction` as JSON due to the double encoding of binary data in that proto e.g. `{"signedTransactionBytes": "base64..."}`. This provides a poor user experience for JSON that they might as well just use binary.

If we want to workaround this, there's no protobuf format that maps to something that accepts something that isn't byte encoded. That is there's no message that contains both `TransactionBody` and `SignatureMap`. Even if we did create one, it's not clear how we could accept base64/hex for nested binary fields like `ConsensusSubmitMessageTransactionBody.message` without maintaining our own input wrappers for every transaction type. This would be burdensome and brittle.

Also a concern is the overhead of JSON over protobuf for large payloads like 6 KiB transactions and 128KiB jumbo transactions. With hex or base64 encoding and the additional bytes adding by field names, whitespace, and quotation marks it could greatly bloat the size of the payload to require us to accept up to 300 KiB of input or more.

A JSON input would be mostly be useful for ad-hoc manual user requests to the API, but REST APIs are written for programmatic use by machines. Those machines should already understand how to construct a transaction since they're already doing it elsewhere, so those could easily just use binary protobuf.

For all these reasons, I think we should keep it simple and only accept binary protobuf.

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+created: 2025-07-02
+last-call-date-time:
+discussions-to:
+requires: HIP-1259 Fee Collection Account
+---
+
+# Abstract
+
+This proposal introduces a simplified and transparent fee model for Hiero networks. It reduces code complexity, enhances
+network performance, and improves predictability for fee calculations. Tools like SDKs and the Mirror Node Explorer can
```
exploreriii, line 24: The transaction from:
This proposal introduces a simplified and transparent fee model for Hiero networks. It reduces code complexity, enhances
network performance, and improves predictability for fee calculations. 

to:
Tools like SDKs and the Mirror Node Explorer can accurately estimate fees using the same logic as consensus nodes. The method for converting USD fees to HBAR remains
unchanged.

Is ambiguous. Will the proposal allow tools like SDKs and Mirror Node Explorer to accurately estimate fees with the same logic as consensus? or do they currently - if so should be a separate paragraph?

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+
+A transaction will have a `base` plus `extras` for each additional feature. Fees are split into three components:
+
+- Node fee: Paid to the submitting node, based on bytes and signatures. Encourages nodes to accept transactions.
+- Network fee: Covers gossip, consensus, signature verifications, fee payment, and blockchain storage.
+- Service fee: Covers execution costs, state saved in the Merkle tree, and additional costs to the blockchain storage.
+
+Configuring the fee schedule is the responsibility of the deploying network (e.g. Hedera) to set. Since fee schedules
+must be deterministic across nodes, this HIP simply defines the breakdown of base fee and extras for each transaction
+type and how to calculate the `node`, `network`, and `service` components of the fee. If new transactions are added in
```
exploreriii, line 44: This is confusing - the node/network/service components of the fee and 'the fee'? All that is mentioned so far is base fee + extra fee.

maybe we can add an
##End User section
Detailing what they see - absence of node / network / service? 

## Backend section
Detailing the node / network / service breakdown, and how that is grouped into base + extras
( you clarify this later)

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+3. **Network Governance**
+    1. **As a Hiero network governing authority** (e.g., Hedera Council), I want to validate that the network applies
+       fees correctly across all services, so I can maintain trust and fairness in the ecosystem.
+4. **Node Operators**
+    1. **As a block node or mirror node operator**, I want to offer value-added services that provide visibility into
+       the fees charged to user accounts, so I can help users understand and audit their fee-related activity.
+    2. **As a node operator**, I want assurance that the node's portion of collected fees is accurately deposited to my
+       node account, so I can monitor revenue and maintain operations.
+
+# Specification
```
exploreriii, line 103: I assume this is about the proposed new changes but the confusion arises as it is written in the present tense
It would be good to have a brief section indicating how this is different from the current implementation, so the user can understand the net benefits

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+The *node* fee component describes the fee to be paid to the node that submitted the transaction to the network. This
+fee exists to compensate the node for the work it performed to pre-check the transaction before submitting it, and
+incentivizes the node to open ports to the world to accept new transactions from users. This fee is calculated in a
+consistent manner for *all* transactions. It is always based on the size of the transaction (in bytes) and the number of
+signatures. Every node has a specific number of bytes it can gossip to other nodes per second. From the perspective of
+the node, every transaction offers fair reward, so there is no incentive to prefer one transaction over another.
+
+The *network* fee component covers the cost of gossip, consensus, signature verifications, fee payment, and blockchain
+storage. This component is typically combined with the *service* fee to determine the total fee paid into the network’s
+Fee Collection Account (0.0.802).
```
exploreriii: is the node fee paid directly to the node?

does Fee Collection Account (0.0.802) exist, or is it a new HIP?

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+accurately estimate fees using the same logic as consensus nodes. The method for converting USD fees to HBAR remains
+unchanged.
+
+# Motivation
+
+The current fee schedule configuration and implementation is complex, leading to bugs and inconsistencies. In contrast,
+this HIP makes understanding fees as easy as ordering a pizza: a base fee plus extras. It makes costs clear and
+consistent.
+
+# Rationale
```
exploreriii, line 34: This section is not clear, leaving questions what is in the base or extra, but the document is later clearer.

This section does not clearly state that network, node, service fees each have their own base + extra components, which if added up create the final base + extra that the user sees

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+Unreadable transactions simply have an *unreadable fee*, there is no *node fee* component, since a bad node should not
+earn a reward for sending bad bytes!
+
+### Invalid Transactions
+
+When a node is given a transaction from a user, it checks to make sure the payer can pay for that transaction. The node
+also verifies other properties of the transaction, such as verifying the transaction doesn’t use an incompatible set of
+fields, or specify two different transaction bodies, or by submitting more transactions than they should have based on
+the throttles, or other such problems. For all cases where a due-diligence failure has occurred, the node itself is
+charged (and not the payer). For an invalid transaction, only the *network fee* is charged, since the transaction is
```
exploreriii, line 150: this contrasts with the previous statement For an invalid transaction, only the *network fee* is charged vs For all cases where a due-diligence failure has occurred, the node itself is charged
Are these the same type of invalid tranactions?

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+
+The amount a node will be charged for unreadable transactions is defined in the configuration, and is usually set to a
+punitive fee, and should *at least* be large enough to cover the cost of the maximum number of bad bytes they sent. The
+total possible number of bytes is already limited on a consensus node by the maximum network-configured limits (outside
+the scope of this HIP).
+
+Unreadable transactions simply have an *unreadable fee*, there is no *node fee* component, since a bad node should not
+earn a reward for sending bad bytes!
+
+### Invalid Transactions
```
exploreriii, line 144: what is an invalid transaction? 
I also notice some changes to the key terms used which causes confusion

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+fields, or specify two different transaction bodies, or by submitting more transactions than they should have based on
+the throttles, or other such problems. For all cases where a due-diligence failure has occurred, the node itself is
+charged (and not the payer). For an invalid transaction, only the *network fee* is charged, since the transaction is
+otherwise unhandled (no *service* component) and the node should not be rewarded for sending an invalid transaction (no
+*node* component).
+
+### Unhandled Transactions
+
+If a transaction has been received, can be parsed, and didn’t fail due-diligence checks, but is not handled, then the
+payer will be charged a *node fee* and a standard *network fee* to cover the cost of gossip, signature verification, fee
```
exploreriii, line 157: is it the same node and network fee?
no service fee?

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+
+1. **Throttled Transactions**. When a node is given a transaction from a user at ingest, the node checks whether the
+   throttle conditions are met for that transaction before submitting it to the network. But a malicious node may ignore
+   throttles and send too much data to the network (this would count as a due-diligence failure, see above). Or
+   sometimes, such as with smart contracts, throttling must happen during the *handle transaction* phase, and in this
+   case, the payer must pay for part of the transaction cost, even though the transaction wasn’t handled.
+2. **Atomic Batch Transactions**. In an Atomic Batch Transaction, it is possible that some of the transactions succeed,
+   and others fail, and others are not executed at all. For example, an Atomic Batch Transaction may have 4
+   transactions. The first two succeed, but the third fails, and the fourth is not executed at all. In this case, the
+   user will be charged the full cost of the first three transactions, but only the “unhandled transaction” cost for the
```
exploreriii, line 168: i thought there was no service fee for unhandled transaction 3? what is the 'full cost'?

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+> [!NOTE]
+> Throughout this document prices for fees are included. These prices are for **descriptive purposes only**. The 
+> actual values in a valid config are the responsibility of the network owners, such as the Hedera Council for the 
+> Hedera network. No prices in this HIP are considered authoritative.
+
+The configuration file is a JSON document in Protobuf-JSON format. The file defines:
+
+1. **Extras**: The set of extra charges which may be used for any *node*, *network*, or *service* fee.
+2. **Node**: The configuration for computing the node fee; applicable to all transactions
+3. **Network**: The configuration for computing the network fee; applicable to all transactions
```
exploreriii, line 221: refer to unhandled and bad transactions

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+4. **Services**: The set of service fee calculations for all transactions and queries, grouped by gRPC service.
+5. **Unreadable**: Used when transaction bytes are not valid Protobuf.
+
+For each *node*, *network*, and *service* fee, there is a `base` which is the fixed price in USD, along with zero or
```
exploreriii: This bit here is perfect!! should be prioritised at the top of the document

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+      {"name": "Signatures", "includedCount": 1}
+    ]
+  }
+}
+```
+
+### Network
+
+The *network* fee is also computed exactly the same way for all transactions. It is also defined in one place in the
+configuration and applicable to all transactions. It is specified as a multiplier of the node fee.
```
exploreriii, line 311: is this different to the uber surcharge multiplier?

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+💡
+
+NOTE: In this transaction we see that the base price is `499000000` tinycents, or `0.0499` USD. At runtime, for a basic
+`CryptoCreate` transaction, the total fee will be *node* + *network* + *service:*
+
+```text
+totalChargedKeys = (max(1, numKeys) - 1) * costPerKey;
+```
+
+Since `Keys` has an `includedCount` of 1, if the transaction only sets a single key on the new account, then it is not
```
exploreriii, line 354: oh, you mean the extra bill, the included count, refers to the number of instances that WOULD NOT face the extra bill?

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+```json
+{
+  "name": "CryptoService",
+  "transactions": [
+    {
+      "name": "CryptoCreate",
+      "base": 499000000,
+      "extras": [
+        {
+          "name": "Keys",
```
exploreriii, line 334: where do we learn the cost of these supplementary keys in the extras? i don't see a cost for the extra section?

exploreriii (commented):
I have left some first impression style comments (feel free to close them) as found this PR difficult to interpret.
I believe the document gets clearer as it goes on, I achieve good but not perfect understanding 
In the beginning, mostly, there is some ambiguous phrasing which causes confusion

This section in particular:
# Rationale
I can create a couple chunked PR's to illustrate some ideas - i don't think much needs to be changed but a little can help

HIP/hip-1261.md
```diff
@@ -0,0 +1,844 @@
...
+  "node": { ... },
+  "network": { ... },
+  "services": [ ... ],
+  "unreadable": { ... }
+}
+```
+
+### Extras
+
+Extras are fees *in addition to* the `base`. They are defined once in the configuration, and referenced by name
```
joshmarinacci: Since there is a small fixed number of extras (about a dozen) we should use an enum for the Extras name.  We should also reuse the existing enums for service names.

HIP/hip-1261.md
rbair23: This HIP builds on top of HIP-1259 which is the fee collection HIP and defines 0.0.802. And you are right, all fees (including node) go into 0.0.802 as defined in that HIP, but we have bookkeeping that tells us how much goes to each node when payout happens. I'll clean this up.

HIP/hip-1261.md
rbair23: OK, I will rewrite it and push an update to review.

HIP/hip-1261.md
rbair23: I'll work on the tense to make it clear what is new.

HIP/hip-1261.md
```diff
@@ -0,0 +1,845 @@
...
+- Node fee: Paid to the submitting node, based on bytes and signatures. Encourages nodes to accept transactions.
+- Network fee: Covers gossip, consensus, signature verifications, fee payment, and blockchain storage.
+- Service fee: Covers execution costs, state saved in the Merkle tree, and additional costs to the blockchain storage.
+
+Configuring the fee schedule is the responsibility of the deploying network (e.g. Hedera) to set. Since fee schedules
+must be deterministic across nodes, this HIP simply defines the breakdown of base fee and extras for each transaction
+type and how to calculate the `node`, `network`, and `service` components of the fee. If new transactions are added in
+the future, their specification will include their fee structure.
+
+Some transactions are not fully executed. These transactions are priced consistently regardless of the type of
```
SimiHunjan: Are there examples of the "Some transactions.."

HIP/hip-1261.md
```diff
@@ -0,0 +1,845 @@
...
+
+The *node* fee component describes the fee to be paid to the node that submitted the transaction to the network. This
+fee exists to compensate the node for the work it performed to pre-check the transaction before submitting it, and
+incentivizes the node to open ports to the world to accept new transactions from users. This fee is calculated in a
+consistent manner for *all* transactions. It is always based on the size of the transaction (in bytes) and the number of
+signatures. Every node has a specific number of bytes it can gossip to other nodes per second. From the perspective of
+the node, every transaction offers fair reward, so there is no incentive to prefer one transaction over another.
+
+The *network* fee component covers the cost of gossip, consensus, signature verifications, fee payment, and blockchain
+storage. This component is typically combined with the *service* fee to determine the total fee paid into the network’s
```
SimiHunjan: suggest - "blockchain storage (persisting state).."

HIP/hip-1261.md
```diff
@@ -0,0 +1,845 @@
...
+Fee Collection Account (0.0.802).
```
SimiHunjan: Do want to define the fee collection account as a term at the top in the motivation section? 

Also, if it possible to configure the fee collection account for different networks we probably don't want to define the fee collection account ID here . Maybe we can day the default fee collection account in Hiero is 802, but can be configured to something else in other networks.

Something like:

In Hiero, the default fee collection account is 0.0.802. However, this value is network-configurable, and other networks may designate a different account.

HIP/hip-1261.md
```diff
@@ -0,0 +1,845 @@
...
+
+![Transaction Fee Diagram](../assets/hip-1261/tx-fees.png)
+
+Normally a transaction fee will be made up of node, network, and service components added together. But this is not true
+for all transactions.
+
+### Unreadable Transactions
+
+If a transaction cannot be parsed, then the node that included that transaction in their event will be charged for
+sending bad bytes. Honest nodes will never send unreadable bytes. A node sending unreadable bytes may be attempting to
```
SimiHunjan: bad transaction body bytes?

HIP/hip-1261.md
```diff
@@ -0,0 +1,845 @@
...
+attack other nodes. Unreadable transactions *are not stored in state*, and the *hash* of the transaction will be stored
+in the blockstream and not the data. See
+the [Block Stream HIP-1056](https://github.com/hiero-ledger/hiero-improvement-proposals/blob/a589eabde3099ca2de384c0f2408ffe8f2fa6ba4/HIP/hip-1056.md)
+for details.
+
+The amount a node will be charged for unreadable transactions is defined in the configuration, and is usually set to a
```
SimiHunjan: Can we point to the configuration or add the name of it?

HIP/hip-1261.md
SimiHunjan: Maybe we can define the terms used in this page at the top of this section

HIP/hip-1261.md
```diff
@@ -0,0 +1,845 @@
...
+
+## Configuration
+
+Fees are configured using a JSON file and stored in file `0.0.111` on the network. All fees are defined in USD, as
+*tinycents*. 10^8 *tinycents* form a cent in USD. 10^10 tinycents are one dollar USD. By using tinycents, we provide a
+high degree of precision in pricing while avoiding floating point math which is imprecise for financial calculations.
+
+> [!NOTE]
+> Throughout this document prices for fees are included. These prices are for **descriptive purposes only**. The 
+> actual values in a valid config are the responsibility of the network owners, such as the Hedera Council for the
```
SimiHunjan: It is also true networks can choose to have no fees at all

HIP/hip-1261.md
SimiHunjan: I believe you can define the cost to whatever you want it to be for the given network

HIP/hip-1261.md
joshmarinacci: I have updated the impl protobuf to use an enum for extras and it works pretty well.  https://github.com/hiero-ledger/hiero-consensus-node/blob/dc40bb2199f9d57a799bccde26b2aff468a420ee/hapi/hedera-protobuf-java-api/src/main/proto/fees/fee_schedule.proto

HIP/hip-1261.md
joshmarinacci: The cost of the key is defined in the 'extras' section. For the extra called 'Keys' you would get how many keys are used in the transaction, subtract the included count, and if the remaining count is greater than zero multiply it by the per-key cost in the extras section.

HIP/hip-1261.md
joshmarinacci: Yes. The included count is what's included for free. Anything above that is what you have to pay for.

HIP/hip-1261.md
jsync-swirlds: The question is whether any given Hiero network might want to add to the extras for their particular network.
Can a network be deployed with new "extras" without code changes if we define them as strings?

I don't know that we can entirely predict what a specific private Hiero network might need.

joshmarinacci: No. You can’t add an extra without updating code.  The extras refer to parameters about the transaction set in code so to use a new parameter you’d have to update code to set the new parameter value. Fundamentally the extras are tied to code changes.

HIP/hip-1261.md
joshmarinacci: Services can be configured to be free if so desired.

HIP/hip-1261.md
jsync-swirlds: To clarify.  I am asking more "will this be true forever". If we make the API an enum, then we're more-or-less tied to that enum forever.  We can add values, but cannot remove any or change to something other than an enum, because an unknown array of users will be depending on our API.

If we leave it more flexible, then if we find a way to make the code more flexible in the future, we can take advantage of that flexibility at that point.

I'm just asking that we consider the trade-off and make this decision with a broader view that includes private networks and future services that may exist for some Hiero networks but will never be deployed to the Hedera instance.

joshmarinacci: I *personally* prefer the enums because it's more type safe and makes it harder to fat finger a config file, but I'll defer to you all. I can't think of when you'd want to add an extra without making a code change to go along with it.

HIP/hip-1261.md
Reccetech: @joshmarinacci Feedback on this?

Reccetech: @joshmarinacci Feedback?

HIP/hip-1261.md
Reccetech: @joshmarinacci Feedback?

HIP/hip-1261.md
Reccetech: @joshmarinacci Feedback?

HIP/hip-1261.md
Reccetech: @joshmarinacci Feedback?

HIP/hip-1261.md
jsync-swirlds: Perhaps it is worth noting, this `FeeSchedule` is not the same definition as the existing `FeeSchedule`.
The former is unchanged (but likely won't be used).
This definition is in  different package and different files.

That likely doesn't entirely resolve the concern (entities parsing file `0.0.111` will likely still experience a change of API), but it should be slightly more manageable in that the old definition will remain unchanged for older data.

joshmarinacci: We decided to use 0.0.113 for the new file, so the old file will still exist and we can test the new code without disrupting the old stuff. It also means we can use feature flags to flip between the old and new system. *Eventually* we will delete the old file and old FeeSchedule, but that wouldn't happen until the new stuff has been in production for a while.

HIP/hip-1261.md
joshmarinacci: I like that idea. @rbair23 ?

HIP/hip-1261.md
joshmarinacci: I would expect the new protobuf format to be exposed through a new api route on the mirror node. We should not remove the old one until the new one has been deployed for a while.

HIP/hip-1261.md
joshmarinacci: I don't have a personal preference.

HIP/hip-1261.md
joshmarinacci: Agreed.

HIP/hip-1261.md
joshmarinacci: This is on the mirror node side (right?) so I don't have a preference. As a client side developer I would expect to provide the transaction ID and have the API load up the transaction binary blob on it's own.

HIP/hip-1261.md
```diff
@@ -0,0 +1,847 @@
...
+  "unreadable": {"fee": 100000000000}
+}
+```
+
+Since an unreadable transaction could not even be parsed, it is not possible for it to have any extras. It therefore has
+a single fee, charged to the node that sent this unreadable transaction. It is intended to be punitive, but is up to the
+network owners to determine the appropriate price.
+
+### Queries
+
```
joshmarinacci, line 379: The updated protobuf does not distinguish between services and queries. Queries can use  ServiceFeeDefinition.free = true to indicate they are free. Do we still need to keep the separate?

HIP/hip-1261.md
joshmarinacci: We could also add a top level version field to allow for future changes.

HIP/hip-1261.md
Reccetech: Decision to keep as Enums.

HIP/hip-1261.md
joshmarinacci: Per the meeting today, we will keep the service transactions and and queries as the same type and not break them out separately in the protobuf. The prose should be updated to reflect this.

HIP/hip-1261.md
joshmarinacci: I don't understand your question here. Could you clarify @SimiHunjan ?

HIP/hip-1261.md
Reccetech: Decision to only accept Binary protobuf.

HIP/hip-1261.md
Reccetech: Decision to only do gRPC API. We should add API to SDKs.
@SimiHunjan

HIP/hip-1261.md
Reccetech: JSON RPC Relay main consumer of this API. @steven-sheehy to check with JSON RPC team on change.
Add version to fee file. 

Keep API structure as is - but interrogate new simple fee file for values.
Can fee calculator directly interrogate file instead of getting info from API? Don't need new public API.

HIP/hip-1261.md
Reccetech: Hiero networks can be free. Add one liner to HIP as user story.

HIP/hip-1261.md
Reccetech: Add content to HIP about which configuration file.

HIP/hip-1261.md
Reccetech: Add clarity to HIP on the meaning of bad bytes - if any of the parts cannot parse.

HIP/hip-1261.md
Reccetech: Reference the HIP that defines the Fee Collection Account HIP. Remove references to 802. Just call it the Fee Collection Account.

HIP/hip-1261.md
Reccetech: Improve language? There are some TX that will succeed and some will fail.

HIP/hip-1261.md
Reccetech: @joshmarinacci to update HIP with this change.

HIP/hip-1261.md
Reccetech: No change needed.

HIP/hip-1261.md
Reccetech: Added to updated draft
    2. **As a Hiero network governing authority** I want to create a Hiero network that does not have any transaction fees (i.e. free).

HIP/hip-1261.md
Reccetech: Change to "The amount a node will be charged for unreadable transactions is defined in the Fee Schedule, and is usually set to apunitive fee, and should *at least* be large enough to cover the cost of the maximum number of bad bytes they sent. The total possible number of bytes is already limited on a consensus node by the maximum network-configured limits (outside the scope of this HIP)."

HIP/hip-1261.md
Reccetech: Changed "bad bytes" to "unparsable bytes
"If a transaction cannot be parsed, then the node that included that transaction in their event will be charged for
sending unparsable bytes. Honest nodes will never send unreadable bytes. A node sending unreadable bytes may be attempting to
attack other nodes. Unreadable transactions *are not stored in state*, and the *hash* of the transaction will be stored
in the blockstream and not the data. See the [Block Stream HIP-1056](https://github.com/hiero-ledger/hiero-improvement-proposals/blob/a589eabde3099ca2de384c0f2408ffe8f2fa6ba4/HIP/hip-1056.md)
for details.

The amount a node will be charged for unreadable transactions is defined in the Fee Schedule, and is usually set to a
punitive fee, and should *at least* be large enough to cover the cost of the maximum number of bad bytes they sent. The total possible number of bytes is already limited on a consensus node by the maximum network-configured limits (outside the scope of this HIP).

Unreadable transactions simply have an *unreadable fee*, there is no *node fee* component, since a bad node should not earn a reward for sending unparsable bytes!"

HIP/hip-1261.md
Reccetech: Changed to "The *network* fee component covers the cost of gossip, consensus, signature verifications, fee payment, and blockchain storage. This component is typically combined with the *service* fee to determine the total fee paid into the network’s Fee Collection Account (see HIP-1259).

HIP/hip-1261.md
Reccetech: Changed to: "The *network* fee component covers the cost of gossip, consensus, signature verifications, fee payment, and blockchain storage (persisting state)."

HIP/hip-1261.md
Reccetech: I changed to" In addition during operation some transactions will pass and some will fail. All transactions are priced consistently regardless of the type of transaction, and if they pass or fail."
@joshmarinacci I am still not happy with this. Not sure what we are trying to say in this sentence.

HIP/hip-1261.md
Reccetech: @exploreriii Does this answer your question?

HIP/hip-1261.md
Reccetech: Network fee is computed separately.

HIP/hip-1261.md
Reccetech: We modified the document to clarify that "bad" means unparsable. 
@joshmarinacci Does it make sense to reference "unhandled" transactions in this section?

HIP/hip-1261.md
Reccetech: Changed to 
"The endpoint accepts a `Transaction` (as defined in the Hiero API protobufs) as a binary protobuf.
It returns a Binary protobuf breaking down the fee into node, network, and service components, including the base fee and any applicable extras. Fees are reported in tinycents (USD-denominated, with 10^10 tinycents equaling $1 USD) to match the configuration format."

HIP/hip-1261.md
Reccetech: @joshmarinacci Can you modify the endpoint definition section per the above comments.

HIP/hip-1261.md
Reccetech: @steven-sheehy Have you checked with the JSON-PRC team on this? If not I can help.

HIP/hip-1261.md
Reccetech: The consensus is that we will stick with "extras" for the naming. 
@joshmarinacci

HIP/hip-1261.md
joshmarinacci: I'll add a version field to the start of the protobuf. The calculator can read the file directly.  The old route will still work, but it won't report current fees once everything cuts over to the new system. Eventually we will need a new mirror node api for the new fees system.

HIP/hip-1261.md
joshmarinacci: done.

Reccetech (approved):
HIP has incorporating feedback from community and gone through additional review..

HIP/hip-1261.md
mgarbs: The sentence was updated

HIP/hip-1261.md
Reccetech: These are two types of invalid transactions.

HIP/hip-1261.md
Reccetech: The consensus nodes do the calcuations for fees. SDKs and Mirrors nodes simply consume and display those fee calculations. 
This is a bit different from EVM Gas- where the consensus node and mirror node do support gas estimations per industry best practice.

HIP/hip-1261.md
Reccetech: I have added terms section.

HIP/hip-1261.md
Reccetech: correct. Node and Network fee will be charge but no service fee as transaction was unsuccessful.

HIP/hip-1261.md
Reccetech: Failed Transaction (Transaction 3 in the example)

- It was executed by the network, but failed (e.g., semantically wrong, out of gas, inconsistent with state).
- Because the network performed the work, it still incurs full costs.
- That means the payer is charged the full cost: node fee + network fee + service fee.
- This is consistent with how we treat any bad transaction — the network must recover resources used, even if execution fails.

Unhandled Transaction (Transaction 4 in the example)

- It was never executed at all (because batch execution stopped after a prior failure).
- Since no service work was done, only the minimal costs are charged: node fee + network fee.
- There is no service fee for unhandled transactions.

**Summary**
Failed = Full cost (node + network + service)
Unhandled = Partial cost (node + network, no service)

So, the text’s “full cost of the first three transactions” means:
Transaction 1 (success) → full cost.
Transaction 2 (success) → full cost.
Transaction 3 (failure but executed) → full cost.
Transaction 4 (unhandled) → unhandled cost (no service fee).

hendrikebbers:
Why are the extras defined as a string (name of extras) and not as an enum?

joshmarinacci:
> Why are the extras defined as a string (name of extras) and not as an enum?

Extras are defined as an enum. 
https://github.com/joshmarinacci/hiero-consensus-node/blob/simple-fees/hapi/hedera-protobuf-java-api/src/main/proto/fees/fee_schedule.proto#L79

Reccetech (commented):
Approved.

PR #1263, opened 2025-08-11
HIP-1263: Add `STALE` status code

rbair23:
This HIP proposes adding a new `STALE` status code to the `ResponseCodeEnum` to indicate when a transaction has been 
included in a stale event that will never reach consensus. This allows clients to immediately detect and resubmit stale 
transactions instead of waiting for the current 3-minute timeout, enabling faster recovery and improved load testing 
capabilities.

### DIFF
--- a/HIP/hip-1263.md
+++ b/HIP/hip-1263.md
@@ -1,5 +1,5 @@
 ---
-hip: xxxx
+hip: 1263
 title: Add STALE Status Code to Transaction Receipts
 author: TBD
 type: Standards Track
--- a/HIP/hip-1263.md
+++ b/HIP/hip-1263.md
@@ -1,7 +1,7 @@
 ---
 hip: 1263
 title: Add STALE Status Code to Transaction Receipts
-author: TBD
+author: Richard Bair <@rbair23>
 type: Standards Track
 category: Service
 needs-hiero-approval: Yes
--- a/HIP/hip-1263.md
+++ b/HIP/hip-1263.md
@@ -30,7 +30,6 @@ This creates several problems:
 2. **Load Testing Limitations**: During network stress testing, stale events prevent maintaining maximum load because 
    clients cannot immediately resubmit failed transactions
 3. **Poor User Experience**: Applications cannot provide immediate feedback about stale transaction conditions
-4. **Missed Telemetry**: Networks cannot track stale event frequency for monitoring and optimization
 
 While stale events are rare on mainnet under normal conditions, they can occur during high-load scenarios or network 
 partitions, making early detection valuable for both operational monitoring and testing scenarios.
@@ -51,16 +50,6 @@ auto-resubmit, maintaining consistent semantics.
 ### Minimal Impact
 Adding a new response code is a purely additive change. Existing clients that don't handle `STALE` will continue to 
 function as before, while updated clients can take advantage of the immediate feedback.
-
-### Alternative Approaches Considered
-- **Event-level notifications**: Could notify about stale events rather than individual transactions, but this would 
-  require clients to track which transactions were in each event
-- **Async callbacks**: Could use a separate notification mechanism, but this adds complexity and doesn't integrate 
-  with existing receipt-based workflows
-- **Extended timeout fields**: Could add timing information to receipts, but this requires more complex client logic
-
-The status code approach provides the simplest integration with existing client patterns while delivering immediate 
-value.
 
 ## User Stories
 
@@ -94,7 +83,7 @@ enum ResponseCodeEnum {
      * prevented the containing event from progressing through consensus. This is distinct from
      * other failure modes as the transaction itself was not rejected for validation reasons.
      */
-    STALE = 401;  // Next available code in sequence
+    STALE = 514;  // Next available code in sequence
 }
 ```
 
@@ -106,6 +95,9 @@ When a consensus node determines that an event containing user transactions has 
 2. **Timing**: Receipts should be made available as soon as the stale condition is detected, without waiting for the 
    normal timeout period
 3. **Content**: All other receipt fields should be populated normally where applicable (transaction ID, etc.)
+
+When a transaction with a STALE status is resubmitted to the network, it transitions back to UNKNOWN status on ingest,
+or when first encountered by pre-handle.
 
 ### Stale Event Detection
 
@@ -124,12 +116,13 @@ If consensus nodes implement automatic resubmission of stale transactions in the
 - Only transactions that remain stale and are **not** auto-resubmitted should receive the `STALE` status
 - This maintains clear semantics: `STALE` means "this specific transaction instance will not execute"
 
+This HIP does not authorize automatic resubmission by consensus nodes, it only considers how this status code would be
+forwards compatible with such future behavior.
+
 ### Impact on Mirror Node
 
-Mirror nodes should:
-1. **Record STALE Status**: Include `STALE` status codes in transaction records like any other status
-2. **Indexing**: Ensure STALE transactions are properly indexed and queryable
-3. **Metrics**: Consider exposing stale transaction counts in metrics and dashboards
+Mirror nodes will not be aware of this status code directly. Stale events are included in block streams, but there are
+no transaction records or receipts in the block stream for stale transactions.
 
 ### Impact on SDK
 
@@ -192,6 +185,7 @@ A reference implementation should include:
 
 2. **Test Cases**:
    - Verification that `STALE` receipts are generated correctly
+   - Verification that stale records / receipts are not included in block streams or record streams
    - Timing tests ensuring immediate availability
    - Integration tests with SDK retry logic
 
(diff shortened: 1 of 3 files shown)


HIP/hip-1263.md
```diff
@@ -0,0 +1,245 @@
...
+    
+    /**
+     * The transaction was included in an event that has become stale and will never reach consensus.
+     * The transaction will not be executed and should be resubmitted if desired.
+     * 
+     * This status indicates the transaction was valid and properly formed, but network conditions
+     * prevented the containing event from progressing through consensus. This is distinct from
+     * other failure modes as the transaction itself was not rejected for validation reasons.
+     */
+    STALE = 401;  // Next available code in sequence
```
akdev: 514 is the next available code in sequence

HIP/hip-1263.md
```diff
@@ -0,0 +1,245 @@
...
+
+The specific criteria for determining when an event becomes "stale" are implementation-defined and may vary based on:
+- Network topology and health
+- Consensus algorithm timing constraints  
+- Node-specific timeout configurations
+
+This HIP does not mandate specific stale detection algorithms, only that when such detection occurs, the `STALE` status 
+should be used.
+
+### Future Auto-Resubmission Compatibility
```
akdev, line 112: This sounds speculative. Do we need to include it in this hip?

HIP/hip-1263.md
```diff
@@ -0,0 +1,212 @@
...
+SDKs should:
+1. **Status Handling**: Add `STALE` to status code enumerations and error handling logic
+2. **Retry Logic**: Consider implementing intelligent retry patterns that immediately resubmit stale transactions
+3. **Documentation**: Update documentation to explain stale conditions and recommended handling
+
+## Backwards Compatibility
+
+This change is fully backwards compatible:
+
+1. **Existing Clients**: Clients that don't handle the `STALE` status code will continue to function normally. They may
```
akdev, line 138: It might be worth noting that status will be set to STALE in certain situations where the status is currently set to OK. STALE will not replace or impact the behavior when other response codes are set.

HIP/hip-1263.md
```diff
@@ -0,0 +1,212 @@
...
+3. **Documentation**: Update documentation to explain stale conditions and recommended handling
+
+## Backwards Compatibility
+
+This change is fully backwards compatible:
+
+1. **Existing Clients**: Clients that don't handle the `STALE` status code will continue to function normally. They may 
+   not recognize the specific status value, but this doesn't break existing functionality.
+
+2. **Timeout Behavior**: The existing 3-minute client-side timeout behavior remains available as a fallback for
```
akdev, line 141: Should be specific and mention that response code RECEIPT_NOT_FOUND needs to be handled as a fallback?

akdev (changes requested):

HIP/hip-1263.md
```diff
@@ -0,0 +1,212 @@
...
+## Abstract
+
+This HIP proposes adding a new `STALE` status code to the `ResponseCodeEnum` to indicate when a transaction has been 
+included in a stale event that will never reach consensus. This allows clients to immediately detect and resubmit stale 
+transactions instead of waiting for the current 3-minute timeout, enabling faster recovery and improved load testing 
+capabilities.
+
+## Motivation
+
+Currently, when a transaction is included in an event that becomes "stale" (will never reach consensus), clients have
```
derektriley, line 23: Trying to think about this scenario, a client could submit a Transaction to multiple nodes, with the same TransactionID. If Node A produces a self event with the transaction that becomes stale, Node B is the first node to create a self event for the same TransactionID which is not stale, then the transaction will come to consensus. If you query for a receipt from Node A, there will be a period of time where you would see UNKNOWN -> STALE -> SUCCESS (after reaching consensus through the submission to Node B). And if another Node C, handles the transactions from Node B's event, prior to ingesting the transaction sent to it, returns DUPLICATE_TRANSACTION.

Is my understanding off here? Just thinking this through and making sure the verbiage around the STALE status code is correct, maybe it could reach consensus if the transaction was submitted to multiple nodes? Just based on how our DeduplicationCache is implemented currently `Set<TransactionID>`

HIP/hip-1263.md
```diff
@@ -0,0 +1,212 @@
...
+    
+    /**
+     * The transaction was included in an event that has become stale and will never reach consensus.
+     * The transaction will not be executed and should be resubmitted if desired.
+     * 
+     * This status indicates the transaction was valid and properly formed, but network conditions
+     * prevented the containing event from progressing through consensus. This is distinct from
+     * other failure modes as the transaction itself was not rejected for validation reasons.
+     */
+    STALE = 514;  // Next available code in sequence
```
jsync-swirlds, line 86: This might be a little brief.
I would suggest something like `TRANSACTION_IN_STALE_EVENT`.

kantorcodes:
This HIP looks like a great value add. Just some minor thoughts: 

1. I believe the auto resubmission text within this HIP should perhaps be removed until a future HIP is required to define that behavior specifically. 

2. It would be helpful to include a mermaid diagram to understand the flow of such transactions and the specific conditions that trigger them.

HIP/hip-1263.md
rbair23: I think you are right, and this is great thing to document properly. This could introduce some weirdness. At the least we should say that you can ONLY transition to STALE from UNKNOWN. In any other case it won't be stale, just a duplicate.

We could probably clean this up a bit, but it gets a little more complex. We have a list of all transactions with the same transaction ID. We could say that if the first one becomes stale, then we attempt to execute the next one in line. Only if all of them are STALE do you get STALE. And as soon as a new one is floating through pre-handle, we switch from STALE to UNKNOWN again.

That might clean it all the way up.

HIP/hip-1263.md
rbair23: The reason for doing so is to make sure we don't box ourselves in. I wanted to make sure we were forwards compatible with this idea so it is an option in the future.

HIP/hip-1263.md
rbair23: Are there such situations? What are they?

HIP/hip-1263.md
rbair23: I think it is OK to say the existing behavior is the same, but if you have some recommended sentence to use instead I'll apply it.

HIP/hip-1263.md
akdev: i am thinking of Oleg's original issue which had status set to `OK` https://github.com/hiero-ledger/hiero-consensus-node/issues/19710.

HIP/hip-1263.md
bugbytesinc: Agreed, STALE as the name of the code seems to be a bit of jargon, perhaps FAILED_TO_REACH_CONSENSUS which is from the outside, is more of what happened?  The concept of being in a stale event from a consensus implementation view is maybe not important other than it failed, but not because of precheck.

HIP/hip-1263.md
akdev: `TransactionReceiptQuery` returned `OK` but the status must have been `UNKNOWN` since the transaction never reached consensus which was not mentioned in https://github.com/hiero-ledger/hiero-consensus-node/issues/19710. I don't think my original comment is helpful anymore.

HIP/hip-1263.md
SimiHunjan: Like the more descriptive response code as well

jsync-swirlds: I agree that `FAILED_TO_REACH_CONSENSUS` may be a more appropriate name.
@rbair23, any thoughts on this change?

rbair23: On first blush I like the suggestion. I'm inclined to make this change.

rbair23: The only problem I have is that `FAILED_TO_REACH_CONSENSUS` sounds terminal. But it might not be. There are a few ways it can change from `FAILED_TO_REACH_CONSENSUS` to something like `OK`.

1. If you get `FAILED_TO_REACH_CONSENSUS` and then resubmit the transaction, the state can change
2. If you submit duplicate transactions, one could become `FAILED_TO_REACH_CONSENSUS` and then a second later the duplicate transaction comes along and actually gets handled (it is no longer technically a duplicate, it is the first to come to consensus).

Maybe this is splitting hairs and doesn't really matter. But I think it will be surprising for people? Do we have any other status codes, besides UNKNOWN, that aren't terminal?

bugbytesinc: I thought the UNKNOWN code was more of the "I don't know right now ask me later" response code for getting a receipt from the node that you just sent the TX to.  If STALE/FAILED_TO_REACH_CONSENSUS can't be the response until after the TX has expired (and failed to be populated) then thinking about it, a different code probably would not lighten load on the network.  But if it _was_ known before TX expiration (and returned a code saying so) the client could stop polling early.  

For SDKs that simultaneously post to multiple nodes, would they not just return when the first node returns an OK to start with?

To me it is a sort of failure condition in the specific context of the TX and Node.  I mean you can get a DUPLICATE_TRANSACTION "error" under the multi-node submit scenario too.  The multi-node-submit features of the SDKS already deal with this IIRC.

rbair23:
I am going to withdraw this HIP. @lbaird had a chance to review it and has a much better idea. We can eliminate the stale events almost completely, making it extraordinarily unlikely to ever happen. We have also found that even adding a status code breaks clients, so we have to invoke the compatibility process if we want to introduce a status code.

All we have to do is be careful about *when* we create an event. By tightening that up, we can prevent almost all stale events from ever occurring.

PR #1264, opened 2025-08-13
HIP-1264: Dollar-denominated values

Neurone:
**Description**:

This PR introduces the draft proposal for the _Dollar-denominated values_ HIP.

Discussion started here: https://github.com/hiero-ledger/hiero-improvement-proposals/discussions/1230

### DIFF
--- a/_data/draft_hips.json
+++ b/_data/draft_hips.json
@@ -24,14 +24,14 @@
     "title": "HIP-1313: Enable a new high volume entity creation option",
     "number": 1313,
     "url": "https://github.com/hiero-ledger/hiero-improvement-proposals/pull/1313",
-    "headRefOid": "a43a2270d80b52a3f02fad7a33e1fa4431515448",
+    "headRefOid": "f7e058123eb3557148e62c0b7d151c113365ca19",
     "files": {
       "edges": [
         {
           "node": {
             "path": "HIP/hip-1313.md",
             "changeType": "ADDED",
-            "additions": 342,
+            "additions": 341,
             "deletions": 0
           }
         }


HIP/hip-1264.md
```diff
@@ -0,0 +1,280 @@
...
+
+- Max network fees
+- ...
+
+### SDK Integration Examples
+
+```javascript
+// Transfer $1.50 with an Hedera transfer, using $0.01 as max transaction fee
+const txTransfer = new TransferTransaction()
+    .addDollarTransfer(senderAccount, -1.5)
```
kpachhai: How will the decimals be handled? For example, since you put -1.5 here, it makes sense you're referring to this as $1.50 however, since USD is being represented by a token, is the amount a floating-point number representing dollars or will developers need to specify amounts in the smallest units(cents), similar to how HBAR is handled in tinybars?

HIP/hip-1264.md
```diff
@@ -0,0 +1,280 @@
...
+- Separate denomination flags: Would complicate transaction structure and validation
+
+## User stories
+
+- As a service provider, I want to set a price in dollars for my HCS paid topics and HTS royalties so that I can maintain consistent revenue regardless of HBAR price fluctuations.
+- As a business owner, I want to create scheduled payments in dollars so that the recipient receives the agreed dollar amount even if the transaction executes weeks later when HBAR prices may have changed.
+- As a user, I want to send a specific dollar amount to someone without having to calculate HBAR conversions or worry about price changes between transaction creation and execution (e.g., long lasting scheduled transaction)
+- As a developer, I want to set network fee limits in dollars so that my applications don't fail when HBAR prices change and my preset limits become insufficient.
+- As a developer, I want to avoid conversion bugs in my application by letting the network handle HBAR-dollar calculations at consensus time with provable accuracy and exchange rate.
+- As a developer, I want to my smart contracts to manage and transfer dollar amounts without relying on 3rd party owned oracles.
```
kpachhai: When you send some hbar to a contract, you can send this via msg.value and this works out of the box for hbar obviously since it's a native token. Will there need to be a support for USD for this value passing via smart contract calls to payable functions or directly to a contract? If so, this HIP doesn't mention this anywhere so I was just curious.

HIP/hip-1264.md
```diff
@@ -0,0 +1,280 @@
...
+- Transaction replay: Dollar amounts are converted at consensus time, preventing value manipulation
+- Validation: Network validates that only one denomination type is used per field
+
+The proposal actually improves security by:
+
+- Eliminating client-side conversion errors and bugs
+- Providing consensus-level validation of exchange calculations
+- Reducing fraud potential from manipulated conversion rates
+
+## How to Teach This
```
kpachhai: I think we should also add in this section the fact that it will be important for wallets and applications to clearly communicate that users are not holding "Hedera Dollars". They are still holding HBAR. This feature is a "dollar-denominated intent" that gets converted to an HBAR transfer at the last second. So, users need to understand their underlying asset is still volatile, even if their transactions can be priced in stable dollars.

HIP/hip-1264.md
Neurone: I left the precision handling still in the [open topics list](https://github.com/hiero-ledger/hiero-improvement-proposals/blob/d3821536f98b45aa507a2a4fd257dd12ba4aaea4/HIP/hip-1264.md#open-issues) (_**Precision Handling**: Specification of decimal precision and rounding rules for dollar amounts_). I don't have a strict preference but I would go with 8 decimals like HBAR, unless there are other proposals.

Neurone: Technically developers can use the SDK as an helper for dollar's fractions, but under the hood the amount would be represented with its minimum amount possible (i.e., tinyUSD).

Neurone: I updated the spec mentioning tinyUSD as a starting point for the discussion. 100,000,000 tinyUSD = 1 USD.

HIP/hip-1264.md
Neurone: For the sake of compatibility, I would leave `msg.value` as it is. The system will always send the exact amount of HBAR defined in the transaction or the smart contract's code.

One possible evolution of the HIP would be to allow using the token ID from the HSCS side, enabling it to be transferred like a normal ERC20 via HTS system contract. However, I'm skeptical about including it from the start because of the potential for misunderstandings (e.g., you cannot query the balance or allow the contract, etc). For this reason, I would prefer to not support it at this time and to keep this HIP as simple as possible.

Do you think excluding this feature would be a critical miss for the HIP?

HIP/hip-1264.md
Neurone: Yes, definitely this does not impact any user or contract balance. We didn't mention any change to balance handling in this HIP, but for the sake of clarity I'll include your notes; thanks for that.

HIP/hip-1264.md
kpachhai: Thanks for clarifying the scope.

​To answer your question, no, I don't believe excluding this deep smart contract integration is a critical miss for this HIP. 

The value proposition for dollar-denominated transfers, custom fees, and scheduled transactions is already strong on its own and addresses the most common pain points developers today face.

​The more complex smart contract interaction can be reserved for a future HIP(if needed) once the community has adopted this core feature.

PR #1267, opened 2025-08-18
fix: revise HIP response code and mirror node consideration

lukelee-sl:
**Description**:
Revise HIP with a couple of update to enhance details
- update the returned response code to `CONSENSUS_GAS_EXHAUSTED`
- add a section describing mirror node impact - name deprecation of `gasConsumed`
- add description describing behavior on throttle breach


HIP/hip-1249.md
```diff
@@ -153,6 +153,10 @@ operations not covered by other throttles.
 
 If a precompile/system call exhausts ops mid-transaction, handle as in Execution Outcomes.
 
+## Mirror Node Considerations
+The mirror node computed a value called `gasConsumed` in order to provide a more accurate representation of gas used by
+the evm.  As the `gasUsed` value will accurately reflect the gas used by the evm, the `gasConsumed` value will be deprecated.
```
rbair23: What do we mean by deprecated? Will it be removed? Maybe we should just say its value will match gasUsed and should be ignored going forward? Removing it could break people. Sometimes people interpret deprecated as "don't use this" and sometime as "removed". I think we mean deprecated in the former sense. To remove confusion maybe we say deprecated but not removed for compatibility sake?

HIP/hip-1249.md
```diff
@@ -189,8 +193,9 @@ Educate developers on key changes:
 
 1. **Simplified Billing**: No more 80% minimum charge—pay only for gas used. This reduces costs for overestimated limits
    and eases gas estimation.
-2. **Throttle Failures**: Transactions may still fail due to exhausted ops/sec (error: THROTTLED_AT_CONSENSUS), even if
-   valid. Advise retry logic and monitoring network load.
+2. **Throttle Failures**: A transaction may overfill the ops per second throttle up to a 15M gas per transaction limit.  
+   Subsequent transactions will fail (error: CONSENSUS_GAS_EXHAUSTED) until enough time passes to refill the throttle capacity,
```
rbair23: Not until we refill all the throttle capacity, just until there is *some* throttle capacity.

rbair23 (changes requested):

HIP/hip-1249.md
lukelee-sl: Good point.  I'll edit.

HIP/hip-1249.md
lukelee-sl: Yes, I meant not removed but ignored..  The field will not be removed from the mirror node end-point, but should be removed from hash scan.  Currently, `gasConsumed` will not always match `gasUsed` because there is some gas that (like storage costs during deployment) which are missing from  `gasConsumed`.

I'll make an edit to say that going forward, `gasConsumed` will equal `gasUsed`

HIP/hip-1249.md
```diff
@@ -153,6 +153,11 @@ operations not covered by other throttles.
 
 If a precompile/system call exhausts ops mid-transaction, handle as in Execution Outcomes.
 
+## Mirror Node Considerations
```
IvanKavaldzhiev, line 156: There can be an empty line below the heading, so that it's better separated as the rest of the paragraphs.

HIP/hip-1249.md
mgarbs: Is this resolved @lukelee-sl ?

mgarbs (approved):

PR #1272, opened 2025-08-24
feat: add entity hip

kantorcodes:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-1270.md
+++ b/HIP/hip-1270.md
@@ -1,0 +1,78 @@
+---
+hip: 1270
+title: Unified Mirror Node entity endpoint for shared entity metadata
+author: Michael Kantor (@kantorcodes)
+requested-by: Hashgraph Online
+type: Standards Track
+category: Mirror
+status: Draft
+discussions-to:
+created: 2025-08-24
+updated: 2025-08-24
+---
+
+## Abstract:
+
+Introduce a new Mirror Node API endpoint, `GET /api/v1/entities/{id}`, that resolves any Hedera entity identifier (shard.realm.num or alias) and returns a compact, normalized view with the entity type and a minimal set of common, cross-entity fields. This reduces ambiguity when a client only has an identifier and needs to determine the entity type and basic metadata without calling multiple type-specific endpoints.
+
+## Motivation:
+
+Today, clients must guess an entity's type and probe multiple endpoints (e.g., `/accounts/{id}`, `/contracts/{id}`, `/tokens/{id}`, `/topics/{id}`, etc.). This is cumbersome when:
+
+- A user pastes an identifier of unknown type
+- The identifier is an alias that could map to an account
+- A workflow needs only shared metadata (creation time, deletion status, ledger, memo) before navigating to type-specific details
+
+A single resolver endpoint improves developer experience, reduces round trips, and establishes a stable contract for cross-entity discovery that aligns with how Mirror Node already models entities internally. Including this endpoint should reduce egress traffic and improve performance.
+
+## Rationale:
+
+- A unified resolver reflects the Mirror Node's normalized `entity` model.
+- Returning a minimal, common shape keeps the response stable over time and avoids leaking type-specific concerns.
+- Supporting multiple identifier forms (0.0.x, alias) matches existing Mirror Node resolution behavior and avoids fragmentation.
+
+## User stories
+
+- As a wallet, I want to paste any identifier and immediately know if it is an account, contract, token, topic, file, or schedule so I can route users appropriately.
+- As an indexer, I want a single API to normalize identifiers to `0.0.x` and detect the entity type before fetching type-specific details.
+- As a developer, I want to quickly check basic entity metadata (created timestamp, deletion status, memo) without guessing which endpoint to query.
+
+## Specification:
+
+The REST API will be extended with a unified resolver endpoint that accepts any Hedera entity identifier and returns shared entity metadata.
+
+### Route:
+
+/api/v1/entities/{id}
+
+### Proposed Response Structure:
+
+```json
+{
+  "entity_id": "0.0.1234",
+  "type": "ACCOUNT|CONTRACT|TOKEN|TOPIC|FILE|SCHEDULE|NODE",
+  "memo": "optional string",
+  "deleted": false,
+  "created_timestamp": "1695854412.123456789",
+  "expiry_timestamp": "1727390412.000000000",
+  "auto_renew_period": 7776000,
+  "auto_renew_account": "0.0.5678",
+  "alias": "optional alias string"
+}
+```
+
+### Existence Checks
+
+Consider `HEAD /api/v1/entities/{id}` for existence checks without a body.
+
+## Backwards Compatibility
+
+Additive. No existing endpoints change. Clients can continue using type-specific endpoints directly.
+
+## Security Implications
+
+None beyond existing Mirror Node data access. The endpoint only aggregates already-public metadata. Rate limits and existing auth (if any) apply.
+
+## Copyright/license
+
+This document is licensed under the Apache License, Version 2.0 — see [LICENSE](../LICENSE) or <https://www.apache.org/licenses/LICENSE-2.0>.


HIP/hip-1272.md
```diff
@@ -0,0 +1,78 @@
...
+### Route:
+
+/api/v1/entities/{id}
+
+### Proposed Response Structure:
+
+```json
+{
+  "entity_id": "0.0.1234",
+  "type": "ACCOUNT|CONTRACT|TOKEN|TOPIC|FILE|SCHEDULE|NODE",
```
hendrikebbers: Today the Mirror Node does not have endpoints for all the mentioned types (like File). Should we support all directly or concentrate on the types that are supported today?

HIP/hip-1272.md
```diff
@@ -0,0 +1,78 @@
...
+  "memo": "optional string",
+  "deleted": false,
+  "created_timestamp": "1695854412.123456789",
+  "expiry_timestamp": "1727390412.000000000",
+  "auto_renew_period": 7776000,
+  "auto_renew_account": "0.0.5678",
```
hendrikebbers: Not all types have all attributes. A FIle does not have an auto_renew_account (see https://docs.hedera.com/hedera/sdks-and-apis/sdks/file-service/create-a-file)

HIP/hip-1272.md
```diff
@@ -0,0 +1,78 @@
...
+  "alias": "optional alias string"
```
hendrikebbers: Instead of having all the attributes I would like to minimize the API and provide something like this:

{
  "id": "0.0.1234",
  "type": "ACCOUNT|CONTRACT|TOKEN|TOPIC|FILE|SCHEDULE|NODE",
   "link": "/contracts/0.0.1234"
}

While this does not include all details of the entity it provides a link to fetch additional details.

HIP/hip-1272.md
```diff
@@ -0,0 +1,78 @@
...
+}
+```
+
+### Existence Checks
+
+Consider `HEAD /api/v1/entities/{id}` for existence checks without a body.
```
hendrikebbers: We should remove this for now since Mirror Node does it for no endpoint at all. If we want body-less checks for Mirror Node that should be specified in a sperate HIP

HIP/hip-1272.md
```diff
@@ -0,0 +1,78 @@
...
+category: Mirror
+status: Draft
+discussions-to:
+created: 2025-08-24
+updated: 2025-08-24
+---
+
+## Abstract:
+
+Introduce a new Mirror Node API endpoint, `GET /api/v1/entities/{id}`, that resolves any Hedera entity identifier (shard.realm.num or alias) and returns a compact, normalized view with the entity type and a minimal set of common, cross-entity fields. This reduces ambiguity when a client only has an identifier and needs to determine the entity type and basic metadata without calling multiple type-specific endpoints.
```
hendrikebbers: For discussion: is "entity" the best name?

HIP/hip-1272.md
```diff
@@ -0,0 +1,78 @@
...
+- The identifier is an alias that could map to an account
+- A workflow needs only shared metadata (creation time, deletion status, ledger, memo) before navigating to type-specific details
+
+A single resolver endpoint improves developer experience, reduces round trips, and establishes a stable contract for cross-entity discovery that aligns with how Mirror Node already models entities internally. Including this endpoint should reduce egress traffic and improve performance.
+
+## Rationale:
+
+- A unified resolver reflects the Mirror Node's normalized `entity` model.
+- Returning a minimal, common shape keeps the response stable over time and avoids leaking type-specific concerns.
+- Supporting multiple identifier forms (0.0.x, alias) matches existing Mirror Node resolution behavior and avoids fragmentation.
```
hendrikebbers: Should EVM-compatible addresses for contracts (and accounts?) be added?

HIP/hip-1272.md
```diff
@@ -0,0 +1,78 @@
...
+
+The REST API will be extended with a unified resolver endpoint that accepts any Hedera entity identifier and returns shared entity metadata.
+
+### Route:
+
+/api/v1/entities/{id}
+
+### Proposed Response Structure:
+
+```json
```
hendrikebbers, line 72: This is only for 200. Pls add response for a non existing entity (bad request)

hendrikebbers (commented):
I like the idea in general. I added some thoughts about limit the scope of the HIP and make the API more solid for the future (and any data type)

HIP/hip-1272.md
bugbytesinc: That is what the APIs use to name the HAPI ID, even the .NET SDK has moved to it.

HIP/hip-1272.md
bugbytesinc: Likely Yes, it could become the second most common reason to use this endpoint.  Since going forward you must use the non long zero address if it exists for EVM interaction, everyone will need to query at least one of the mirror endpoint to get what the EVM recognizes as the proper EVM address for an account or contract.

HIP/hip-1272.md
bugbytesinc: Consider {id} also be an alias, most of the mirror endpoints support both Entity id and Alias (EVM address)?

steven-sheehy: Just to be clear, the accounts endpoint supports an entity ID, an EVM address, or an alias. Our alias is a base32 encoded form of the public key and only applies to accounts. 

For this reason, we should not accept alias as input since if you have that you already know it's an account. I think what you're looking for is the EVM address since every entity type can be converted into an EVM address. So think the wording around alias should be changed to EVM address.

kantorcodes: > Consider {id} also be an alias, most of the mirror endpoints support both Entity id and Alias (EVM address)?

Great suggestion! Added :)

HIP/hip-1272.md
kantorcodes: Done!

HIP/hip-1272.md
kantorcodes: Great suggestion, added

HIP/hip-1272.md
kantorcodes: Correct, `Entity` is what is used in SDKs, but happy to rename to something else

HIP/hip-1272.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Motivation:
+
+Today, clients must guess an entity's type and probe multiple endpoints (e.g., `/accounts/{id}`, `/contracts/{id}`, `/tokens/{id}`, `/topics/{id}`, etc.). This is cumbersome when:
+
+- A user pastes an identifier of unknown type
+- The identifier is an alias that could map to an account
+- A workflow needs only shared metadata (creation time, deletion status, ledger, memo) before navigating to type-specific details
+
+A single resolver endpoint improves developer experience, reduces round trips, and establishes a stable contract for cross-entity discovery that aligns with how Mirror Node already models entities internally. Including this endpoint should reduce egress traffic and improve performance.
```
steven-sheehy, line 26: The motivation and rationale make sense and I agree with the intent. However, we should explore possible other alternatives that meet these goals while reducing the API surface and not exposing our database internals. Exposing an entity endpoint exposes our internal database structure in regards to how we store the common details in a single table. In the future, we may want to refactor our db to store each entity type isolated in a separate table similar to how consensus nodes store them in separate key/value stores. At that point, it may become computationally expensive to support such an endpoint. And since it's not a minimal response of just ID -> type it would require a lot of memory to cache.

Also, not all consensus node state objects are entities generated from the same incrementing identifier and stored in the entity table. An example of this is the [Node](https://github.com/hiero-ledger/hiero-consensus-node/blob/main/hapi/hedera-protobuf-java-api/src/main/proto/services/state/addressbook/node.proto) object that is manipulated via `NodeCreate/Update/Delete` transactions. It is a top-level object but uses a different sequence ID. It's possible there may be others in the future. It might be confusing to users to pass a node ID to the entities endpoint and it return an error.

Possible alternatives include:
* Return a link or error message to the correct endpoint in the error response of the wrong endpoint
* Redirect calls to the correct endpoint via a 3xx response

Assuming the user is not satisfied with the minimal response from `/api/v1/entities/{id}`, they would always need at least two calls to capture the full type-specific details with the current approach in the HIP (e.g. `/api/v1/entities/0.0.1000` -> `/api/v1/accounts/0.0.1000`). Either of my above options would allow automatic discovery of the correct type and URL within one call and the full details within two (e.g. `/api/v1/contracts/0.0.1000` -> `/api/v1/accounts/0.0.1000`). While it may be a little weird that you have to guess at the initial endpoint, if you guess correctly it will actually be faster than `/api/v1/entities`. So this may also meet your requirements.

HIP/hip-1272.md
```diff
@@ -0,0 +1,112 @@
...
+
+`GET /api/v1/entities/{id}`
+
+Note: The term "entities" aligns with existing Mirror Node APIs and HAPI nomenclature for Hedera identifiers.
+
+### Parameters:
+
+- `{id}`: The entity identifier, which can be:
+  - Hedera entity ID format: `shard.realm.num` (e.g., `0.0.1234`)
+  - Alias (base32 encoded)
```
steven-sheehy: We should remove alias here as it's not applicable to all types.

HIP/hip-1272.md
```diff
@@ -0,0 +1,112 @@
...
+  - EVM address (0x-prefixed hex string for accounts and contracts)
+
+### Supported Entity Types:
+
+The endpoint will initially support entity types that have existing Mirror Node endpoints:
+- `ACCOUNT`
+- `CONTRACT` 
+- `TOKEN`
+- `TOPIC`
```
steven-sheehy, line 63: We also have SCHEDULE which has a `/api/v1/schedules/{id}`

HIP/hip-1272.md
```diff
@@ -0,0 +1,112 @@
...
+
+Future support may be added for `FILE`, `SCHEDULE`, and `NODE` as Mirror Node endpoints become available.
```
steven-sheehy: NODE will never be added as it's not an entity. See other comment.

HIP/hip-1272.md
```diff
@@ -0,0 +1,112 @@
...
+- `memo`: The entity's memo field (may be empty string)
+- `evm_address`: (Optional) The EVM address for accounts and contracts
+- `link`: Path to the type-specific endpoint for full entity details
+
+#### Error Response (404 Not Found):
+
+```json
+{
+  "error": "Entity not found",
+  "id": "0.0.9999"
```
steven-sheehy: This is not our standard error response structure. It should look like

```json
{
  "_status": {
    "messages": [
      {
        "message": "Not found"
      }
    ]
  }
}
```

HIP/hip-1272.md
kantorcodes: > The motivation and rationale make sense and I agree with the intent. However, we should explore possible other alternatives that meet these goals while reducing the API surface and not exposing our database internals. Exposing an entity endpoint exposes our internal database structure in regards to how we store the common details in a single table. In the future, we may want to refactor our db to store each entity type isolated in a separate table similar to how consensus nodes store them in separate key/value stores. At that point, it may become computationally expensive to support such an endpoint. And since it's not a minimal response of just ID -> type it would require a lot of memory to cache.
> 
> Also, not all consensus node state objects are entities generated from the same incrementing identifier and stored in the entity table. An example of this is the [Node](https://github.com/hiero-ledger/hiero-consensus-node/blob/main/hapi/hedera-protobuf-java-api/src/main/proto/services/state/addressbook/node.proto) object that is manipulated via `NodeCreate/Update/Delete` transactions. It is a top-level object but uses a different sequence ID. It's possible there may be others in the future. It might be confusing to users to pass a node ID to the entities endpoint and it return an error.
> 
> Possible alternatives include:
> 
> * Return a link or error message to the correct endpoint in the error response of the wrong endpoint
> * Redirect calls to the correct endpoint via a 3xx response
> 
> Assuming the user is not satisfied with the minimal response from `/api/v1/entities/{id}`, they would always need at least two calls to capture the full type-specific details with the current approach in the HIP (e.g. `/api/v1/entities/0.0.1000` -> `/api/v1/accounts/0.0.1000`). Either of my above options would allow automatic discovery of the correct type and URL within one call and the full details within two (e.g. `/api/v1/contracts/0.0.1000` -> `/api/v1/accounts/0.0.1000`). While it may be a little weird that you have to guess at the initial endpoint, if you guess correctly it will actually be faster than `/api/v1/entities`. So this may also meet your requirements.

I understand the concerns you've outlined. For possible alternatives, I'm not sure these would reduce the risks though:

1. Return a link or error message to the correct endpoint in the error response of the wrong endpoint: 
   
In this case, the Mirror Node Endpoint would still need to figure out the unrelated entity thus adding db load, but with a more confusing experience for the user, developer tooling and agents. Since it is a 400+ status code error, it'd require unsafe casting of the response for clients too. 

2.  Redirect calls to the correct endpoint via a 3xx response

This is _slightly_ better, but now the risk is... I request an account because I think its an account. I'm redirected to a contract. My code is written to parse an account and fails. It's not deterministic unless the user specifically adds a query parameter like `allowRedirects` when making these calls. Whereas, a dedicated entity endpoint has a specific order of operations behind it. 

Another side point, providers like HGraph expose the `entities` in their services today. Couldn't we do this today for the rest API? 

Image

HIP/hip-1272.md
kantorcodes: Thanks, updated!

HIP/hip-1272.md
bugbytesinc: In practice systems are calling multiple endpoints as it is right now.  If `address` returns, an explorer or other system still needs to call `contract` to determine if the account is a contract account or other account, and sometimes get a feel for if it has been instantiated yet.   Yes, for `scheduled` transactions and `files`, you might already have an idea of what you're looking for, but if those endpoints 404, then you might be calling `address` and `contract` to double check and inform the error you return to the original requestor.  The question as I see it, does `entities` reduce this 20 Questions game?

kantorcodes: > In practice systems are calling multiple endpoints as it is right now. If `address` returns, an explorer or other system still needs to call `contract` to determine if the account is a contract account or other account, and sometimes get a feel for if it has been instantiated yet. Yes, for `scheduled` transactions and `files`, you might already have an idea of what you're looking for, but if those endpoints 404, then you might be calling `address` and `contract` to double check and inform the error you return to the original requestor. The question as I see it, does `entities` reduce this 20 Questions game?

These are good points. There will definitely be cases where you'll need more information. I guess this why a GraphQL interface like Hgraph is so nice. 

There are also some cases like in the context of NLP in Agents. A user might provide an unclear prompt for an entity, and this endpoint would help the Agent determine what they are referring to.

kantorcodes:
@steven-sheehy Just circling back to see if you had any further feedback / ideas

HIP/hip-1272.md
steven-sheehy: I'm looking at the feasibility of another alternative that would probably make everyone happy if it works. Let me research further and get back to you.

PR #1273, opened 2025-08-27
Fix EOF syntax error in draft HIPs workflow

mgarbs:
Removes duplicate EOF that was causing ReferenceError in fetch-draft-hips.js script.

### DIFF
--- a/.github/workflows/update-draft-hips.yml
+++ b/.github/workflows/update-draft-hips.yml
@@ -152,7 +152,6 @@ jobs:
             console.error('Failed to fetch and filter PRs:', error);
             process.exit(1);
           });
-          EOF
 
       - name: Run Script
         run: node fetch-draft-hips.js


.github/workflows/update-draft-hips.yml
```diff
@@ -152,7 +152,6 @@ jobs:
             console.error('Failed to fetch and filter PRs:', error);
             process.exit(1);
           });
-          EOF
```
PavelSBorisov: @mgarbs The `EOF` is necessary when using hederoc and should not be removed as that can introduce unstable behavior. 
Judging by the errors in [recent runs](https://github.com/hiero-ledger/hiero-improvement-proposals/actions/runs/17287353572/job/49066960179):
```
/home/runner/_work/_temp/81a06b92-69d7-4f95-a500-e8c5de4d02ea.sh: line 109: warning: here-document at line 2 delimited by end-of-file (wanted `'EOF'')
```
The heredoc isn't able to catch the EOF line properly as it's expecting some deformed variant.
My best guess is that this would be fixed by keeping the EOF line and correcting line 48 
from:
`cat << \'EOF\' > fetch-draft-hips.js`
to:
`cat << 'EOF' > fetch-draft-hips.js`

In any case, if we can't utilize heredoc properly (keeping EOF at the end to avoid unstable behavior), it might be better to look into keeping a dedicated script in the repo.

PavelSBorisov (approved):

PR #1285, opened 2025-09-15
Update 1139 after last call discusssions

ty-swirldslabs:
Update HIP from discussion after last call

**Description**:
Updating with bigger clarifying table of functionality, and define functionality more specifically / implications to SDKs. 

**Notes for reviewer**:
Confirm the tables are correct with functionality / impacts.


mgarbs:
This commit needs to be verified and have dco sign off in order to pass the auto checks :|

PR #1292, opened 2025-09-23
Create hip-0000-hiero-anoncreds-method.md

Reccetech:
New informational HIP for Hiero Anoncreds Method


HIP/hip-0000-hiero-anoncreds-method.md
```diff
@@ -0,0 +1,95 @@
...
+## High-Level Summary
+
+The Hiero AnonCreds Method defines how AnonCreds objects are anchored, resolved, and maintained on the Hiero network:
+
+- **Schemas**: JSON schema definitions are stored using Hiero Consensus Service (HCS), aligned with HCS-1 file storage conventions.  
+- **Credential Definitions**: Published as HCS objects containing cryptographic material for credential issuance and verification.  
+- **Revocation Registry Definitions**: Anchored on HCS for long-term persistence.  
+- **Revocation Registry Entries**: Updated dynamically on-chain, providing real-time revocation status.  
+- **Resolution**: Performed through Hiero Mirror Node APIs, allowing verifiers and wallets to retrieve the latest state.  
+- **Identifiers**: Objects are referenced using DID-compatible identifiers (e.g., `did:hedera:<network>:<topic>#<objectType>/<objectId>`).
```
AlexanderShenshin: Regarding identifier example, I think we can use exact format defined in spec:
`<issuer-did-on-hedera>/anoncreds/<anoncreds-version>/<object-type>/<hcs-topic-id>`

If we want to include Hedera DID syntax and related details (so it's clear where network is defined, for example), we'll have something like this:
`did:hedera:<network>:<did-public-key>_<did-topic-id>/anoncreds/<anoncreds-version>/<object-type>/<object-topic-id>`

Reccetech: Change made.

Reccetech:
@mgarbs Can we move this HIP to review. Thanks!

mgarbs (approved):

PR #1299, opened 2025-09-29
HIP-1299 `account_id` updates for Address Book Node Entries.

jsync-swirlds:
### Description
This proposal refines the rules for managing the account ID associated with
Node entries in the Dynamic Address Book (DAB). The DAB is a feature that
stores information about network nodes, such as their identifiers and associated
accounts, directly in the network's state instead of in Hedera File Service (
HFS) files. This makes updates more secure and consistent.

The refinements focus on how account IDs can be updated, removed, or set, with
an emphasis on security through required signatures, preventing reuse of the
same account ID across multiple nodes, and handling cases where accounts lack
sufficient funds. These changes provide flexibility for node operators while
protecting the network from disruptions.

### DIFF
--- a/HIP/hip-1299.md
+++ b/HIP/hip-1299.md
@@ -6,10 +6,11 @@ type: Standards Track
 category: Service
 needs-hiero-approval: Yes
 needs-hedera-review: Yes
-status: Draft
+status: Last Call
+last-call-date-time: 2025-10-27T07:00:00Z
 created: 2025-09-29
-updated: 2025-10-07
-requires: HIP-869
+updated: 2025-10-13
+requires: 869
 ---
 
 ## Abstract


steven-sheehy (dismissed):
LGTM

HIP/hip-1299.md
```diff
@@ -0,0 +1,207 @@
...
+
+No changes are made to NodeDeleteTransactionBody, which removes the entire node
+entry.
+
+### Implementation Details
+- **Validation**: For setting a new account ID, the system checks that the
+  account exists and the transaction is signed by its key. If the node admin key
+  isn't signing, it also needs the current account's signature. For removal (
+  setting to `0.0.0`), it's signed by either the current account or the node
+  admin. Repeating the same account ID is allowed (idempotent). Invalid accounts
```
ibankov, line 110: So currently we don't allow NodeUpdate transactions without signing with the admin key. Do we have to allow just `nodeAccountID` update without adminKey signature and just with the old account id key. Its not really clear from this description.

jsync-swirlds: Yes, we will need to permit this specific case when implementing this PR.
That means that in checking signatures, if the account ID is to be changed, then the signature requirements for a `nodeUpdate` transaction may become a bit more complex.

- If account ID is changed
   - If admin_key signs, the existing account ID does _not_ need to sign.
   - If admin_key does _not_ sign, then the existing account ID _must_ sign.
   - If the _new_ account ID is _not_ `0.0.0`
      - the _new_ account ID _must_ sign.

If _any_ other fields change, then `admin_key` _must_ sign.

HIP/hip-1299.md
```diff
@@ -0,0 +1,198 @@
...
+  are rejected. The system ensures no other node uses the same account ID.
+- **Effective Timing**: Changes apply immediately.
+- **Roster Construction**: Nodes without a valid account ID when a new roster is
```
rbair23:
```suggestion
- **Roster Construction**: Nodes without a valid account ID or with insufficient funds when a new roster is
```

rbair23 (approved):

HIP/hip-1299.md
jsync-swirlds: Added.

HIP/hip-1299.md
```diff
@@ -0,0 +1,199 @@
...
+- **Effective Timing**: Changes apply immediately.
+- **Roster Construction**: Nodes without a valid account ID, or with
+  insufficient funds, when a new roster is adopted are excluded from that new
+  roster.
+- **Reward Distribution**: Rewards go to the active account ID. If it fails,
+  they go to `0.0.801`. Receiver signatures are ignored, and no hooks run.
+- **Penalties**: Penalties deduct from the active account ID. If balance is too
+  low to pay current penalties, the node is removed from the roster at the next
+  roster update, and may be immediately prevented from submitting new consensus
+  events.
```
mhess-swl, line 121: When a node account's balance is too low to pay the penalties, typically it's still treated as a due diligence failure, and the node pays what it can of the penalty (the account is drained of its remaining balance). Is that also the intended effect here?

jsync-swirlds: Yes, what can be paid will be, and if the balance is insufficient then that node will be prevented from submitting new consensus events for some period of time.

PR #1302, opened 2025-10-03
Update HIP-1195 with latest contract changes

Neeharika-Sompalli:
Update HIP-1195 with latest contract changes. Also added a condition that account cannot be deleted with non-zero hooks


tinker-michaelj (approved):
LGTM, tyvm @Neeharika-Sompalli !

mgarbs (approved):

PR #1331, opened 2025-11-07
 An `ExtendedAssetChanges` extra for large atomic swaps

tinker-michaelj:
**Description**:
 - Improve network usability and revenue potential by changing the `CryptoTransfer` list size limits from hard caps to overage pricing cutoffs.
 - Keep full security by refining `CryptoTransfer` throttle resolution and setting a price for the `ExtendedAssetChanges` extra that makes it _strictly more expensive_ to consume a given amount of `CryptoTransfer` throttle via transactions that incur the extra.


tinker-michaelj:
> From a developer experience perspective, this HIP significantly improves usability by removing artificial transaction limits. However, I'd suggest adding:
> 
> 1. Code examples showing before/after implementations for developers currently working around the 10-item limits
> 2. Fee calculator example demonstrating how to estimate costs for transfers with varying numbers of asset changes
> 
> In order to help developers quickly understand the practical impact and update their implementations accordingly.

For developers using HAPI or fee estimators these changes should be completely transparent.

mgarbs (dismissed):

HIP/hip-1331.md
```diff
@@ -0,0 +1,232 @@
...
+
+This HIP allows some transactions that would previous fail to succeed at a higher price. That is, some tranctions that
+would have failed with `TOKEN_TRANSFER_LIST_SIZE_LIMIT_EXCEEDED` will now succeed at a materially high price. This is
+_conceivably_ a breaking change for clients that somehow rely on failure semantics rather than explicit transaction fee
+limits to budget their `CryptoTransfer` submissions. (Though this seems highly contrived.)
+
+Record and receipt formats are unchanged.
+
+The meaning of the four `ledger.*maxLen` properties changes from "hard cap" to "included count."
+
```
akdev, line 200: If this is considered a backward-compatibility issue, we could introduce a flag on the transaction that the SDK can set explicitly, for example `setExtendedAssetChanges(true)`. The default would remain false, preserving backward compatibility.

mgarbs (approved):

PR #1414, opened 2026-03-03
HIP-1414: Introduce EVM transfer semantics to CryptoTransfer

lukasz-hashgraph:


HIP/hip-1414.md
```diff
@@ -0,0 +1,188 @@
...
+### Impact on hooks
+
+All child EVM calls should be executed in the following order:
+
+1. Execute all `allow` hooks (order unchanged)
+2. Execute all `allowPre` hooks (order unchanged)
+3. Execute transfers and contract calls as described in this HIP
+4. Execute all `allowPost` hooks (order unchanged)
+
+## Backward compatibility
```
Ferparishuertas: We should consider the cryptotransfer systemcontract function on the scope of this HIP


https://github.com/hashgraph/hedera-smart-contracts/blob/main/contracts/system-contracts/hedera-token-service/README.md

HIP/hip-0000.md
```diff
@@ -0,0 +1,188 @@
+---
+hip: 0000
+title: CryptoTransfer Contract Calls
+author: Lukasz Gasior (@lukasz.gasior)
```
0x-ds: Looking at other HIPs, noticed that either github handle or emails were used.

HIP/hip-1414.md
```diff
@@ -0,0 +1,188 @@
...
+Currently, when Contract Account appears as a HBAR recipient on the transfer list of a CryptoTransfer transaction it directly receives the funds, bypassing the execution of its Smart Contract logic.
+In this HIP we propose to change this behavior, so that the logic (EVM code) of the Contract Account is executed in the mentioned scenario.
+
+## Motivation
+
+There are three main drivers for introducing this feature:
+- Having a consistent behavior across the system: HBAR transfer between two accounts should have the same effect, no matter if executed via EVM (call) or SDK (CryptoTransfer). In case of a Smart Contract recipient that means running their code.
+- Give user more control and reliability: owners of Smart Contracts should be able to handle incoming HBAR transfers via their EVM code.
+- Compatibility with the upcoming HIP-1340 update: HIP-1340 introduces "Smart Accounts" via Code Delegation. As a prerequisite to completing HIP-1340 we need to define the CryptoTransfer behavior for regular Smart Contract calls.
+
```
0x-ds, line 29: Again, based on other HIPs, do we need Rationale section?

HIP/hip-0000.md
```diff
@@ -0,0 +1,188 @@
...
+### Impact on Hashscan
+
+Hashscan should ensure that any child Contract Calls will correctly show up on the list (likely already covered by existing mechanisms).
+
+## Security Implications
+
+Child contract calls should only see the call "signed" by the sender of each individual call.
+Even if the original CryptoTransfer transaction was signed by multiple accounts.
+
+TODO: or maybe not?
```
0x-ds: Curious if it make sense to move TODOs to Open issues section?

HIP/hip-0000.md
```diff
@@ -0,0 +1,188 @@
...
+1. Separate out all senders (negative AccountAmount) and receivers (positive AccountAmount) into two lists.
+2. Sort both lists comparing `abs(amount) DESC` then comparing `account shard ASC` then comparing `account realm ASC` then comparing `account number ASC` then comparing `account alias ASC`.
+3. Iterate both lists like so:
+  int senderIdx = 0;
+  int receiverIdx = 0;
+  while (senderIdx < senders.size() && receiverIdx < receivers.size()) {
+      final var senderEntry = senders.get(senderIdx);
+      final var receiverEntry = receivers.get(receiverIdx);
+      final var amount = Math.min(senderEntry.amount, receiverEntry.amount);
+      bilateralTransfer.add(new BilateralTransfer(sender, receiver, amount));
```
0x-ds: If I read this right, `sender` -> `senderEntry` and `receiver` -> `receiverEntry`? Ditto in lines below this one

HIP/hip-0000.md
```diff
@@ -0,0 +1,188 @@
...
+}
+```
+
+Receiver's `EvmContractCall` information should be collected when preparing `BilateralTransfer`s.
+
+Handling logic of CryptoTransfer transaction should then be updated to execute the transfers one by one, following the order of the `BilateralTransfer`s list.
+
+Specifically, for each item on the list:
+
+1. Update the balances of sender and receiver accounts
```
0x-ds: Curious what happens if we send another non-HBAR token (since my understanding is CryptoTransfer supports that)?

HIP/hip-1414.md
```diff
@@ -0,0 +1,188 @@
...
+2. Verify that `EvmContractCall` was provided on the corresponding receiving `AccountAmount` entry, fail if it's missing
+3. Using the gas limit and call data provided in `EvmContractCall` dispatch a child `ContractCall` originating from the sender account
+and targetting the receiving Smart Contract Account. If the result of the call is an error (e.g. EVM revert), fail the `CryptoTransfer` - revert all other transfers and the effects of EVM calls made so far, including hook EVM calls. Fee charges for any execution made so far should be kept.
+
+Note that a single `EvmContractCall` entry on the receiving account may be used in more than one child call execution - in this case the same gas limit and call data values should be used for each execution independently.
+
+Setting `EvmContractCall` on a sender account or on a receiving account that is not a Smart Contract is allowed, but the value will not be used.
+
+The verification of `receiverSigRequired` flag for all recipient accounts should be done before any child calls are dispatched.
+
```
0x-ds, line 147: Curious if we need any aggregate (across all calls) gas limit (checks)?

HIP/hip-1414.md
```diff
@@ -0,0 +1,188 @@
...
+2. Verify that `EvmContractCall` was provided on the corresponding receiving `AccountAmount` entry, fail if it's missing
+3. Using the gas limit and call data provided in `EvmContractCall` dispatch a child `ContractCall` originating from the sender account
+and targetting the receiving Smart Contract Account. If the result of the call is an error (e.g. EVM revert), fail the `CryptoTransfer` - revert all other transfers and the effects of EVM calls made so far, including hook EVM calls. Fee charges for any execution made so far should be kept.
+
+Note that a single `EvmContractCall` entry on the receiving account may be used in more than one child call execution - in this case the same gas limit and call data values should be used for each execution independently.
+
+Setting `EvmContractCall` on a sender account or on a receiving account that is not a Smart Contract is allowed, but the value will not be used.
+
+The verification of `receiverSigRequired` flag for all recipient accounts should be done before any child calls are dispatched.
```
0x-ds, line 146: Curious EvmContractCall results reported in the CryptoTransfer transaction record?

HIP/hip-1414.md
```diff
@@ -0,0 +1,188 @@
...
+## Security Implications
+
+Child contract calls should only see the call "signed" by the sender of each individual call.
+Even if the original CryptoTransfer transaction was signed by multiple accounts.
+
+TODO: or maybe not?
+
+Users must be aware that sending CryptoTransfer might trigger a contract call.
+This is ensured by the fact that they must explicitly provide `EvmContractCall` details for each such call.
+
```
0x-ds, line 259: Can a call also invoke CryptoTransfer? Is reentrancy possible?

0x-ds (commented):
Great job putting this together!

HIP/hip-1414.md
lukasz-hashgraph: Yes it can - via the system contract. Users should be aware of it and handle it accordingly. I'll add a note about that 👍 Thanks

HIP/hip-0000.md
acuarica: +1

HIP/hip-0000.md
acuarica: +1

HIP/hip-1414.md
```diff
@@ -0,0 +1,188 @@
...
+      final var amount = Math.min(senderEntry.amount, receiverEntry.amount);
+      bilateralTransfer.add(new BilateralTransfer(sender, receiver, amount));
+      sender.amount -= amount;
+      receiver.amount -= amount;
+      if (sender.amount == 0) senderIdx++;
+      if (receiver.amount == 0) receiverIdx++;
+  }
+```
+
+Note that even though an account may only appear as a sender/recipient once, it might be involved in multiple `BilateralTransfer`s (e.g. consider a transfer of 10 units to account A from accounts B and C, 5 units withdrawn from each).
```
acuarica, line 87: is it possible for the same account to be both in the sender and receiver lists? If yes, should it be treated as a special case? Should we include some comment clarifying this case?

HIP/hip-0000.md
```diff
@@ -0,0 +1,188 @@
...
+All child EVM calls should be executed in the following order:
+
+1. Execute all `allow` hooks (order unchanged)
+2. Execute all `allowPre` hooks (order unchanged)
+3. Execute transfers and contract calls as described in this HIP
+4. Execute all `allowPost` hooks (order unchanged)
+
+## Backward compatibility
+
+There may be existing Smart Contracts which revert when they receive an EVM call (e.g. if they don't implement a payable `receive` function), that can still receive HBAR via CryptoTransfer without any issues.
```
acuarica:
```suggestion
There may be existing Smart Contracts which revert when they receive an EVM call (e.g. if they don't implement a payable `receive` or `fallback` function), that can still receive HBAR via CryptoTransfer without any issues.
```

HIP/hip-1414.md
```diff
@@ -0,0 +1,188 @@
...
+
+There may also be existing applications that use CryptoTransfer, which currently work without providing `EvmContractCall` details - they will continue to function for Smart Contract Accounts created before the fixed timestamp, but they will need to be adjusted if they wish to support Smart Contract Accounts created on or after that timestamp.
+
+## Impact on Mirror Node
+
+Mirror Node will need to handle the updated format of `CryptoTransfer` and store the corresponding `EvmContractCall` information.
+
+### Impact on RPC Relay
+
+No impact on RPC Relay is expected.
```
acuarica: How this new `CryptoTransfer` semantics would be exposed to Relay users? Say for example one of the `ContractCall` emits some events, how are those events reachable from the Relay? What would be the EVM transaction hash (if any) of this `CryptoTransfer`?

HIP/hip-0000.md
Ferparishuertas: HTS transfer dont trigger smart contract execution in this context.
Does this clarify or where you highlighting another topic?

HIP/hip-1414.md
Ferparishuertas: The semantics are the same as nowadays

- if it is an EthereumTransaction via the usual value transfer, txhash is the usual eth tx hasg
- if it is and HAPI tx, the tx hash is the usual synthetic eth hash, the truncated hapi tx hash to 32 bytes

In both cases,  logs will be part of these transactions.
this already happens for HTS transfers and will happen soon with Hooks

HIP/hip-0000.md
Ferparishuertas: I think its and/or,  as they can receive funds with or without calldata. Users need to decide which cases they  want to support

HIP/hip-1414.md
Ferparishuertas: The algorithm already exists and is already implemented in multitransfers. @lukasz-hashgraph will mention it . to reduce the focus on the actual algorithm definition

HIP/hip-0000.md
```diff
@@ -0,0 +1,188 @@
...
+  }
+```
+
+Note that even though an account may only appear as a sender/recipient once, it might be involved in multiple `BilateralTransfer`s (e.g. consider a transfer of 10 units to account A from accounts B and C, 5 units withdrawn from each).
+
+The same notion of `BilateralTransfer` and the same algorithm should be used across the system whenever this concept is needed (e.g. when generating synthetic EVM events for native transfers).
+
+We will now extend this base algorithm to support the primary goal of this HIP: Contract Calls.
+
+We propose to extend CryptoTransfer's AccountAmount with an additional optional field:
```
Ferparishuertas: As discussed within the team, the payer is the one paying fees and gas, not the hbar senders. That is important from a product perspective.

Based on this we should 
- have a global optional gaslimit per cryptotranfer
- do not have a gaslimit per AccountAmount
- have a global optional calldata
- have a optional calldata per AccountAmount that can override the global calldata for this destination 

Gaslimit will be global, and all the transaction will deduct gas from gaslimit on usage, and remaining gas will be passed through the transfer chain.

HIP/hip-1414.md
```diff
@@ -0,0 +1,188 @@
...
+There may be existing Smart Contracts which revert when they receive an EVM call (e.g. if they don't implement a payable `receive` function), that can still receive HBAR via CryptoTransfer without any issues.
+We propose to introduce a backward compatibility mechanism to support this use case.
+Contract Calls should only be dispatched for Contract Accounts created on or after a fixed timestamp (exact value TBD).
+Any Smart Contract Accounts created before that should be treated the same way as non-contract accounts. Specifically:
+- do not dispatch child contract calls
+- if they're the recipient and `EvmContractCall` is set, then the value is ignored
+
+There may also be existing applications that use CryptoTransfer, which currently work without providing `EvmContractCall` details - they will continue to function for Smart Contract Accounts created before the fixed timestamp, but they will need to be adjusted if they wish to support Smart Contract Accounts created on or after that timestamp.
+
+## Impact on Mirror Node
```
Ferparishuertas: I suggests to add a impacts subsection on the spec as on 1340, and has subitems like Mirror node  impact at the same level

HIP/hip-1414.md
```diff
@@ -0,0 +1,188 @@
...
+
+The sender of the call (accessible via the ORIGIN op code) should be the sender account of the corresponding `BilateralTransfer`. It should be the account from which funds are withdrawn, including when allowance is used.
+
+The value (accessible via the CALLVALUE op code) should be the amount of the corresponding `BilateralTransfer`.
+
+The nonce of the sender account should NOT be incremented.
+
+The effects (state changes) of all previous transfers (regardless of whether they trigger a call or not) and hook calls should be visible in subsequent calls (e.g. when querying for the HBAR balance of an account processed earlier).
+
+### Impact on hooks
```
Ferparishuertas: I suggest to call it Hooks Interoperability as on 1340

HIP/hip-0000.md
```diff
@@ -0,0 +1,188 @@
...
+Contract Calls should only be dispatched for Contract Accounts created on or after a fixed timestamp (exact value TBD).
+Any Smart Contract Accounts created before that should be treated the same way as non-contract accounts. Specifically:
+- do not dispatch child contract calls
+- if they're the recipient and `EvmContractCall` is set, then the value is ignored
+
+There may also be existing applications that use CryptoTransfer, which currently work without providing `EvmContractCall` details - they will continue to function for Smart Contract Accounts created before the fixed timestamp, but they will need to be adjusted if they wish to support Smart Contract Accounts created on or after that timestamp.
+
+## Impact on Mirror Node
+
+Mirror Node will need to handle the updated format of `CryptoTransfer` and store the corresponding `EvmContractCall` information.
```
Ferparishuertas: Mirror Node will need to handle the updated format of `CryptoTransfer`, store the corresponding `EvmContractCall` information, and generate a contractresuts entry

HIP/hip-1414.md
Ferparishuertas: makes sense to me

HIP/hip-1414.md
Ferparishuertas: I suggest to be more specific here as cryptotransfer will generate contractResults, and will be included within the EVM block on this new cases, extending the existing the existing case. We need to ensure that these transaction are correctly represented, and tested

HIP/hip-0000.md
lukasz-hashgraph: 👍 good catch, thx

HIP/hip-0000.md
lukasz-hashgraph: Yup, `senderEntry.accountId`. Thx, updated here and in the lines below.

HIP/hip-0000.md
lukasz-hashgraph: Any tokens transfers should happen after we've processed all HBAR transfers.
Thanks for pointing that out - I'll clarify.

non-HBAR token transfers do not trigger a call, even if recipient is a contract.

HIP/hip-1414.md
lukasz-hashgraph: There is a single list and an account can only appear there once (either as a sender or receiver).

HIP/hip-1414.md
lukasz-hashgraph: I've updated this HIP with a new approach to gas charging, please re-review 🙏

HIP/hip-1414.md
lukasz-hashgraph: It should work pretty much the same as it does for hooks. That said - I don't yet have a clear picture of how it's being handled 🤔 I'll get back to your question

HIP/hip-1414.md
```diff
@@ -0,0 +1,250 @@
...
+
+This default amount seems to be a reasonable balance between the two aspects of maintaining backward compatibility:
+- supporting existing applications with minimal disruption
+- protecting from unintential high charges
+
+### Impact on HTS System Contract
+
+HTS System Contract will be extended with a new variant (V3) of the `cryptoTransfer(...)` method with the following signature:
+
+`cryptoTransfer(((address,int64,bool)[]),(address,(address,int64,bool)[],(address,address,int64,bool)[])[],(bytes,int64))`
```
0x-ds: Might be a type mismatch for `(bytes, int64)` tuple, as I think we mentioned it should be `uint64` on line 86

HIP/hip-1414.md
```diff
@@ -0,0 +1,250 @@
...
+category: Service
+needs-hiero-approval: Yes
+needs-hedera-review: Yes
+hedera-review-date:
+hedera-acceptance-decision:
+status: Draft
+last-call-date-time: 
+discussions-to: 
+created: 2026-02-17
+updated: 2026-02-17
```
Ferparishuertas, line 16: We need a required entry. referencing 1340

HIP/hip-1414.md
```diff
@@ -0,0 +1,250 @@
+---
+hip: 1414
+title: CryptoTransfer Contract Calls
```
Ferparishuertas: I think we should focus the HIP on the a global cryptotransfer concept, not only contract.

I suggest to call the HIP Cryptotransfer as native EVM transactions, explain the sources of cryptotransfer and that they need to be consistent across the board, executing smart contract or delegation. We can focus on contract specificics narratives as a start point to make it understandanle, and then go into the other details

HIP/hip-1414.md
```diff
@@ -0,0 +1,250 @@
...
+
+We will now extend this base algorithm to support the primary goal of this HIP: Contract Calls.
+
+We propose to extend `CryptoTransferTransactionBody` with an additional optional field:
+
+```
+/**
+ * Specifies details of EVM contract calls executed within this transfer.
+ */
+message EvmContractCalls {
```
Ferparishuertas, line 99: I suggest to rename this to EVMCall, to match a more generic naming for ContractAccount and EOAs

Ferparishuertas (commented):
Thanks for putting this together. I left a couple of comments

HIP/hip-1414.md
lukasz-hashgraph: `Evm` (only E in capital) seems to be the naming we follow in protobuf. I went with `EvmContractCalls` because it serves the same purpose as `EvmHookCall` (providing call data and gas limit configuration to the call), but this single record applies to all calls within the transfer, hence plural.

HIP/hip-1414.md
lukasz-hashgraph: It's the other way around, HIP-1340 requires this one.

HIP/hip-0000.md
lukasz-hashgraph: Right, I assumed this is already being done for all child calls (e.g. hooks), but I'll mention it 👍

HIP/hip-1414.md
lukasz-hashgraph: There isn't a dedicated section in HIP-1340 - all "Impact on *" headings are under "Specification" - and that's what the HIP template suggests.
Looks like I have misaligned the "Impact on Mirror Node" heading though - pushed it one step deeper.

HIP/hip-0000.md
lukasz-hashgraph: Agreed - the HIP has been updated.

HIP/hip-1414.md
lukasz-hashgraph: Make sense I'll add a bit more background 👍
I guess you meant to say "explain sources of value/hbar transfer" (CryptoTransfer, Selfdestruct, etc)? Otherwise I'm not sure I understand what sources sources of CryptoTransfer means?

lukasz-hashgraph: As for the title of the HIP, wdyt about: "Introduce EVM transfer semantics to CryptoTransfer"? It captures all the changes we propose (including the follow-up HIP-1340) quite accurately.

"Cryptotransfer as native EVM transactions" might be a bit misleading.

HIP/hip-1414.md
lukasz-hashgraph: Good catch 👍

HIP/hip-1414.md
```diff
@@ -0,0 +1,251 @@
...
+Call details should be explicitly provided in the `EvmContractCalls` field, otherwise a default value will be used.
+
+## Reference Implementation
+
+The reference implementation must be complete before any HIP is given the status
+of “Final.” The final implementation must include test code and documentation.
+
+## Rejected Ideas
+
+None at this time.
```
gkozyryatskyy, line 272: Should we explain that we considered not supporting this, but decided to be EVM-oriented even for HAPI calls?

HIP/hip-1414.md
```diff
@@ -0,0 +1,251 @@
...
+Child contract calls should only see the call "signed" by the sender of each individual call.
+Even if the original CryptoTransfer transaction was signed by multiple accounts.
+
+Users must be aware that:
+- sending a CryptoTransfer transaction might trigger zero or more EVM contract calls
+- re-entrancy is possible (the EVM call itself might dispatch a child CryptoTransfer via a System Contract)
+
+## How to Teach This
+
+Sending a CryptoTransfer transaction may result in zero or more Smart Contract calls.
```
gkozyryatskyy, line 262: Should we add more details on the exact situations where there will be more than 0 Smart Contract calls?

HIP/hip-1414.md
```diff
@@ -0,0 +1,251 @@
...
+
+Currently, when Contract Account appears as a HBAR recipient on the transfer list of a CryptoTransfer transaction it directly receives the funds, bypassing the execution of its Smart Contract logic.
+In this HIP we propose to change this behavior, so that the logic (EVM code) of the Contract Account is executed in the mentioned scenario.
+
+## Motivation
+
+There are three main drivers for introducing this feature:
+- Having a consistent behavior across the system: HBAR transfer between two accounts should have the same effect, no matter if executed via EVM (call) or SDK (CryptoTransfer). In case of a Smart Contract recipient that means running their code.
+- Give user more control and reliability: owners of Smart Contracts should be able to handle incoming HBAR transfers via their EVM code.
+- Compatibility with the upcoming HIP-1340 update: HIP-1340 introduces "Smart Accounts" via Code Delegation. As a prerequisite to completing HIP-1340 we need to define the CryptoTransfer behavior for regular Smart Contract calls.
```
gkozyryatskyy: nit: lets make a clickable links from [HIP-1340](https://hips.hedera.com/hip/hip-1340)?
We can apply this to all "HIP references" in this doc

HIP/hip-1414.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Rationale
+
+This HIP aims to satisfy the needs of users and the upcoming HIP-1340 while attempting to maintain maximum backward compatibility.
+
+## User Stories
```
gkozyryatskyy, line 55: both provided user stories are referring to `Smart Contract` accounts. Should we add a user story related to [HIP-1340](https://hips.hedera.com/hip/hip-1340)? 
Like:
- As an `EOA`s owner with Code Delegation ...
- As a network user, transferring `HBAR` via `CryptoTransfer` to `EOA` with Code Delegation ...
etc...

HIP/hip-1414.md
```diff
@@ -0,0 +1,251 @@
...
+- As a network user, I expect that transferring HBAR to a Contract Account using CryptoTransfer results in execution of that Smart Contract's EVM code
+
+## Specification
+
+Since `CryptoTransfer`'s `TransferList` is just a list of accounts and their respective balance changes (which all must sum up to 0), we shall start by introducing a notion of a transfer between two accounts and a canonical algorithm for extracting those transfers from the `TransferList`. For the sake of the algorithm we'll call this ephemeral record a `BilateralTransfer`.
+`BilateralTransfer` in its base form consists of three fields: `sender`, `receiver` and `amount`.
+
+To collect `BilateralTransfer`s from the `TransferList` we propose to adopt the following algorithm, which is currently being used for generating synthetic EVM `Transfer(...)` events:
+
+1. Separate out all senders (negative AccountAmount) and receivers (positive AccountAmount) into two lists.
```
gkozyryatskyy: for `senders` we should `abs(sender.amount)` in resulted lists. For `senderEntry.amount -= amount;` to work as expected. And in this case we will not need to `abs(amount)` for sorting.

Mb it is worth to mention

HIP/hip-1414.md
```diff
@@ -0,0 +1,251 @@
...
+2. Sort both lists comparing `abs(amount) DESC` then comparing `account shard ASC` then comparing `account realm ASC` then comparing `account number ASC` then comparing `account alias ASC`.
```
gkozyryatskyy: comparing `account number ASC`, if `account number` exists? 
In theory, there can be account w/o `account number` (In the same time, this should be the case if you are getting data from record stream and accounts are already created).

Same to `alias` sorting, if it will be there

gkozyryatskyy: [Existed CN sorting](https://github.com/hiero-ledger/hiero-consensus-node/blob/7fc300578c5c301b49e52b9457a7e138aa842893/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/exec/systemcontracts/hts/transfer/TransferEventLoggingUtils.java#L69) do not include `alias` sorting (same for [MN sorting](https://github.com/hiero-ledger/hiero-mirror-node/blob/1580984db55052eaedbe35d3c4f0ba2c2d6f59ec/importer/src/main/java/org/hiero/mirror/importer/parser/record/entity/TransferEventsGenerator.java#L21)), This is because `List<AccountAmount>` already expected to be sorted by `alias` in record stream.

So mb it make sense to remove alias sorting from this also or add it to existed algos.
If you are getting your source data from record stream, mb it is better to exclude this sorting to match exactly.

HIP/hip-1414.md
```diff
@@ -0,0 +1,251 @@
...
+### Impact on RPC Relay
+
+No impact on RPC Relay is expected.
+
+### Impact on SDK
+
+SDKs will need to add support for setting the `EvmContractCalls` field.
+
+### Impact on Hashscan
+
```
Ferparishuertas, line 237: We need to show the EvmContractCall Params uses in Hashscan if set, on any parent or child transaction

HIP/hip-1414.md
```diff
@@ -0,0 +1,251 @@
...
+```
+
+HTS System Contract does not currently support hooks, so nested hook invocations (like `Hook call -> ... -> Hook call`) are not possible.
+
+### Throttle limits
+
+All calls made within the transfer are subject to the same throttle limits as any other top-level Smart Contract transaction (see HIP-1249).
+As a result, the transfer might fail with a `CONSENSUS_GAS_EXHAUSTED` error.
+
+### Impact on Mirror Node
```
Ferparishuertas, line 215: We usually forget about all the dependencies, and we should detail alll

- MN need to ensure gas estimation via rest api. includes de extra charges for contract account invocations
- MN needs to update apis with the new information. I suggest to add a TODO, allowing MN to detail all the changes
- MN needs to represent any cryptotransfer with contract execution with a contractResult

HIP/hip-1414.md
```diff
@@ -0,0 +1,251 @@
...
+
+Mirror Node will need to handle the updated format of `CryptoTransferTransactionBody` and store the corresponding `EvmContractCalls` information.
+It will also need to process the child EVM calls and store the results (similarly to hook EVM calls).
+
+### Impact on RPC Relay
+
+No impact on RPC Relay is expected.
```
Ferparishuertas: We usually forget about all the dependencies, and we should detail alll

- eth_estimateGas should respond with accuracy to any estimation, that includes cryptotransfer with delegation implications or contract account invocations
- eth_call invocations should be correctly simulated
- We need to ensure that parent cryptotransfers with contract account or delegated accounts execution, show contract results within MN and are correctly represented in the relay. Same for inner cryptotransfers, being both, child and parent correctly respresented.

HIP/hip-1414.md
```diff
@@ -0,0 +1,251 @@
...
+
+The payer of the original CryptoTransfer transaction pays for all gas charges incurred when executing the contract calls up to the specified `cumulative_gas_limit` value.
+
+Similarly to hooks (see HIP-1195 -> Gas charging) we propose a reduced intrinsic gas rate for contract calls. We propose adding another network property to give this effect while keeping it customizable for Hiero network operators.
+
+```
+cryptotransfer.evm.intrinsicGasCost=1000
+```
+
+If `EvmContractCalls` is not provided we propose to use a default of `3300` gas units available individually for each call. This value consists of the standard gas stipend of 2300 available for `tranasfer()` calls in Solidity plus the 1000 of intrinsic gas mentioned above.
```
Ferparishuertas: tranasfer -> tranafer

HIP/hip-1414.md
```diff
@@ -0,0 +1,251 @@
...
+
+The transfer list is, by default, limited to 10 items. That gives a maxium of 9 `BilateralTransfer`s per transaction.
```
Ferparishuertas: maxium -> maximum

HIP/hip-1414.md
```diff
@@ -0,0 +1,251 @@
...
+The maximum gas charge for users not providing custom `EvmContractCalls` details will therefore be: `3300 * 9 = 29700` gas units.
+
+This default amount seems to be a reasonable balance between the two aspects of maintaining backward compatibility:
+- supporting existing applications with minimal disruption
+- protecting from unintential high charges
```
Ferparishuertas: unintential-> unintentional

HIP/hip-1414.md
```diff
@@ -0,0 +1,251 @@
...
+Handling logic of CryptoTransfer transaction will then be updated to execute HBAR transfers one by one, following the order of the `BilateralTransfer`s list.
+
+As a first step assign `remaining_gas := cumulative_gas_limit`. If `EvmContractCalls` information was not provided use a default of `remaining_gas := 3300` (further explained in the section dedicated to gas charging).
+User is allowed to set `cumulative_gas_limit := 0`, in which case no EVM calls should succeed. Do not use a default of `3300` if `cumulative_gas_limit` is set to 0.
+
+Then, for each item on the list:
+
+1. (All transfers) Update the HBAR balances of sender and receiver accounts.
+2. (Only if receiver is a Contract Account) Using `remaining_gas` and call data from `EvmContractCalls` dispatch a child `ContractCall` originating from the sender account
+and targetting the receiving Smart Contract Account. Use `0x` (empty) for call data if `EvmContractCalls` was not provided. If the result of the call is an error (e.g. EVM revert), fail the `CryptoTransfer` - revert all other transfers and the effects of EVM calls made so far, including hook EVM calls. Fee charges for any execution made so far should be kept.
```
Ferparishuertas: targetting -> targeting

Ferparishuertas (commented):
Some extra comments

HIP/hip-1414.md
Ferparishuertas: Can be closed

HIP/hip-1414.md
Ferparishuertas: Can be closed

HIP/hip-1414.md
```diff
@@ -0,0 +1,251 @@
...
+Similarly to hooks (see HIP-1195 -> Gas charging) we propose a reduced intrinsic gas rate for contract calls. We propose adding another network property to give this effect while keeping it customizable for Hiero network operators.
+
+```
+cryptotransfer.evm.intrinsicGasCost=1000
+```
+
+If `EvmContractCalls` is not provided we propose to use a default of `3300` gas units available individually for each call. This value consists of the standard gas stipend of 2300 available for `tranasfer()` calls in Solidity plus the 1000 of intrinsic gas mentioned above.
+
+The transfer list is, by default, limited to 10 items. That gives a maxium of 9 `BilateralTransfer`s per transaction.
+The maximum gas charge for users not providing custom `EvmContractCalls` details will therefore be: `3300 * 9 = 29700` gas units.
```
Ferparishuertas, line 171: We should define the formula for the default gaslimit, as it will be part of the Sidecar, and final MN contractResult.

numberofTx + 1  * 3300 ?

HIP/hip-1414.md
lukasz-hashgraph: As discussed offline: this HIP will build on top of HIP-1340 and hence require it.

HIP/hip-1414.md
lukasz-hashgraph: Like what should I put in this section? "We considered not to create this HIP at all."? 😄

HIP/hip-1414.md
lukasz-hashgraph: What kind of details do you mean? The whole "Specification" section pretty much explains just that, doesn't it?

HIP/hip-1414.md
lukasz-hashgraph: Yup, good point. I've included EOA support in this HIP.

HIP/hip-1414.md
lukasz-hashgraph: Yup, good point.

HIP/hip-1414.md
lukasz-hashgraph: Added a note 👍

lukasz-hashgraph: This algorithm is meant to be used during transaction processing (before we generate any outputs). The list is raw, as submitted by the user - unsorted.

lukasz-hashgraph: I added a note below, let me know if that solves your doubts:

> The same notion of `BilateralTransfer` and the same algorithm should be used across the system whenever this concept is needed (e.g. when generating synthetic EVM events for native transfers, where it's already used). The (pseudo)code above only serves as an example, implementators are free to adjust it (e.g. use different primitives or exact collection types) as long as it produces the same resulant entries in the exact same order (which might require extra care e.g. if the transfer list comes partially pre-sorted).

HIP/hip-1414.md
lukasz-hashgraph: tranafer -> transfer 😄

HIP/hip-1414.md
lukasz-hashgraph: The default is not a single value. If the value is not provided by the user than a default _set of rules_ should be applied (as described).
I will clarify what we discussed though (if gasLimit not provided than set gasLimit = gas actually used).

PR #1415, opened 2026-03-04
Add HIP-1415 for configurable HBAR decimals

stoqnkpL:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-1415.md
+++ b/HIP/hip-1415.md
@@ -1,5 +1,5 @@
 ---
-hip: <TBD>
+hip: 1415
 title: Configurable Native Coin Decimal Precision
 author: Stoyan Panayotov <@stoqnkpL>
 requested-by: 
@@ -10,7 +10,7 @@ needs-hedera-review: Yes
 status: Draft
 created: 2026-03-04
 updated: 2026-03-04
-discussions-to: https://github.com/hiero-ledger/hiero-improvement-proposals/discussions/<TBD>
+discussions-to: https://github.com/hiero-ledger/hiero-improvement-proposals/discussions/1416
 ---
 
 ## Abstract
--- a/HIP/hip-xxx.md
+++ b/HIP/hip-xxx.md
@@ -1,0 +1,292 @@
+---
+hip: <TBD>
+title: Configurable Native Coin Decimal Precision
+author: Stoyan Panayotov <@stoqnkpL>
+requested-by: 
+type: Standards Track
+category: Core
+needs-hiero-approval: Yes
+needs-hedera-review: Yes
+status: Draft
+created: 2026-03-04
+updated: 2026-03-04
+discussions-to: https://github.com/hiero-ledger/hiero-improvement-proposals/discussions/<TBD>
+---
+
+## Abstract
+
+This HIP introduces a genesis-time configuration property, `nativeCoin.decimals`, that controls the number of decimal places in a Hiero network's native coin subunit. The value MUST be an integer between 0 and 18 inclusive, with a default of 8 (preserving the existing 1 HBAR = 10⁸ tinybar relationship). Once set at genesis, the value is persisted as an immutable singleton in network state and SHALL NOT change for the lifetime of the network. All denomination-sensitive paths — fees, staking rewards, EVM wei conversions, exchange rate calculations, and query responses — automatically scale to the configured subunit.
+
+## Motivation
+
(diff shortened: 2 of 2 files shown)


HIP/hip-1415.md
```diff
@@ -0,0 +1,292 @@
...
+- **DeFi-oriented networks** may prefer 6 decimal places (matching USDC/USDT conventions) or 18 decimal places (matching Ethereum's native wei precision) to simplify integration with existing tooling and smart contracts.
+- **Regulatory environments** may mandate specific precision levels for auditing, reporting, or compliance purposes.
+
+Without this configurability, network operators must either accept the hardcoded 8-decimal assumption or maintain a permanent fork of the codebase — neither of which is sustainable for ecosystem growth.
+
+## Rationale
+
+### Scale at boundary, not internally
+
+The fee schedule and internal fee computation logic are calibrated for 10⁻⁸ HBAR (the "default tinybar"). Rather than modifying every internal calculation to use a configurable base — which would be invasive, error-prone, and hard to audit — this proposal adopts a **"scale at boundary"** design:
```
rbair23, line 36: I think this is problematic due to the loss of precision. When going from 8 to 6 decimals, this is probably OK. Going from 8 to 18 is a problem Maybe this HIP can be for 0-8 decimals but not more, and a later HIP goes up to 18 and includes the broader set of changes? I don't really like that idea.

HIP/hip-1415.md
```diff
@@ -0,0 +1,292 @@
...
+needs-hedera-review: Yes
+status: Draft
+created: 2026-03-04
+updated: 2026-03-04
+discussions-to: https://github.com/hiero-ledger/hiero-improvement-proposals/discussions/1416
+---
+
+## Abstract
+
+This HIP introduces a genesis-time configuration property, `nativeCoin.decimals`, that controls the number of decimal places in a Hiero network's native coin subunit. The value MUST be an integer between 0 and 18 inclusive, with a default of 8 (preserving the existing 1 HBAR = 10⁸ tinybar relationship). Once set at genesis, the value is persisted as an immutable singleton in network state and SHALL NOT change for the lifetime of the network. All denomination-sensitive paths — fees, staking rewards, EVM wei conversions, exchange rate calculations, and query responses — automatically scale to the configured subunit.
```
Ferparishuertas, line 18: Polkadot Substrate allow precision definition on the native token. Maybe we want to follow a similar naming strategy

"properties": {
    "tokenSymbol": "MYT",
    "tokenDecimals": 12,
    //"ss58Format": 42
  },

https://docs.substrate.io/build/chain-spec

HIP/hip-1415.md
```diff
@@ -0,0 +1,292 @@
...
+
+### Range 0–18
+
+The lower bound of 0 supports whole-unit-only coins (no fractional subunits). The upper bound of 18 matches Ethereum's wei precision (10¹⁸ wei per ether), which is the maximum precision needed for full EVM compatibility. Values above 18 would exceed the precision of a 64-bit `long` when multiplied by practical balance amounts, introducing overflow risk without practical benefit.
+
+## User Stories
+
+**As a network operator**, I want to configure the native coin's decimal precision at genesis so that my network's smallest unit matches my tokenomic design (e.g., 6 decimals for a stablecoin-like denomination or 18 for full EVM compatibility).
+
+**As an SDK developer**, I want to query the network's configured decimal precision via the `NetworkGetVersionInfo` API so that my application can correctly format and parse native coin amounts without hardcoding assumptions.
```
Ferparishuertas, line 56: I suggest to add a custom RPC method tp the relay , called network_properties to show relevant config information from the network in an ethereum style

Ferparishuertas (commented):
Thank you for putting the work on the HIP.

I just added a couple of comments

HIP/hip-1415.md
stoqnkpL: Actually, I think we can't allow `nativeCoin.decimals` in range [9, 18] because the total supply of 50 billion whole coins means `50_000_000_000 * 10^decimals` overflows `Long.MAX_VALUE` at decimals > 8. At 18 decimals we'd be able to support at most 9 whole coins. The long (int64) type is locked at the protobuf wire format level — every balance, transfer amount, fee, and staking amount is defined as int64/uint64/sint64 in ~80 .proto files with ~316 occurrences. This is the consensus wire format between nodes and the public API contract consumed by all SDKs. So I'd suggest allowing ranges [0,8].

stoqnkpL: Probably we'd want to make the total supply configurable as well.

stoqnkpL:
Will not go forward with this for now.

PR #1424, opened 2026-03-11
Block and State Merkle Tree Hashing Scheme

Nana-EC:
**Description**:

This HIP specifies the complete hashing scheme and Merkle tree structures used by the Hiero network to compute
verifiable, tamper-evident digests of all data in both the block stream and consensus node state.
Building on the block stream format introduced in [HIP-1056](https://hips.hedera.com/hip/hip-1056), this specification defines
- the node-prefixing convention that provides domain separation between leaf and internal nodes
- the three distinct Merkle tree types used for different parts of the block; the fixed 16-leaf block root structure with its eight defined sub-trees and eight reserved expansion slots
- the streaming incremental Merkle tree algorithm designed for efficient hash computation as block items are produced
- the complete set of protobuf messages for the block footer, block proof variants, and filtered or redacted item representations.

Together, these primitives enable any party to verify
the integrity of any individual item in any block or state, and to generate concise Merkle proofs for both recent and
historical blocks.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


assets/hip-1424/Hashing - Block Root.png
jsync-swirlds, line 1: Block Proof has a single signature, not repeated.

assets/hip-1424/state-merkle-tree-hashing.png
jsync-swirlds, line 1: The note top-left is missing `0x01` for nodes with a single child.

assets/hip-1424/Hashing - Block Root Tree.png
jsync-swirlds, line 1: comment bottom-right is confusing.
Instead of "they all have prefix ...", perhaps just note that
> Some single-child nodes must exist in order to have stable paths. Single-child nodes have a prefix of `0x01`.

HIP/hip-XXXX-block-stream-hashing.md
```diff
@@ -0,0 +1,684 @@
...
+  Richard Bair <@rbair23>
+requested-by: Hiero Block Builders and Validators, External Verifiers
+type: Standards Track
+category: Core
+needs-hiero-approval: Yes
+needs-hedera-review: Yes
+status: Draft
+last-call-date-time: 2026-04-01T07:00:00Z
+created: 2026-02-27
+updated: 2026-02-27
```
jsync-swirlds: Should created and/or updated be the current date?

HIP/hip-XXXX-block-stream-hashing.md
```diff
@@ -0,0 +1,684 @@
...
+  Nana Essilfie-Conduah <@Nana-EC>, Jasper Potts <@jasperpotts>,
+  Joseph Sinclair <@jsync-swirlds>, Leemon Baird <@lbaird>,
+  Richard Bair <@rbair23>
+requested-by: Hiero Block Builders and Validators, External Verifiers
+type: Standards Track
+category: Core
+needs-hiero-approval: Yes
+needs-hedera-review: Yes
+status: Draft
+last-call-date-time: 2026-04-01T07:00:00Z
```
jsync-swirlds: Probably should not set this unless/until the date is set.

HIP/hip-1424.md
```diff
@@ -0,0 +1,684 @@
...
+
+- `0x00` — leaf node
+- `0x01` — internal node with a single (left) child
+- `0x02` — internal node with two children
+
+As these prefix bytes are different, no hash produced for a leaf can equal a hash produced for an internal node for any
+input, regardless of the leaf content or internal node child hashes.
+This is the minimal, maximally efficient form of domain separation for a binary tree. For a thorough discussion of why
+this is necessary see [the length-extension attack on Wikipedia](https://en.wikipedia.org/wiki/Length_extension_attack)
+and [the Merkle tree second-preimage attack](https://www.rareskills.io/post/merkle-tree-second-preimage-attack).
```
jsync-swirlds: For consistency, we might want to use `https://en.wikipedia.org/wiki/Preimage_attack` for the link here.

HIP/hip-1424.md
```diff
@@ -0,0 +1,684 @@
...
+## Rationale
+
+### SHA-384 as the Hash Function
+
+SHA-384 (a truncation of SHA-512) is used throughout this scheme. It provides a 384-bit (48-byte) digest, which is
+larger than the 256-bit output of SHA-256 or Keccak-256 used by EVM chains.
+This larger output eliminates the classes of collision attacks that have theoretical relevance against 256-bit hashes
+in second-preimage contexts. SHA-384 is standardized ([FIPS 180-4](https://csrc.nist.rip/publications/detail/fips/180/4/final)),
+widely hardware-accelerated, and offers a comfortable security margin well above the 128-bit post-quantum security level that
+256-bit hashes approach asymptotically.
```
jsync-swirlds: Half of this is irrelevant and unhelpful.
```suggestion
This larger output offers a comfortable security margin well above the 128-bit post-quantum security level that 256-bit hashes approach asymptotically.
```

HIP/hip-1424.md
```diff
@@ -0,0 +1,684 @@
...
+produces byte-for-byte identical digests from the same data. That agreement is the prerequisite for threshold
+signatures to be meaningful.
+
+
+## Rationale
+
+### SHA-384 as the Hash Function
+
+SHA-384 (a truncation of SHA-512) is used throughout this scheme. It provides a 384-bit (48-byte) digest, which is
+larger than the 256-bit output of SHA-256 or Keccak-256 used by EVM chains.
```
jsync-swirlds: Also, we should tie this to the Hiero standard use of SHA2-384
```suggestion
SHA-384 (a truncation of SHA-512) is used throughout this scheme. SHA-384 is the standard hash function for Hiero networks.
The SHA-384 algorithm produces a 48-byte digest, rather than the 32-byte digest of SHA-256 or Keccak-256 used by EVM chains.
```

HIP/hip-1424.md
```diff
@@ -0,0 +1,684 @@
...
+This HIP specifies the complete hashing scheme and Merkle tree structures used by the Hiero network to compute
+verifiable, tamper-evident digests of all data in both the block stream and consensus node state.
+
+Building on the block stream format introduced in [HIP-1056](https://hips.hedera.com/hip/hip-1056), this specification
+defines
+- the node-prefixing convention that provides domain separation between leaf and internal nodes
+- the three distinct Merkle tree types used for different parts of the block; the fixed 16-leaf block root structure
+    with its eight defined sub-trees and eight reserved expansion slots
+- the streaming incremental Merkle tree algorithm designed for efficient hash computation as block items are produced
+- the complete set of protobuf messages for the block footer, block proof variants, and filtered or redacted item representations.
```
jsync-swirlds: Some corrections, and removing the block item definitions that belong in the Block Stream HIP.

```suggestion
- A node-prefixing convention that provides domain separation between leaf and internal nodes
- A "Merkle Mountain Top" structure that combines eight defined subtrees, eight reserved expansion subtrees, and the fixed structure that combines these subtrees into a single unified Merkle structure.
- the streaming incremental Merkle tree algorithm designed for efficient hash computation as streamed block items are processed
```

HIP/hip-1424.md
jsync-swirlds, line 1: This is an OK start, but it needs a thorough editing pass to correct misapprehension and not-quite-on-track misdirections.

I'll send a global edit as soon as I can.

jsync-swirlds (changes requested):
A few small changes, and a note that the actual HIP document needs broad rework.

jsync-swirlds (dismissed):

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+<dt>Internal node</dt><dd>A Merkle tree node that has one or two children and
+carries no data of its own.
+</dd>
+<dt>Domain separation</dt><dd>The use of distinct prefixes on hash inputs to
+ensure that hashes computed for different node types occupy disjoint regions
+of the hash output space, preventing cross-type collisions.
+</dd>
+<dt>Block root hash</dt><dd>The single 48-byte SHA-384 digest that summarizes
+the entire content of a block, including all transactions, outputs, state
+changes, the merkle tree of previous blocks, and the chain of previous blocks.
```
mhess-swl: We could add something about block footer and block proof items being metadata only, and not part of the hash.

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+<dt>Streaming binary Merkle tree</dt><dd>An incremental algorithm that builds a
+Merkle tree root hash one leaf at a time while maintaining only O(log N) bytes
+of intermediate state.
+</dd>
+<dt>Complete balanced binary Merkle tree</dt><dd>A tree in which all levels are
+full except the lowest, and all leaves on the lowest level are positioned as far
+left as possible. Used for the state tree.
+</dd>
+<dt>Merkle Mountain Top</dt><dd>A tree with a predetermined number of leaves
+and fixed sub-tree positions that is topped by the block root.
```
mhess-swl:
```suggestion
<dt>Merkle Mountain Top</dt><dd>A tree with predetermined, fixed sub-tree positions that is topped by the block root.
```

'predetermined number of leaves' conflicts with statements elsewhere directing that subroots not be treated as leaves.

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+</dd>
+</dl>
+
+## User Stories
+1. As a consensus node implementor, I need a precise specification of the
+   hashing algorithm so that independent implementations of the consensus node
+   produce byte-for-byte identical block root hashes, enabling reliable
+   verification of the block chain across time.
```
mhess-swl, line 158: Another potential add:

As a consensus node implementor, I need a block verification format that facilitates state proofs of any node (regardless of type) in any block.

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+
+| Node type | Prefix | Formula |
+|---|---|---|
+| Empty leaf | `0x00` | `hash(0x00)` |
+| Data leaf | `0x00` | `hash(0x00 \|\| leaf_bytes)` |
+| Internal, 1 child | `0x01` | `hash(0x01 \|\| child_hash)` |
+| Internal, 2 children | `0x02` | `hash(0x02 \|\| left_hash \|\| right_hash)` |
+
+These four domains are disjoint: no hash computed with one prefix is likely to
+equal a hash computed with a different prefix for any input.
```
mhess-swl:
```suggestion
These four domains are disjoint: no hash computed with one prefix will feasibly be
equal to a hash computed with a different prefix for any input.
```

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+
+---
+
+#### 1. Fixed Size Merkle Tree (Block Merkle Mountain Top)
+
+Used for: the top level of every block.
+
+This is a tree with exactly **16 leaves**, permanently fixed in shape. Its leaf
+positions are assigned once and SHALL NOT change. The fixed shape enables state
+proofs to encode paths that remain valid across all past and future blocks.
```
mhess-swl:
```suggestion
proofs to encode paths that remain valid across all past and future blocks. In its current form, only the first 8 subtrees are defined. The other 8 are left for future expansion, and are not encoded in the Merkle Mountain calculation. Until one or more of these future nodes are utilized, their subtree parent node is treated as a node with a single (left) child. 
```

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+[complete balanced binary tree](https://en.wikipedia.org/wiki/Binary_tree#Types_of_binary_trees)
+in which all levels are full except the lowest, and all leaves on the lowest
+level are positioned as far left as possible. All data in state is stored as
+key-value pairs in leaves of this single tree. This tree is not a search tree —
+lookup is handled by separate index structures. It is a pure Merkle tree for
+integrity verification.
+
+The root hash of the state tree at the beginning of each block is published in
+the block footer as `start_of_block_state_root_hash`, making state provable from
+the block root.
```
mhess-swl:
```suggestion
the block root at the block's start time.
```

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+Used for: all six processed sub-trees (Historical Blocks, Consensus Headers,
+Input Items, Output Items, State Changes, Trace Data). The previous block root
+and root of state items are calculated, at least in part, using different
+algorithms.
+
+This algorithm enables the root hash of a Merkle tree to be computed
+incrementally as leaves become available, one at a time, without ever storing
+the entire tree in memory. Only O(log N) intermediate hashes need to be
+maintained between additions. The root hash can be queried at any point,
+including mid-stream, without interrupting the computation.
```
mhess-swl:
```suggestion
including mid-stream, without interrupting the computation. The entire incremental tree can also be reconstructed from a known set of hashes and leaf count at any point. 

The root hash of an empty incremental tree, consistent with the empty tree definition above, MUST also be `hash(0x00)`. 
```

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+
+![Block Root Tree Edge Cases](../assets/hip-XXXX/Hashing%20-%20Block%20Root%20Tree.png)
+
+
+## Backwards Compatibility
+
+This HIP specifies the hashing scheme for the block stream format introduced by
+HIP-1056; it does not alter how historical record files were hashed or signed —
+those retain their original RSA-signed SHA-384 digests and are wrapped as block
+stream blocks with a `SignedRecordFileProof`. The fixed leaf positions in the
```
mhess-swl:
```suggestion
stream blocks with a `SignedRecordFileProof` (see [HIP-1427](https://hips.hedera.com/hip/hip-1427) for more details). The fixed leaf positions in the
```

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+would need to include two separate paths for the timestamp and the item to be
+proven), increasing proof size for every proof ever produced. Making the
+timestamp a direct child of the block root eliminates most of this overhead.
+
+## Open Issues
+
+
+## References
+- [HIP-1056: Block Streams](https://hips.hedera.com/hip/hip-1056) — the parent HIP establishing
+  the block stream format.
```
mhess-swl, line 803: - [HIP-1427: Historical Record File Wrapping for Block Stream Cutover](https://hips.hedera.com/hip/hip-1427) - the HIP defining the block stream format for historical record files.

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+changes, the merkle tree of previous blocks, and the chain of previous blocks.
+</dd>
+<dt>Sub-tree</dt><dd>A self-contained Merkle tree whose root appears as one of
+the 16 fixed leaves of the block root tree.
+</dd>
+<dt>Streaming binary Merkle tree</dt><dd>An incremental algorithm that builds a
+Merkle tree root hash one leaf at a time while maintaining only O(log N) bytes
+of intermediate state.
+</dd>
+<dt>Complete balanced binary Merkle tree</dt><dd>A tree in which all levels are
```
mhess-swl:
```suggestion
<dt>Complete balanced binary Merkle tree</dt><dd>A Merkle tree in which all levels are
```

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+    }
+
+    /**
+     * Compute the Merkle tree root hash from the current state. This does not modify the internal state, so can be
+     * called at any time and more leaves can be added afterward.
+     *
+     * @return the Merkle tree root hash
+     */
+    public byte[] computeRootHash() {
+        byte[] merkleRootHash = hashList.getLast();
```
mhess-swl, line 593:
```suggestion
        if (hashList.isEmpty()) {
            return digest.digest(LEAF_PREFIX);
        }
        
        byte[] merkleRootHash = hashList.getLast();
```

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+created: 2026-03-11
+updated: 2026-03-12
+requires: 1056
+---
+
+## Abstract
+
+This HIP specifies the complete hashing scheme and Merkle tree structures used
+by the Hiero network to compute a verifiable, tamper-evident digest of all data
+in the block stream.
```
jasperpotts: This is how the hashing works for both block contents and state contents.
```suggestion
in the block and state.
```

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+
+Building on the block stream format introduced
+in [HIP-1056](https://hips.hedera.com/hip/hip-1056), this specification defines
```
jasperpotts, line 30:
```suggestion
in [HIP-1056](https://hips.hedera.com/hip/hip-1056), this specification defines the merkle tree structure of the block and how the block root hash is computed. This includes:
```

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+
+##### Reference Implementation (Java)
+
+```java
+/**
+ * A class that computes a Merkle tree root hash in a streaming fashion. It
+ * supports adding leaves one by one and computes the root hash without storing
+ * the entire tree in memory. It uses SHA-384 as the hashing algorithm and
+ * follows the prefixing scheme for leaves and internal nodes. Leaves can be
+ * optionally double-hashed before being added to the tree.
```
jasperpotts: There is no case of double hashing of leaves in the block tree as far as I know.
```suggestion
 * follows the prefixing scheme for leaves and internal nodes. 
```

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+    public void addLeaf(byte[] data) {
+        final long i = leafCount;
+        final byte[] e = hashLeaf(data);
+        hashList.add(e);
+        for (long n = i; (n & 1L) == 1; n >>= 1) {
+            final byte[] y = hashList.removeLast();
+            final byte[] x = hashList.removeLast();
+            hashList.add(hashInternalNode(x, y));
+        }
+        leafCount++;
```
jasperpotts:
```suggestion
    addNodeByHash(hashLeaf(data));
```

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+
+This HIP specifies the complete hashing scheme and Merkle tree structures used
+by the Hiero network to compute a verifiable, tamper-evident digest of all data
+in the block stream.
+
+Building on the block stream format introduced
+in [HIP-1056](https://hips.hedera.com/hip/hip-1056), this specification defines
+- A node-prefixing convention that provides domain separation between leaf and
+  internal nodes.
+- A "Merkle Mountain Top" structure that combines eight defined subtrees, eight
```
jasperpotts: Are we sure about "Merkle Mountain" / "Merkle Mountain Top" terminology? 
The HIP introduces these terms extensively (used throughout). The original document never uses "Mountain" — it refers to "Block Tree Root" and "Block Root fixed size Merkle Tree".

HIP/hip-1424.md
```diff
@@ -0,0 +1,594 @@
...
+type: Standards Track
+category: Core
+needs-hiero-approval: Yes
+needs-hedera-review: Yes
+status: Draft
+created: 2026-03-11
+updated: 2026-03-12
+requires: 1056
+---
+
```
jasperpotts, line 22: ### Missing Sections in the HIP vs Original Design Doc   
Are we going to document all these else where or should they be added here?                                                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                            
 - **Block Footer** — The design doc has a full section with the BlockFooter protobuf definition (fields: previous_block_root_hash, root_hash_of_all_block_hashes_tree, start_of_block_state_root_hash) and an important callout that the footer item is NOT included in any streaming block item merkle trees. The HIP only briefly mentions the footer in passing (line 233).  
                                                                                                                                                                                                                                                                                                                                                                            
 - **Block Proof** — The design doc has a full section describing BlockProof, TssSignedBlockProof, StateProof, SignedRecordFileProof, and RecordFileSignature protobuf messages. None of this is in the HIP. This includes the important detail that multiple proofs can exist per block and that proofs are always the last items after the footer.                             
                  
 -  **Filtered Block Items** — The design doc defines SubMerkleTree enum, FilteredSingleItem, and FilteredMerkleSubTree protobuf messages (including filtered_leaf_count). Entirely absent from the HIP.                                                                                                                                                                          
                  
 -  **RedactedItem** — The design doc defines the RedactedItem message with item_hash, signed_transaction_hash, and tree fields, and explains the distinction from filtered items (redacted items retain the signed transaction hash for event reconstruction). Not in the HIP.                                                                                                   
                  
 -  **MerkleLeaf Protobuf** — The design doc defines the MerkleLeaf, StateItem, StateKey, StateValue protobuf messages. The HIP references leaf bytes with (see below) at line 181 but never actually provides these definitions.                                                                                                                                                 
                  
 -  **Previous Block Hashes explanation** — The design doc has a dedicated section explaining why there are two chain types: (1) block-to-block chain for recent block proofs, and (2) streaming merkle tree of all hashes for concise historical proofs (e.g., 84M blocks back = 4GB via chain vs 1.3KB via tree). The HIP mentions both structures in the table but doesn't     
  explain the dual-chain rationale.

jasperpotts (changes requested):

HIP/hip-1424.md
jsync-swirlds: We might be clearer to replace "known set of hashes" with "complete set of hashes".
I don't think we can reconstruct the tree without _all_ of the hashes (unless we also require a starting state of the incremental hasher).

HIP/hip-1424.md
jsync-swirlds: The use of tree, root, and related was constantly confusing people.
We've had a lot better clarity and comprehension using the `Merkle Mountain` and `Merkle Mountain Top` terms.

We can switch back if we believe it's more correct, but if it's for comprehension, it seems that the "mountain" terms work better.

HIP/hip-1424.md
Nana-EC: Block structure and block Item definition should be defined and presented in HIP 1056, and may be referenced here.
**MerkleLeaf Protobuf** also probably makes more sense in HIP 1056.
I'll add the **Previous Block Hashes explanation** here though

Nana-EC: When explaining the State Root though I see the benefit of the State Item, Key and Value. Let me add those

HIP/hip-1424.md
```diff
@@ -0,0 +1,685 @@
...
+![Simplified Merkle Mountain Top](../assets/hip-1424/simplified-merkle-mountain-top.svg)
+
+The illustration above shows the simplified Merkle Mountain Top structure and
+highlights the key data nodes that describe a block's initial contents.
+
+The block root has two direct children:
+
+1. **Left child — Consensus Timestamp**: A leaf whose value is the block's consensus
+   timestamp, defined as the consensus timestamp of the first transaction in the
+   first round of that block.
```
mhess-swl: The consensus timestamp of the block can _also_ be defined by
* The first state change
* The consensus timestamp of an empty event
* The consensus timestamp of an empty round

Perhaps the wording could be '...timestamp, defined as the consensus timestamp of the first block item that advances consensus time' (or similar).

HIP/hip-1424.md
```diff
@@ -0,0 +1,685 @@
...
+2. **Right child — Merkle Mountain Top**: A fixed 16-leaf sub-tree whose leaves
+   are the root hashes of the distinct block item type sub-trees.
+
+The hashing logic for the initial 8 assigned leaves is illustrated below.
```
mhess-swl: We need to be careful designating any nodes as 'leaves'--leaves are _always_ hashed with a specific `0x0` prefix, whereas these sub-roots/sub-trees may or may not be leaves.

HIP/hip-1424.md
```diff
@@ -0,0 +1,685 @@
...
+defines all hash inputs, prefixes, tree shapes, and algorithms completely,
+ensuring that any compliant implementation produces byte-for-byte identical
+digests from the same data. That agreement is the prerequisite for threshold
+signatures and state proofs to be meaningful.
+
+## Rationale
+
+### SHA-384 as the Hash Function
+
+SHA-384 (a truncation of SHA-512) is used throughout this scheme. SHA-384 is the
```
jasperpotts: (a truncation of SHA-512) is only half correct, the starting bytes are different as well

HIP/hip-1424.md
```diff
@@ -0,0 +1,685 @@
...
+standard hash function for Hiero networks. The SHA-384 algorithm produces a
+48-byte digest, rather than the 32-byte digest of SHA-256 or Keccak-256 used by
+EVM chains. This larger output offers a comfortable security margin well above
+the 128-bit post-quantum security level that 256-bit hashes approach
```
jasperpotts: Maybe should make it clearer that SHA-384 is chosen to be post quantum. Can also say that signing of the chain can change over time but the hashing that glues the chain together has to be post-quantum from day one to be safe.

> SHA-384 is considered quantum-safe and is approved by the NSA for high-security applications in the post-quantum era, offering 192-bit preimage security and 128-bit collision resistance. Part of the SHA-2 family, it is robust against quantum attacks like Grover's algorithm due to its large 384-bit output.

https://pqshield.com/what-is-cnsa-2-0-the-nsas-post-quantum-roadmap/

HIP/hip-1424.md
```diff
@@ -0,0 +1,685 @@
...
+full except the lowest, and all leaves on the lowest level are positioned as far
+left as possible. Used for the state tree.
+</dd>
+<dt>Merkle Mountain Top</dt><dd>A fixed-shape tree with predetermined sub-tree
+positions, topped by the block root.
+</dd>
+</dl>
+
+## User Stories
+1. As a consensus node implementor, I need a precise specification of the
```
jasperpotts, line 155: What about user stories from the network user point of view?

- As a Dapp Developer, I want to be able to create concise merkle proof of any data in the block.
- As a Dapp Developer, I want to be able to filter a block to just a subset that I need while keeping cryptographic trust of the content.

Also would be nice to have a story for wanting to resign the chain. Hence why we added all blocks merkle tree.

- As a producer and consumer of the blocks I need to always be able to trust the blocks. As cryptography evolves like with post-quantum I need the complete history of blocks to be trusted with the latest safe signing algorithms.

HIP/hip-1424.md
```diff
@@ -0,0 +1,685 @@
...
+  }
+}
+
+message StateValue {
+  oneof value {
+    Account account = 1;
+    // ... additional value types
+  }
+}
+```
```
jsync-swirlds: This is obsolete.
We removed the "MerkleLeaf" concept a few months ago.

mgarbs (dismissed):

agents/proposal/hip-1424-image-updates.md
jsync-swirlds, line 1: You might want to remove the AI plan file, and perhaps instruct the AI to use `.agent` (which is in `.gitignore`) instead of `agent`.

assets/hip-1424/block-n-merkle-mountain-hashing.svg
jsync-swirlds, line 1: Note here, there is no difference between nodes 1-8 or 9-15.
The description under the first 8 nodes could be applied to all 16 without having to separate the treatment of reserved nodes from the allocated nodes.

assets/hip-1424/state-merkle-tree-hashing.png
jsync-swirlds: The state merkle tree still has single-child nodes, so it's important to still add the `0x01` case here.

assets/hip-1424/16-leaf-merkle-mountain-top.svg
mhess-swl, line 1: What do the colored backgrounds mean? Maybe worth defining in at least one of the graphics.

HIP/hip-1424.md
mhess-swl: Can resolve

HIP/hip-1424.md
mhess-swl: Can resolve

mhess-swl (approved):
Looks great! Thanks @Nana-EC!

PR #1427, opened 2026-03-12
Historical Record File Wrapping for Block Stream Cutover

Nana-EC:
**Description**:
Before the Hedera network transitions from record streams to block streams (as defined by [HIP-1193](https://hips.hedera.com/hip/hip-1193)),
all historical record files produced from network genesis through the cutover point must be converted into the block
stream format defined by [HIP-1056](https://hips.hedera.com/hip/hip-1056).
This conversion is performed offline as a one-time operation using a purpose-built CLI tool.
The process produces two artifacts:
- **jumpstart file** that the consensus node loads at first startup after the upgrade to resume block hashing from
the correct cryptographic state
- **verified collection of wrapped historical blocks** that Tier-1 block nodes use to backfill the complete Hedera
blockchain history from genesis.

This HIP specifies the conversion process, the format of both output artifacts, and the verification requirements that
the wrapped record blocks must satisfy before they are accepted into production.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


mgarbs (commented):
- assets/hip-1427/pipeline.png
- assets/hip-1427/wrap-pipeline.png
- assets/hip-1427/wrapped-block.png

Have a couple issues with the text bleeding over the boarders

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+
+### Consensus Node Integration
+
+The consensus node SHALL implement a jumpstart file loader that:
+
+1. Reads `jumpstart.bin` from a configurable path at startup.
+2. Uses the stored block number as the predecessor block number for the first live block.
+3. Initializes the streaming Merkle tree used for the all-blocks hash (`root_hash_of_all_block_hashes_tree`) with the
+   stored leaf count and intermediate hash list.
+4. Sets the `previous_block_root_hash` field of the first live block's `BlockFooter` after cutover to the stored 48-byte hash.
```
jsync-swirlds: This is incorrect.  It assumes a jumstart is magically available exactly at the point of cutover, which isn't true.
Jumpstart is loaded a full release _before_ cutover specifically because we cannot provide the jumpstart in a timely manner.


```suggestion
1. Stores partial WRB hash data for every record file created during a full release starting 2 releases before cutover.
2. Reads `jumpstart.bin` from a configurable path at startup 1 release before cutover.
3. Uses the jumpstart block number to find the correct starting point in the stored WRB partial hash data.
4. Initializes the streaming Merkle tree used for the all-blocks hash (`root_hash_of_all_block_hashes_tree`) with the
   stored leaf count and intermediate hash list.
5. Sets the `previous_block_root_hash` field of the next block's `BlockFooter` and the same value used in the WRB hash calculation from the Jumpstart file.
6. Catches up to "current" block before resuming regular operation.
7. Continues WRB hashing with complete information and maintains the previous block root hash and all blocks hashes tree from that block forward.
```

Key concepts:
1. Consensus must calculate WRB data independently for a full release starting at least `2` releases _before_ cutover
2. Consensus must use the block number in the jumpstart file to find the correct point from which to continue the WRB calculation forward with full data needed to calculate and maintain previous block root hash and the all blocks hashes tree.
3. Consensus must load the jumpstart at least `1` release _before_ cutover.
4. Consensus continues the WRB hash chain and all hashes tree after loading the jumpstart file for at least `1` release.

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+accumulates all block hashes. These three items are pre-computed by the wrapping process for the exact same reason the
+`BlockFooter` exists in the live block stream — they cannot be computed from data that hasn't been produced yet. Load
+the file at startup, initialize the three state variables from it, and the first live block will be correctly linked to
+the full historical chain.
+
+### For block node engineers
+
+Wrapped historical blocks are ordinary block stream blocks whose `BlockProof` uses the `SignedRecordFileProof` variant.
+The block node's verification path should already handle this variant (as defined in HIP-1424). Ingestion of wrapped
+record blocks via the historic file plugin follows the same path as any other block ingestion.
```
jsync-swirlds, line 581: I'm not sure the wrapping process is entirely compatible with the files-historic plugin.  Part of why we're looking at using backfill is to ensure that all of the blocks are processed fully rather than assuming the zip files are exactly as needed without modification.
We also need to have block nodes generating the full data for all ~100M hashes in time to be ready for cutover (part of why we need to be ready to deploy a month early; we're hoping we can ingest the WRB data fast enough to catch up fully in 30 days).

Also, Files-Historic doesn't have an ingestion path; so that cannot follow the same path as any other block ingestion (it bypasses validation and all other plugins).

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+./gradlew :tools:shadowJar
+java -jar tools-and-tests/tools/build/libs/tools-<version>-all.jar blocks wrap --help
+```
+
+The reference implementation SHALL be considered complete when the following have been delivered:
+
+- The `blocks wrap` command produces wrapped record blocks that pass all `blocks validate-wrapped` checks across the
+   full Hedera mainnet history from genesis to the block stream cutover block.
+- The `jumpstart.bin` produced from a full mainnet conversion is loaded and used by the consensus node in integration
+   tests to produce a correctly-linked first live block.
```
jsync-swirlds: There is no such thing as an entirely "full mainnet conversion" because mainnet keeps producing more blocks.
The jumpstart file is produced _up to a point_ and Consensus Node combines that with partial hash data for WRB files going back 30 days, then continues the WRB hash calculation going forward (storing the intermediate data in state after each record file is processed and written to disk).

It may help to review the 3-part process [here](https://github.com/hiero-ledger/hiero-block-node/blob/main/docs/block-node/Cutover-Process.md#how-does-the-consensus-node-catch-up-with-the-last-wrapped-record-block-hash).

Note, this three-part process is conducted at the beginning of release N+1 (the release _before_ cutover).

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+rejected because the ~30 TB record file corpus and the per-block RSA signature verification make real-time conversion
+infeasible within a maintenance window of a few hours.
+Pre-computing the wrapped record blocks offline eliminates this constraint entirely.
+
+### Deriving jumpstart state from the wrapped record blocks at consensus node startup
+
+Rather than a separate jumpstart file, the consensus node could derive its starting state by reading all wrapped record
+blocks and recomputing the streaming Merkle tree. This was rejected because it would require the consensus node to
+process billions of blocks before producing its first live block, delaying the network restart by potentially days or
+weeks. The jumpstart file reduces this to a single file read of under 2 KB.
```
jsync-swirlds:
```suggestion
Rather than a separate jumpstart file, the consensus node could derive its starting state by reading all wrapped record
blocks and recomputing the streaming Merkle tree. This was rejected because it would require the consensus node to
process millions of blocks before producing its next record file, delaying the network restart by potentially days or
weeks. The jumpstart file reduces this to a single file read of under 2 KB.
```

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+```
+
+The reference implementation SHALL be considered complete when the following have been delivered:
+
+- The `blocks wrap` command produces wrapped record blocks that pass all `blocks validate-wrapped` checks across the
+   full Hedera mainnet history from genesis to the block stream cutover block.
+- The `jumpstart.bin` produced from a full mainnet conversion is loaded and used by the consensus node in integration
+   tests to produce a correctly-linked first live block.
+- The wrapped record blocks are accepted and stored by the Tier-1 block node historic file plugin in integration tests.
+- The wrapped record blocks are independently verifiable by at least one third-party implementation.
```
jsync-swirlds, line 610: I think we need to add "mainnet and testnet cutover are complete" here.
Otherwise we might think the conversion is done when we still have a full month of live records to convert.

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+## How to Teach This
+
+### For consensus node engineers
+
+The jumpstart file is the handshake between the historical wrapping process and the live block producer. The consensus
+node needs three things from history: the last block number, the hash of the last block, and the Merkle tree state that
+accumulates all block hashes. These three items are pre-computed by the wrapping process for the exact same reason the
+`BlockFooter` exists in the live block stream — they cannot be computed from data that hasn't been produced yet. Load
+the file at startup, initialize the three state variables from it, and the first live block will be correctly linked to
+the full historical chain.
```
jsync-swirlds: This is not true, not least because the jumpstart is no less than 8 days behind live, so there are hundreds of thousands of WRB hashes to be added once the CN loads the file just to catch up with current, and there will be an entire additional release of WRB generation after that before the cutover.


```suggestion
The jumpstart file is the handshake between the historical wrapping process and the live block producer. The consensus
node needs two things from history: The hash of the last block, and the Merkle tree state that
accumulates all block hashes. These two items are pre-computed by the wrapping process so that the processing
of the history can be completed without interrupting continued operation of the network. The Consensus Node will load
the file at startup, use the block number to find which block in the previous 30 days corresponds to the hashes
in the file, compute the WRB root hashes for every block after that up to the current while maintaining
the hash chain and hash tree, and continue maintaining this data in state from that point forward.
This ensures that the first live block after cutover will be correctly linked to the full historical chain
and will have a correct values for the all blocks merkle tree.
```

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+   hosted for block node operators to download (e.g., a public GCP bucket, a CDN, or a peer-to-peer distribution
+   mechanism) are not specified here. This will be determined as part of the block node launch operations plan.
+
+3. **State Changes in wrapped record block structure**: The Wrapped record block structure currently omits State-Changes
+   In most cases this is correct but likely block 0 and possible amendment captures map require state changes for visibility.
+   Need to confirm this.
+
+4. **wrappedBlocks/<range>.zip format**: Need to confirm what format the range value takes.
+
+## References
```
jsync-swirlds, line 655: We should probably include a reference to the cutover process, since that describes how the jumpstart file is used and also covers many of the process changes.

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+---
+
+## Backwards Compatibility
+
+This HIP defines a process that runs entirely before and independently of the live network. It does not change any
+on-chain transaction processing, any existing API, or any currently-running network component. All changes are
+additive: a new binary artifact (the jumpstart file) and a new set of block stream files are produced.
+
+Historical record files remain available in their original format and location in cloud storage after the conversion;
+the wrapping process is read-only with respect to the source bucket.
```
jsync-swirlds: I don't know if we guarantee this indefinitely.
We need to archive this data to cold storage at some point inorder to realize cost savings.

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+the encoded hash count field) and SHOULD refuse to start if the file is malformed.
+
+---
+
+### Block Node Integration
+
+Tier-1 block nodes SHALL ingest the wrapped record blocks using the block node's historic file persistence plugin.
+The wrapped record blocks are compatible with the block node's standard block ingestion path. Block nodes SHALL verify
+each wrapped record block as it is ingested; if any block fails hash chain verification it SHALL be rejected and the
+ingestion operator SHALL be notified.
```
jsync-swirlds: Not sure why we're claiming the files-historic plugin (which only stores data) is somehow going to _ingest_ blocks.
Much more likely backfill plugin will handle the ingestion.

Only one block node will have the data "placed" in the storage without it being ingested at all.

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+
+1. **Hash chain continuity**: each block's `previous_block_root_hash` in the `BlockFooter` equals the SHA-384 root hash
+   computed from the preceding block. The first block (block 0) SHALL have 48 zero bytes as its previous hash.
+2. **Historical block hash Merkle tree root**: the `root_hash_of_all_block_hashes_tree` in each block's `BlockFooter`
+   equals the root hash that the streaming Merkle tree algorithm produces after incorporating all prior block hashes in
+   order.
+3. **Required item presence**: every wrapped record block SHALL contain at least one `BlockHeader`, one
+   `RecordStreamFile`, one `BlockFooter`, and one `BlockProof`.
+4. **Item ordering**: items SHALL appear in the order: `BlockHeader`, optional `StateChanges`, `RecordStreamFile`,
+   `BlockFooter`, one or more `BlockProof` items, with no duplicates or misplaced items.
```
jsync-swirlds: This is not the correct order.
State Changes come _after_ output items.  `RecordStreamFile` is an output item.

```suggestion
4. **Item ordering**: items SHALL appear in the order: `BlockHeader`, `RecordStreamFile`, optional `StateChanges`,
   `BlockFooter`, one or more `BlockProof` items, with no duplicates or misplaced items.
```

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+
+The consensus node SHALL read this file at first startup after an upgrade (prior to cutover) and use its contents as follows:
+
+- The block number is incremented by one to produce the block number of the next wrapped record block.
+- The 48-byte root hash becomes the `previous_block_root_hash` field in the `BlockFooter` of the next wrapped record block.
+- The leaf count and hash list are used to initialize the streaming Merkle tree so that the future first live block's
+   `root_hash_of_all_block_hashes_tree` correctly incorporates all historical block root hashes.
+
+The maximum possible size of the jumpstart file is `68 + 48 × ceil(log₂(N))` bytes, where `N` is the total number of
+historical blocks. For one billion historical blocks this is approximately `68 + 48 × 30 = 1,508 bytes`.
```
jsync-swirlds: 1. We will have roughly 100 million WRBs, not 1 billion.
2. The actual number of hashes is _up to_ log₂(N), but might be as little as 0. It all depends on the exact number of leaves and how many "orphan" inner nodes there are in the tree.

```suggestion
historical blocks. For one hundred million historical blocks this is approximately `68 + 48 × 27 = 1,364 bytes`.
```

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+| `addressBookHistory.json` | Chronological address book history, used for signature verification |
+| `blockStreamBlockHashes.bin` | Binary registry of all computed block root hashes, indexed by block number |
+| `streamingMerkleTree.bin` | Checkpoint of the streaming Merkle tree state for resume |
+| `completeMerkleTree.bin` | Checkpoint of the complete balanced binary tree state for resume |
+| `jumpstart.bin` | The jumpstart file for the consensus node (see §4.4) |
+
+#### 4.4 Jumpstart file format
+
+The jumpstart file (`jumpstart.bin`) is a binary file written once at the end of a successful conversion run and
+updated incrementally on graceful shutdown via a JVM shutdown hook. Its format is:
```
jsync-swirlds: This should be written frequently once we have live blocks, perhaps hourly.
If we wait for a batch run, we'll be even further behind than we must be.

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+
+1. `BlockHeader` — identifies the block number and carries metadata.
+2. `RecordStreamFile` — carries the raw bytes of the original record file and all RSA-verified signature files from
+   consensus nodes. The record file bytes are preserved exactly as produced by the consensus nodes; no byte is added or removed.
+3. `BlockFooter` — carries the three pre-computed hash values required to complete the block root Merkle tree: (a) the
+   SHA-384 root hash of the immediately preceding block, (b) the root hash of the all-blocks streaming Merkle tree up
+   to and including the previous block, and (c) the state root hash at the start of the block (zero for genesis and
+   historical blocks where state hash is not available).
+4. `BlockProof` (one or more) — contains a `SignedRecordFileProof` carrying the RSA signatures collected from consensus
+   nodes and the record stream format version. At least one-third of the node count's RSA signatures SHALL be present.
```
jsync-swirlds: Do we want to mention the optional additional proofs?

```suggestion
4. `BlockProof` (one or more) — contains at least one `SignedRecordFileProof` carrying the RSA signatures
   collected from consensus nodes and the record stream format version. At least one-third of the node
   count's RSA signatures SHALL be present. Additional 'BlockProof' entries (possibly including `StateProof`
   or `TSSSignedBlock` types) MAY be present.
```

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+
+#### 4.2 Wrapped record block structure
+
+Each wrapped record block SHALL contain the following items in the following order:
+
+![Wrapped Block Structure](../assets/hip-1427/wrapped-block.png)
+
+1. `BlockHeader` — identifies the block number and carries metadata.
+2. `RecordStreamFile` — carries the raw bytes of the original record file and all RSA-verified signature files from
+   consensus nodes. The record file bytes are preserved exactly as produced by the consensus nodes; no byte is added or removed.
```
jsync-swirlds: The RSA signature bytes are in the proof, not the block item.

```suggestion
2. `RecordStreamFile` — carries the raw bytes of the original record file content. The record file bytes are preserved exactly as signed by the consensus nodes; no byte is added or removed.
```

Note, also, the bytes are as _signed_, which is slightly different from the _file_ bytes.

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+each with its own internal structure and hashing algorithm. The wrapping tool handles all three versions transparently
+during conversion, applying the appropriate hash computation for each version when building the block proof.
+
+## User Stories
+
+### Consensus Node Operator
+
+1. **As a consensus node operator**, I need the jumpstart file to be generated and deployed to each consensus node before
+   the upgrade so that the first block produced after the cutover carries the correct previous block root hash and
+   Historical block hash Merkle tree root, maintaining unbroken cryptographic continuity with the historical record file chain.
```
jsync-swirlds:
```suggestion
1. **As a consensus node operator**, I need the jumpstart file to be generated and deployed to each consensus node a
   full release before the cutover so that I can catch up to and maintain the current previous block root hash
   and Historical block hash Merkle tree root prior to the cutover upgrade, maintaining unbroken
   cryptographic continuity with the historical record file chain.
```

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+means 30+ copies of every block exist in cloud storage. The conversion tool selects one canonical record file per block
+number (the first valid file found), collects all available signature files (which together provide the RSA
+multi-signature proof), and discards the redundant duplicate record files. The result is a deduplicated collection in
+which each block is represented exactly once by its content, accompanied by all available signatures.
+
+### Versions of record file format
+
+The historical record files exist in three distinct format versions (v2, v5, and [v6](https://hips.hedera.com/hip/hip-435)),
+each with its own internal structure and hashing algorithm. The wrapping tool handles all three versions transparently
+during conversion, applying the appropriate hash computation for each version when building the block proof.
```
jsync-swirlds: As I understand it, the WRBs all share a single hashing process (nitpick: it's a process, not an algorithm, the algorithm is always SHA2-384) because we store the data as signed rather than the file on disk.

If that understanding is correct (hopefully @jasperpotts can correct me if I'm incorrect), then perhaps rewrite as follows?

```suggestion
The historical record files exist in three distinct format versions (v2, v5, and [v6](https://hips.hedera.com/hip/hip-435)),
each with its own internal structure and hashing process. The wrapping tool handles all three versions transparently
during conversion, ensuring that a single consistent hashing process can be applied to the wrapped blocks.
```

rockysingh (commented):
## Review — Comparing HIP against actual implementation

### Accuracy Issues

**1. Command name mismatch — `validate-wrapped` vs `validate`**
Sections 5.2 and throughout reference `blocks validate-wrapped`, but the actual command is `blocks validate`.

**2. Output files table (§4.3) — `completeMerkleTree.bin` doesn't exist**
The table lists `completeMerkleTree.bin` as an output. The wrap command does **not** produce this file — it's legacy and the code actually deletes it if found. Only `streamingMerkleTree.bin` is produced.

**3. Output files table (§4.3) — missing `wrap-commit.bin`**
The durable watermark file (`wrap-commit.bin`) is not listed. This is a key output that tracks the highest block number successfully written and enables crash recovery.

**4. Wrapped block structure (§4.2) — item 2 should be `RecordFileItem`, not `RecordStreamFile`**
The block contains a `RecordFileItem` (which wraps `RecordStreamFile` along with `creation_time`, `sidecar_file_contents`, and `amendments`), not a raw `RecordStreamFile` directly.

**5. Signature threshold — "at least one-third" vs "1/3 + 1"**
Section 4.2 says "At least one-third of the node count's RSA signatures SHALL be present." The actual implementation requires **1/3 + 1** signatures (strict supermajority of one-third).

**6. Phase 2.3 references only `download-days-v2`**
The HIP specifies `download-days-v2` (gRPC). The current recommended/most stable version is `download-days-v3` (HTTP transport, avoids virtual thread pinning). Worth updating or at least mentioning both.

**7. Validation (5.2) is incomplete — only 5 of 11 validations listed**
The HIP lists 5 validations. The actual implementation performs 11:

Missing from the HIP:
- **RSA Signature validation** — verifies 1/3+1 node signatures per block
- **Address Book Update validation** — discovers address book changes from file 0.0.102 transactions
- **Hash Registry validation** — cross-checks block hashes against `blockStreamBlockHashes.bin`
- **Streaming Merkle Tree validation** — verifies `streamingMerkleTree.bin` matches freshly-computed tree
- **Jumpstart validation** — verifies `jumpstart.bin` contents match computed values
- **Balance Checkpoint validation** — verifies account balances against pre-fetched snapshots at periodic intervals (the current "50B HBAR supply" bullet partially covers this but they're actually two separate validations: running supply check and periodic checkpoint comparison)

---

### Open Issues That Can Be Resolved

**Open Issue 3 — StateChanges**: The implementation **does** include `StateChanges` in block 0 and for amendment blocks. This is handled and could be documented as resolved.

**Open Issue 4 — Zip range format**: The format is `{digit}{0000}s.zip` within a hierarchical directory structure. Block numbers are 19-digit zero-padded, every 3 digits creates one directory level. Example: block 1,234,567 → `000/000/001/234/50000s.zip`. This is known and can be documented.

---

### Minor / Suggestions

- **Phase 1** could mention `mirror update` / `mirror extractDayBlocksFromApi` as alternative approaches that fetch directly from the Mirror Node REST API (instead of requiring CSV dump download)
- **Phase 2.1** could mention `days updateDayListings` as an alternative to `rclone lsjson` + `split-files-listing`
- **Phase 4.1 Stage 2** mentions "the in-memory complete balanced binary tree hasher" — this is legacy and no longer exists in the implementation. Only the streaming Merkle tree is maintained.

HIP/hip-1427.md
```diff
@@ -0,0 +1,596 @@
...
+
+**Record file block**: A logical unit of blockchain history represented by one record file from one node, together with
+all available signature files from all nodes and all sidecar files for that record file. In normal operation all nodes
+produce byte-identical record files, so a single record file plus N signature files constitutes one record file block.
+
+**Wrapped record block (WRB)**: A block stream `Block` protobuf containing a `BlockHeader`, a `RecordStreamFile` item
+carrying the raw record file bytes and all RSA signature files, a `BlockFooter` carrying the three pre-computed hash
+values, and a `BlockProof` of type `SignedRecordFileProof` carrying the RSA signatures.
+
+**Jumpstart file**: A small binary file carrying the last block number, the SHA-384 root hash of the last wrapped
```
derektriley: This indicates the file is handed to each consensus node for loading ahead of the cutover upgrade.

We likely want to reword sections around the jumpstart file, to be more specific in that the jumpstart file will be created, but then a File Update to entity 121 with specific network properties will be done before the software upgrade.

```
blockStream.jumpstart.blockNum
blockStream.jumpstart.previousWrappedRecordBlockHash
blockStream.jumpstart.streamingHasherLeafCount
blockStream.jumpstart.streamingHasherHashCount
blockStream.jumpstart.streamingHasherSubtreeHashes
```

HIP/hip-1427.md
xin-hedera: Does the tool put all node signatures or just enough to meet the at least 1/3 threhold?

Problem is at a certain release, we switched to at least 1/3 consensus weight and it's possible given 1/3 of the node count's RSA signatures, the consensus weight might be less than 1/3 and fail verification.

HIP/hip-1427.md
Nana-EC: If it no longer matches then it diverted as the  CLI intent calls this out.
@rockysingh can you confirm?

HIP/hip-1427.md
Nana-EC: @rockysingh will be updating the tool to manage the transition you noted. I'll update this also

HIP/hip-1427.md
```diff
@@ -0,0 +1,625 @@
...
+
+#### Upgrade Timeline
+
+Assuming a cutover release at time T, a potential timeline would be
+
+| Milestone    | Action |
+|--------------|--------|
+| T−2 release  | Consensus node begins storing partial WRB hash data
+| T−1 release  | Jumpstart file distributed; consensus node reads it on startup
+| T (cutover)  | Wrapped blocks may be staged; network upgrade proceeds
```
jsync-swirlds: This is slightly off...

```suggestion
| Milestone    | Action |
|--------------|--------|
| T−2 release  | Consensus node begins storing partial WRB hash data.
| T−1 release  | Jumpstart file distributed; consensus node reads it on startup<br/>and begins creating _complete_ wrapped blocks. From this point<br/>wrapped blocks may be uploaded or distributed.
| T+ (cutover)  | Network upgrade to full block stream and TSS.
```

HIP/hip-1427.md
jsync-swirlds: Regardless of match, files-historic lacks file ingestion capability.
We will need something to deliver the WRBs to the files-historic plugin one way or another, as we cannot add them without somehow informing the plugin that the new files exist (even if it's just adding a few tens of thousands of zip files in place before starting the block node).

Live WRB blocks still need to be sent via some other dynamic pathway. Otherwise we'd need to have the plugin scanning the filesystem every second, and that's not something we can afford in production code.

rockysingh: The wrap CLI produces valid wrapped blocks in zip archives with bock proofs, block hash registry, streaming merkle tree state, and jumpstart data. The output is structurally correct and fully validated (we're currently running mainnet validation through 92M+ blocks with stake-weighted signature verification).                                                                               
                                                                                                                                                                                                                 However, @jsync-swirlds  is correct that files-historic does not have a file ingestion path today. The plugin receives blocks through the block node messaging facilit. I think there's no mechanism to import pre-existing zip files from disk. The wrap command's zip layout also uses a different directory/path structure than what files-historic produces internally.

jsync-swirlds: The folder/path structure is probably the most important difference.
Also, we need to ensure the compression and structure are compatible.

Alternatively, we could just deliver all of the history via the same pathway used for live blocks. The simplest mechanism might be to use the `publish` API via localhost, which would avoid needing to add any extra support infrastructure.

HIP/hip-1427.md
```diff
@@ -0,0 +1,625 @@
...
+The jumpstart file is a critical input to the first live block. Its 48-byte hash and Merkle tree state MUST be
+distributed to all consensus nodes with the same integrity guarantees as the node software itself (e.g., via signed
+release packages). A tampered jumpstart file would cause consensus nodes to diverge on the Historical block hash Merkle
+tree root in the first live block, preventing threshold signature aggregation.
+
+### Requester-Pays Access Control
+
+The record stream bucket uses requester-pays access. Only parties with configured GCP access can download the source
+data. This provides a natural barrier against unsanctioned bulk downloads but is not a substitute for verifying the
+integrity of downloaded data against the Mirror Node's known record file hashes.
```
jsync-swirlds: Note from meeting, we should remove this paragraph.

HIP/hip-1427.md
```diff
@@ -0,0 +1,670 @@
...
+
+This on-chain distribution mechanism means the
+contents carry the same governance weight as any other network file update.
+
+The consensus node SHALL implement a jumpstart contents loader that:
+
+1. Starting two full releases before cutover, stores partial WRB hash data for every record file produced.
+2. One release before cutover, reads `jumpstart.bin` from a configurable path at startup.
+3. Verifies that the SHA-384 root hash in the jumpstart contents matches the hash computed from the stored WRB
+   partial hash data at the indicated block number.
```
jsync-swirlds, line 466: (correction)
This step has to happen _after_ step (5), as we don't have the root hash to compare with until we calculate it using the jumpstart data.

HIP/hip-1427.md
```diff
@@ -0,0 +1,670 @@
...
+the encoded hash count field). If the contents are malformed or the hash verification in step 3 fails, the consensus
+node SHALL log a detailed error and abort the jumpstart initialization — but SHALL NOT halt startup entirely. In that
+case the node will continue to operate without the historical WRB hash chain linked, and the mismatch MUST be
+investigated and resolved before the cutover upgrade is applied. Operators are expected to retry with corrected
+jumpstart contents once the source of the mismatch is identified.
+
+> **Note on hash mismatch retry**: A mismatch most commonly indicates that the jumpstart contents were generated
+> from a different version of the record file corpus than the partial WRB hash data stored by this node. The operator
+> should verify that the distributed `jumpstart.bin` corresponds to the same conversion run as the node's stored
+> hashes, then re-deploy the correct contents via a `FileUpdate`.
```
jsync-swirlds, line 486: I think we should leave this paragraph out.

1. The jumpstart content is in network state, so it's the same on all nodes.
1. A file update modifies all nodes, not just one; and it's a council-signed update of file 121; not something an operator can modify.

The hash data that results from applying the jumpstart data is also stored in state.
If only one node has a different value somehow; the node will ISS and resolution will be to reconnect.

If the whole network gets the wrong value, then we will fallback as indicated in the paragraphs above.

HIP/hip-1427.md
```diff
@@ -0,0 +1,670 @@
...
+
+---
+
+#### Block Node Integration
+
+Tier-1 block nodes SHOULD ingest the wrapped record blocks using the block node's backfill plugin in order to
+serve complete historical data to clients. However, persisting the full historical corpus from the stream starting
+point is a desired property, not a hard requirement: it is unlikely that every Tier-1 operator — particularly
+Linux Foundation Hiero (LFH) member-run nodes — will retain the complete history indefinitely, and this cannot
```
jsync-swirlds, line 495: I have no idea where this interpretation of LFH came from; guessing it's an LLM hallucination.
```suggestion
Local Full History (LFH) member-run nodes — will retain the complete history indefinitely, and this cannot
```

HIP/hip-1427.md
```diff
@@ -0,0 +1,670 @@
...
+
+| Offset | Length | Field |
+|:---:|:---:|---|
+| 0 | 8 bytes | Block number of the last wrapped record block (big-endian `long`) |
+| 8 | 48 bytes | SHA-384 root hash of the last wrapped record block |
+| 56 | 8 bytes | All-blocks streaming Merkle tree leaf count (big-endian `long`) |
+| 64 | 4 bytes | Number of intermediate hash entries in the streaming Merkle tree (big-endian `int`) |
+| 68 | 48 × N bytes | Intermediate hash list of the streaming Merkle tree (N entries of 48 bytes each) |
+
+The consensus node SHALL read `jumpstart.bin` at first startup after an upgrade (prior to cutover) and use the jumpstart contents as follows:
```
jsync-swirlds, line 367:
```suggestion
The consensus node SHALL read jumpstart configuration at first startup after an upgrade (prior to cutover) and use the jumpstart contents as follows:
```

jsync-swirlds (approved):
This looks good; a few small adjustments, but generally good.

mgarbs (approved):

mgarbs (approved):

PR #1446, opened 2026-04-06
HIP-1446: Multiple other-parents

lpetrovic05:
**Description**:

This HIP proposes extending the Hiero hashgraph consensus algorithm to allow events to
reference multiple other-parents instead of being limited to a single other-parent.

### DIFF
--- a/HIP/hip-1399.md
+++ b/HIP/hip-1399.md
@@ -11,7 +11,7 @@ needs-hiero-approval: Yes
 needs-hedera-review: Yes
 status: Draft
 created: 2026-04-02
-updated: 2026-04-02
+updated: 2026-06-02
 ---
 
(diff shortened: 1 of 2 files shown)


Mark-Swirlds:
Just a few minor suggestions:
* "Internal only" section - mention the small impact on block stream size vs. reduced message quantity
* Metrics - Any metrics to add?  Average Other-Parent Count Per Event to measure impact?

lpetrovic05:
> Just a few minor suggestions:
> 
> * "Internal only" section - mention the small impact on block stream size vs. reduced message quantity
> * Metrics - Any metrics to add?  Average Other-Parent Count Per Event to measure impact?

done

mgarbs (approved):

PR #1448, opened 2026-04-07
HIP-1448 Broadcast mechanism for gossip

abies:
**Description**:

This HIP introduces a broadcast mechanism for event propagation in the Hiero
consensus network. When a node creates a new self-event, it immediately sends
that event directly to all connected peer nodes (neighbours), without waiting
for the next gossip sync cycle.

### DIFF
--- a/HIP/hip-1448.md
+++ b/HIP/hip-1448.md
@@ -8,9 +8,10 @@ type: Standards Track
 category: Core
 needs-hiero-approval: Yes
 needs-hedera-review: Yes
-status: Review
+status: Last Call
+last-call-date-time: 2026-05-20T07:00:00Z
 created: 2025-10-01
-updated: 2026-04-07
+updated: 2026-05-13
 release:
 ---
 
--- a/HIP/hip-1448.md
+++ b/HIP/hip-1448.md
@@ -1,0 +1,230 @@
+---
+hip: 1448
+title: Simple Event Broadcast
+author: Artur Biesiadowski (@abies)
+working-group: Artur Biesiadowski (@abies)
+discussions-to: https://github.com/hiero-ledger/hiero-improvement-proposals/pull/1448
+type: Standards Track
+category: Core
+needs-hiero-approval: Yes
+needs-hedera-review: Yes
+status: Draft
+created: 2025-10-01
+updated: 2026-04-07
+release:
+---
+
+## Abstract
+
+This HIP introduces a broadcast mechanism for event propagation in the Hiero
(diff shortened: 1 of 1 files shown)


Mark-Swirlds:
I noticed the statement - _For super-low latency networks, C2C latency decreases by factor of 2x to 3x._

Could you expand on the meaning, what's a super-low latency network?  Solo or Spheres?  Why is this important and why is mainnet so modest compared to these gains?

Mark-Swirlds:
_No new security implications are introduced. The broadcast connection is already
established and authenticated for gossip purposes. Events sent via broadcast are
serialized, signed, and verified in exactly the same manner as events received
through the sync protocol. No new authentication surface is created._

How are DDoS prevention handled if a node is compromised.  Can the broadcast be used to overwhelm gossip communications?

Mark-Swirlds:
Can we add metrics for KPIs and problem isolation?  For example:

- Broadcast vs. Gossip Event Ratio - For each event received by a node, was it first seen via broadcast or sync?

abies:
> Could you expand on the meaning, what's a super-low latency network? 

Super-low latency networks in this context are ones where ping between the nodes is <1ms (as opposed to hundreds of ms in a mainnet-like environment). 
Maybe I should expand - true benefit (2-3x) was seen in low-latency networks with tuned event creation. Default settings have huge event creation periods, which make it impossible to fully benefit from faster distribution of events. At the same time, decreasing that period on high ping environments is counterproductive. Low-latency, short-event-creation-period setups see the biggest benefit, as they are able to create events often enough to utilize them being available and at the same time, they are able to receive them over broadcast fast enough to not stall event creation.

There might be more factors at play here, we will be investigating it further as part of multiple other parents' HIP.

>How are DDoS prevention handled if a node is compromised? Can the broadcast be used to overwhelm gossip communications?

It has the same behaviour as sync in the case of DOS. Please contact me privately for details; it is not the right forum to discuss the details.

>Broadcast vs. Gossip Event Ratio - For each event received by a node, was it first seen via broadcast or sync?

Most of the events will be sent only by broadcast, with the well behaved network, every event in sync will be a duplicate (so broadcast will show 100% regardless if sync is sending zero or many events). In a bad network, broadcast is disabled, so sync will be at 100%. There will be a very small window of possible environments where that number will be anywhere in the middle.

We can add it, but I'm not sure if it is necessarily a part of this HIP, rather than a possible extension. We have a good number of metrics to help with debugging around that area (amount of broadcast events sent, received, events per sync, etc)

mgarbs (approved):

PR #1474, opened 2026-05-08
Update HIP 1056 Block Streams with latest block items and proofs

Nana-EC:
## Description
This update brings HIP-1056 (Block Streams) in line with the current implementation of
block items, the block Merkle tree, and the three forms of `BlockProof`. The changes are
entirely specification and documentation; no on-chain behaviour is altered.

---

### 1. `BlockItem` type table — expanded from 10 to 13 types

The `BlockItem` message now formally defines 13 `oneof` cases with a routing-category
table covering each item's role (Input, Output, Consensus, State, Trace, or not-hashed).
Two new types are specified:

- **`BlockFooter` (field 12)** — carries the three leaf values needed to assemble the
  block Merkle tree (`previousBlockRootHash`, `rootHashOfAllPreviousBlockHashes`,
  `startOfBlockStateRootHash`). Precedes `BlockProof` in every block.
- **`RedactedItem` (field 19)** — placeholder for a fully redacted item in
  privacy-preserving streams.

The `reserved` field list is updated accordingly (fields 13–18 remain reserved; field 12
is now assigned).

---

### 2. Item routing to five streaming hashers — now normative

The five streaming hashers (`inputTreeHasher`, `outputTreeHasher`,
`consensusHeaderHasher`, `stateChangesHasher`, `traceDataHasher`) and their item-type
routing rules are now documented in a normative table. The section clarifies that routing
is determined by **item type**, not field number, and explains what "not hashed" means
for `BlockProof`, `BlockFooter`, `FilteredItemHash`, `RedactedItem`, and `RecordFileItem`.

---

### 3. Merkle tree topology — fully specified

The block Merkle tree is now documented as a **depth-5 binary tree with eight fixed
leaves**, with:

- Leaf-index table (0–7) mapping each leaf to its source (`BlockFooter` vs. streaming
  hasher) and its semantic meaning.
- SHA-384 domain-separated prefix rules (`0x00` leaf, `0x01` single-child, `0x02`
  two-child).
- An EVM-compatibility note explaining why `stateRootHash` captures the state at the
  **start** of block N (not the end) and why this requires state proofs for changes
  within block N to reference block N+1's footer.

---

### 4. `TraceData` block item — new section

A dedicated section specifies the `TraceData` item type: its purpose (EVM execution logs,
call traces, gas diagnostics), its hasher assignment (`traceDataHasher`, leaf index 7),
and its optionality (zero or more per block). Includes the protobuf definition.

---

### 5. `FilteredItemHash` — expanded with usage guide

The `FilteredItemHash` section is rewritten to explain the privacy-preserving filtered
stream use case end-to-end: how a filtered stream is produced, how a downstream verifier
can still validate the block proof using only the hash stubs, and how full reconstruction
is possible if the original items are later obtained. Field comments are clarified.

---

### 6. `BlockProof` — three forms now formally specified

`BlockProof` now documents three distinct forms:

| Form | When Used | Mechanism |
|------|-----------|-----------|
| **`TssSignedBlockProof`** | Genesis, cutover, key rotation | Direct TSS-BLS aggregate signature over block root hash |
| **State Proof (WRAPS)** | Steady-state when signature fragments arrive late | Sibling-hash Merkle path from block M's root to a later signed block N |
| **`SignedRecordFileProof`** | Historical wrapped record blocks (see HIP-1427) | RSA signatures over the original record file hash, verified against the address book at time of block |

The `TssSignedBlockProof` message is introduced as a new protobuf message carrying the
aggregate BLS signature bytes. The `BlockProof` protobuf fields are updated: fields 2
and 3 (`previous_block_root_hash`, `start_of_block_state_root_hash`) are marked
`deprecated` as their values are now canonical in `BlockFooter`; field 2 is replaced
with `TssSignedBlockProof signed_block_proof`.

---

### 7. `BlockFooter` — new protobuf definition and section

`BlockFooter` is now a first-class citizen of the specification with its own section and
protobuf definition, explaining each of its three fields and the timing rationale for
`start_of_block_state_root_hash`.

---

### 8. Rejected Ideas — three new entries

Added three new Rejected Ideas entries explaining design decisions:

- **Start-of-block vs. end-of-block state root** — EVM compatibility and performance
  rationale.
- **Two forms of `TssSignedBlockProof`** — why both aggregate-signature and WRAPS
  (chained sibling-hash) forms are needed rather than a single form.
- **Five independent streaming hashers** — rationale for type-separated subtrees over a
  single flat tree.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-1056.md
+++ b/HIP/hip-1056.md
@@ -422,7 +422,7 @@ all implementations:
 | 5 | `transaction_result` | Output | Execution result for the preceding signed transaction |
 | 6 | `transaction_output` | Output | Supplementary output produced by transaction execution |
 | 7 | `state_changes` | State | Ledger state mutations (account, file, contract, etc.) |
-| 8 | `filtered_item_hash` | (variable) | SHA-384 hash of a redacted item (privacy-preserving streams) |
+| 8 | `filtered_single_item` | (variable) | In-stream replacement hash for a single filtered item (privacy-preserving streams) |
 | 9 | `block_proof` | (Not hashed) | Cryptographic proof of block integrity; closes the block |
 | 10 | `record_file` | (Not hashed) | Legacy V6 record-stream file carried for historical compatibility |
 | 11 | `trace_data` | Trace | EVM trace or diagnostic data |
@@ -435,7 +435,7 @@ Items marked "(Not hashed)" are parsed for metadata and control flow but do not 
 ```protobuf
 message BlockItem {
     // Reserved for future items that require separate handling for block hash purposes.
-    reserved 13,14,15,16,17,18;
+    reserved 13,14,15,16,17,18,20;
 
     oneof item {
         /**
@@ -517,7 +517,7 @@ message BlockItem {
 ```
 ##### Item Routing to Streaming Hashers
 
-All items in the block other than the final `BlockProof`, `BlockFooter`, a `RecordFileItem`, `FilteredItemHash`, or
+All items in the block other than the final `BlockProof`, `BlockFooter`, a `RecordFileItem`, `FilteredSingleItem`, or
 `RedactedItem` are assigned to one of five streaming Merkle hashers. This separation is important because items are
 hashed into five independent sub-merkle-trees that are combined to form the final block root hash.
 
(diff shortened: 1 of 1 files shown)


HIP/hip-1056.md
```diff
@@ -406,13 +406,36 @@ considerations, in addition to the end of a `Block` or `Transaction`.
...
+| 1 | `block_header` | Output | Opens a block; carries block number, timestamp, hash algorithm |
+| 2 | `event_header` | Consensus | Opens a consensus event within a round |
+| 3 | `round_header` | Consensus | Opens a consensus round |
+| 4 | `signed_transaction` | Input | A single transaction submitted by a client or system |
+| 5 | `transaction_result` | Output | Execution result for the preceding signed transaction |
+| 6 | `transaction_output` | Output | Supplementary output produced by transaction execution |
+| 7 | `state_changes` | State | Ledger state mutations (account, file, contract, etc.) |
+| 8 | `filtered_single_item` | (variable) | In-stream replacement hash for a single filtered item (privacy-preserving streams) |
+| 9 | `block_proof` | (Not hashed) | Cryptographic proof of block integrity; closes the block |
+| 10 | `record_file` | (Not hashed) | Legacy V6 record-stream file carried for historical compatibility |
```
Nana-EC, line 427: Change to output - https://github.com/hiero-ledger/hiero-consensus-node/blob/main/hapi/hedera-protobuf-java-api/src/main/proto/block/stream/block_item.proto#L130

PR #1483, opened 2026-05-28
HIP-1483: New debug and trace APIs on Mirror node side

IvanKavaldzhiev:
**Description**:
Add a HIP definition for some new debug and trace APIs that are going to be added on mirror-node side. In addition some new tracers - prestateTracer and keccak256PreimageTracer will be added to the existing debug endpoint and the newly added debug_traceCall.

Some performance and compatibility optimisations will be added as well:

- the receipts_root field will be now calculated directly on ingestion, instead of being manually calculated on JSON-RPC relay side
- all weibar values used in Ethereum transactions will be saved directly to DB, without conversion to tinybars, so that they don't loose precision.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/HIP/hip-1483.md
+++ b/HIP/hip-1483.md
@@ -13,6 +13,7 @@ hedera-acceptance-decision:
 status: Review
 last-call-date-time: 2026-03-17T07:00:00Z
 created: 2026-05-28
+updated: 2026-06-02
 discussions-to: https://github.com/hiero-ledger/hiero-improvement-proposals/discussions/1484
 ---
 
@@ -41,9 +42,9 @@ The existing `contracts/call` endpoint in Mirror Node will be enhanced via a `st
 
 `eth_simulatev1` support will be added to the Mirror Node. It will allow users to send a list of transactions as a bundle, part of a block at specific block height. It will allow state overrides as well.
 
-A new tracer `keccak256PreimageTracer` will be added to the current and newly added debug related endpoints. It will capture the keccak256 operations during transaction execution and this will help users map the hashed storage slots back to their original keys/indices.
+A new tracer `keccak256PreimageTracer` will be added to debug_traceTransaction and debug_traceCall related endpoints. It will capture the keccak256 operations during transaction execution and this will help users map the hashed storage slots back to their original keys/indices.
 
-Support for `prestateTracer` will be added. This will increase the compatibility for debug functionality between EVM compatible ledgers and Hiero. This tracer shows the accounts touched during the EVM execution and their pre  and/or post state.
+Support for `prestateTracer` will be added. This will increase the compatibility for debug functionality between EVM compatible ledgers and Hiero. This tracer shows the accounts touched during the EVM execution and their pre and/or post state.
 
 ## User stories
 
@@ -103,7 +104,11 @@ RLP encoded EVM transactions utilize `gasPrice`, `maxFeePerGas`, `maxPriorityFee
 
 #### State Overrides
 
-The existing `/api/v1/contracts/call` REST API in Mirror Node will be enhanced with an optional request field `state_overrides`. The state overrides allow the user to simulate a call as if the state was actually set using the passed values instead of the real ones from the database. Users can override the `address`, `balance`, `code`, `nonce`, and `state`. Here is an example of a request:
+The existing `/api/v1/contracts/call` REST API in Mirror Node will be enhanced with an optional request field `state_overrides`. The state overrides allow the user to simulate a call as if the state was actually set using the passed values instead of the real ones from the database. Users can override the `balance`, `code`, `nonce`, and `state`. 
+
+The `state` can be overridden using `state` and `state_diff` fields. The `state` maps overridden values for all slots in the contract storage, while the `state_diff` maps individual slots. Both fields are not inclusive, only one of them should be used at a time.
+
+Here is an example of a request:
 
 ```json
 {
@@ -133,17 +138,30 @@ The existing `/api/v1/contracts/call` REST API in Mirror Node will be enhanced w
 }
 ```
 
+#### Tracer Config
+
+`debug_traceCall` and `debug_traceCallMany` will have a new nested object called `tracer_config`. It will contain all needed information for the tracing process, including:
+
+1. `code` - if enabled the `prestateTracer` includes information about the contract bytecode
+2. `diff` - if enabled the `prestateTracer` includes `pre` and `post` arrays comparing changes before and after the transaction execution
+3. `memory` - if enabled the `opcodeLogger` includes information about the EVM memory at each executed opcode
+4. `only_top_call` - if enabled the `callTracer` includes information only about the root top level call
+5. `stack` - if enabled `opcodeLogger` includes information about the EVM stack at each executed opcode
+6. `storage` - if enabled `opcodeLogger` or `prestateTracer` include information about the contract storage
+7. `timeout` - if enabled the execution will timeout with `HTTP 408 REQUEST_TIMEOUT` error, so that no more resources are waisted
+8. `tracer` - points the tracer type that should be used for the tracing. This field determines the output populated as a response.
+
+If the request contains mismatched parameters e.g. enabling `diff` with `callTracer` or enabling `storage` with `prestateTracer`, there will be a `400 BAD_REQUEST` response.
+
+
 #### debug_traceCall
 
-`debug_traceCall` simulates a transaction against the state at a specified block and returns a step-by-step execution 
-trace. It accepts an optional tracer (e.g. callTracer, prestateTracer, or a custom JS/Go tracer) to shape the output, 
-making it the standard tool for debugging contract logic, profiling gas, and inspecting internal calls before sending a 
-real transaction. Depending on the `tracer` that was provided as an input parameter to the `debug_traceCall` 
-RPC endpoint, a new Mirror Node endpoint `POST api/v1/contracts/call/debug` will be added. It will accept a different 
-tracer type in its request body.
+`debug_traceCall` simulates a transaction against the state at a specified block and returns a step-by-step execution trace. It accepts an optional tracer (e.g. callTracer, prestateTracer, or a custom JS/Go tracer) to shape the output, 
+making it the standard tool for debugging contract logic, profiling gas, and inspecting internal calls before sending a real transaction. Depending on the `tracer` that was provided as an input parameter to the `debug_traceCall` 
+RPC endpoint, a new Mirror Node endpoint `POST api/v1/contracts/call/debug` will be added. It will accept a different tracer type in its request body.
 
 1. For `debug_traceCall` called with `callTracer` the tracer tracks all message frames invoked in the transaction execution.
-This tracer will have an optional `onlyTopCall` config that limits the response only to information about the root call.
+This tracer will have an optional `only_top_call` config that limits the response only to information about the root call.
 
 Example request body:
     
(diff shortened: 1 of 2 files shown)


HIP/hip-xxx.md
```diff
@@ -0,0 +1,949 @@
...
+
+`debug_traceCallMany` support will also be added. It will allow tracing a simulation of many transactions simultaneously. It will have the option of overriding block specific context, adding state override, tracer type, and setting a timeout. In this way users can see what traces a bundle of transactions will produce without actually executing them in the network.
+
+The existing `contracts/call` endpoint in Mirror Node will be enhanced via a `state_override` nested object. It will allow users to replace account related fields and/or the bytecode and the state of a contract, without actually modifying the contract itself.
+
+`eth_simulatev1` support will be added to the Mirror Node. It will allow users to send a list of transactions as a bundle, part of a block at specific block height. It will allow state overrides as well.
+
+A new tracer `keccak256PreimageTracer` will be added to the current and newly added debug related endpoints. It will capture the keccak256 operations during transaction execution and this will help users map the hashed storage slots back to their original keys/indices.
```
bilyana-gospodinova: > A new tracer `keccak256PreimageTracer` will be added to the current and newly added debug related endpoints.

Perhaps it should be specified which will be the affected endpoints.

HIP/hip-xxx.md
```diff
@@ -0,0 +1,949 @@
...
+
+Support for `prestateTracer` will be added. This will increase the compatibility for debug functionality between EVM compatible ledgers and Hiero. This tracer shows the accounts touched during the EVM execution and their pre  and/or post state.
```
bilyana-gospodinova: nit: There is a double space at the end of the paragraph - "their pre  and/or post state".

HIP/hip-xxx.md
```diff
@@ -0,0 +1,949 @@
...
+
+#### Storing weibars
+
+RLP encoded EVM transactions utilize `gasPrice`, `maxFeePerGas`, `maxPriorityFeePerGas`, and `value` amounts in weibar, which have more decimal places than tinybar. In order not to lose precision by converting the weibars into tinybars, these fields will be directly saved in weibar denomination without any conversion. In addition, all contract result REST APIs that return these fields will be enhanced with a boolean `hbar` query parameter. If the value is set to true, then the amount will be converted to tinybars as it was before and if the value is set to false, the saved amount will be directly populated. If the value is omitted, then it will default to true to ensure backwards compatibility with existing clients.
+
+#### State Overrides
+
+The existing `/api/v1/contracts/call` REST API in Mirror Node will be enhanced with an optional request field `state_overrides`. The state overrides allow the user to simulate a call as if the state was actually set using the passed values instead of the real ones from the database. Users can override the `address`, `balance`, `code`, `nonce`, and `state`. Here is an example of a request:
```
bilyana-gospodinova: > Users can override the `address`, `balance`, `code`, `nonce`, and `state`.

It sounds like the `address` can be changed. Won't the address be used as a "key" in order to find the account that is actually modified?

HIP/hip-1483.md
```diff
@@ -0,0 +1,949 @@
...
+	"from": "0x00000000000000000000000000000000000004e2",
+	"gas": 100000000,
+	"gas_price": 100000000,
+	"state_overrides": [
+		{
+            "address": "0x00000000000000000000000000000000000004e4",
+		    "balance": "0xFFFFFFFFFFFFFFFFFFFF",
+		    "code": "0x602a60005260206000f3",
+		    "nonce": "0x1",
+		    "state": [
```
bilyana-gospodinova, line 127: Should we mention the difference between "state" and "state_diff"? And that they are mutually exclusive?

HIP/hip-xxx.md
```diff
@@ -0,0 +1,949 @@
...
+		        "code": "",
+		        "nonce": 841,
+		        "storage": {}
+		      }
+		]
+	}
+}
+```
+
+1. For `debug_traceCall` called with `keccak256PreimageTracer`, the tracer will return a list of keccak256 hashes and the original memory slice that was hashed.
```
bilyana-gospodinova: Shouldn't this be point number 4?

HIP/hip-xxx.md
```diff
@@ -0,0 +1,949 @@
...
+    "bundles": [
+        {
+            "transactions": [
+                {
+                    "to": "0x...",
+                    "data": "0x..."
+                }
+            ],
+            "block_override": {
+            "block_number": "0x100"
```
bilyana-gospodinova: Are there other options for "block_override"? If not, can we simplify the nesting? Example: `"block_override": "0x100"`

HIP/hip-xxx.md
```diff
@@ -0,0 +1,949 @@
...
+        "diff": false,
+        "memory": false,
+        "onlyTopCall": true,
+        "stack": false,
+        "storage": false
+    }
+}
+```
+
+Example response  is:
```
bilyana-gospodinova: nit: extra spacing

HIP/hip-xxx.md
```diff
@@ -0,0 +1,949 @@
...
+                    },
+                    "subtraces": 0,
+                    "trace_address": [],
+                    "type": "call"
+                }
+            ]
+        }
+    ],
+    "opcodes": {},
+    "prestate": {}
```
bilyana-gospodinova:
```suggestion
    "opcodes": null,
    "prestate": null
```
In the other examples these are `null`. Should we apply it here for consistency?

HIP/hip-1483.md
```diff
@@ -0,0 +1,949 @@
...
+        },
+        {
+            "transactions": [ { "to": "0x...", "data": "0x..." } ],
+            "block_override": {
+                "block_number": "0x200"
+            }
+        }
+    ],
+    "block_number": "0x40",
+    "transaction_index": -1,
```
bilyana-gospodinova, line 518: Is "-1" some special value in this case? Shouldn't the number be non-negative?

HIP/hip-xxx.md
```diff
@@ -0,0 +1,949 @@
...
+        {
+            "transactions": [ { "to": "0x...", "data": "0x..." } ],
+            "block_override": {
+                "block_number": "0x200"
+            }
+        }
+    ],
+    "block_number": "0x40",
+    "transaction_index": -1,
+	"timeout": "30s",
```
bilyana-gospodinova: What response will the API return, if the timeout is exceeded?

HIP/hip-1483.md
```diff
@@ -0,0 +1,949 @@
...
+                    "data": "0x..."
+                }
+            ],
+            "block_override": {
+                "block_number": "0x100"
+            }
+        }
+    ],
+    "block_number": "0x40",
+    "transaction_index": -1,
```
bilyana-gospodinova, line 645: In some places the value for this field is "-1", in others it is in hex format - ["0x0"](https://github.com/hiero-ledger/hiero-improvement-proposals/pull/1483/changes#diff-6f4c0d45beba6dc9e0306492176168d468b70ba54869e46eaefc2b05cd461842R723). I think we should keep it consistent.

HIP/hip-xxx.md
```diff
@@ -0,0 +1,949 @@
...
+
+This tracer will be supported by `debug_traceTransaction` and `debug_traceCall`. Similarly to `prestateTracer`, a new endpoint will be defined for handling `keccak256PreimageTracer`
+for `debug_traceTransaction`. It will be `GET /api/v1/contracts/results/{transactionIdOrHash}/keccak256` .
+
+An example response from a `debug_traceTransaction` using a `keccak256PreimageTracer` is:
+
+```json
+{
+	"actions": null,
+	"opcodes": null,
```
bilyana-gospodinova: Since it is a completely new endpoint, do we need to return such empty fields that will never be set in this endpoint?

HIP/hip-xxx.md
IvanKavaldzhiev: Yes, updated.

HIP/hip-xxx.md
IvanKavaldzhiev: Fixed.

HIP/hip-xxx.md
IvanKavaldzhiev: Yes, you're correct. I removed it as a filed that can be changed.

HIP/hip-1483.md
IvanKavaldzhiev: Added.

HIP/hip-xxx.md
IvanKavaldzhiev: Correct, fixed.

HIP/hip-xxx.md
IvanKavaldzhiev: It seems we can also support time, so I left `number` and `time`.

HIP/hip-xxx.md
IvanKavaldzhiev: Fixed.

HIP/hip-xxx.md
IvanKavaldzhiev: Fixed.

HIP/hip-1483.md
IvanKavaldzhiev: Yes, -1 means we should execute the call/batch after all transactions in the block.

HIP/hip-xxx.md
IvanKavaldzhiev: I think we should return HTTP 408 timeout error, I added this detail.

HIP/hip-1483.md
IvanKavaldzhiev: When using decimals it is in a request body, while hex is used in the response.

HIP/hip-xxx.md
IvanKavaldzhiev: I wanted to reuse the scheme defined for the other endpoints, but since this endpoint will be specific only for the keccak256preimage tracer we can define a specific lighter scheme.

HIP/hip-xxx.md
bilyana-gospodinova: Perhaps for `time` there should be some example, so the users will know the format.

HIP/hip-1483.md
```diff
@@ -0,0 +1,971 @@
...
+	"gas": 45000,
+	"gas_price": 68000,
+	"to": "0x00000000000000000000000000000000000005c1",
+	"value": 0,
+	"tracer": "keccak256PreimageTracer",
+	"tracer_config": null
+}
+```
+
+An example response::
```
bilyana-gospodinova, line 371: nit: Double ":"

HIP/hip-1483.md
```diff
@@ -0,0 +1,971 @@
...
+        "only_top_call": true,
+        "stack": false,
+        "storage": false,
+        "timeout": "30s",
+        "tracer": "callTracer"
+    }
+}
+```
+    
+    Example response:
```
bilyana-gospodinova, line 190: nit: Extra tab at the beginning of the line. It is missing on other places.

HIP/hip-1483.md
bilyana-gospodinova: Can you please add this detail in the PR?

HIP/hip-1483.md
```diff
@@ -0,0 +1,971 @@
...
+```
+
+#### **Keccak256 Preimage Tracer**
+
+The Ethereum ecosystem was enhanced with an additional tracer called `keccak256PreimageTracer`.
+
+During EVM execution the KECCAK256 opcode creates a 32-byte Keccak-256 hash on a slice of the memory. This is a one-way operation and the users can’t discover which memory value produced the specific hash. The keccak256 preimage tracer provides a mapping between a keccak hashed value and the preimage (the memory slice that was hashed).
+
+This tracer will be supported by `debug_traceTransaction` and `debug_traceCall`. Similarly to `prestateTracer`, a new endpoint will be defined for handling `keccak256PreimageTracer`
+for `debug_traceTransaction`. It will be `GET /api/v1/contracts/results/{transactionIdOrHash}/keccak256` .
```
bilyana-gospodinova, line 910:
```suggestion
for `debug_traceTransaction`. It will be `GET /api/v1/contracts/results/{transactionIdOrHash}/keccak256`.
```
nit

PR #1485, opened 2026-06-02
HIP-1485: Add mirror node debug and trace APIs

IvanKavaldzhiev:
**Description**:
Add a HIP definition for some new debug and trace APIs that are going to be added on mirror-node side. In addition some new tracers - prestateTracer and keccak256PreimageTracer will be added to the existing debug endpoint and the newly added debug_traceCall.

Some performance and compatibility optimisations will be added as well:

- the receipts_root field will be now calculated directly on ingestion, instead of being manually calculated on JSON-RPC relay side
- all weibar values used in Ethereum transactions will be saved directly to DB, without conversion to tinybars, so that they don't loose precision.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


HIP/hip-1485.md
```diff
@@ -0,0 +1,972 @@
...
+
+The performance of block related APIs on the JSON-RPC relay can be improved by calculating the `receipts_root` field on ingestion instead of calculating it on demand.
+
+## Rationale
+
+Currently the JSON-RPC Relay is performing intensive calculations to populate the `receipts_root` for block related APIs. It fetches data from multiple mirror-node APIs and iterates over multiple items inside a block to populate the field. This whole calculation will be moved to the Mirror Node importer and calculated during block stream ingestion.
+
+When a transaction is sent to the JSON-RPC Relay via `eth_sendRawTransaction` a special denomination called weibar is used. Currently, when saving EVM transactions, Mirror Node converts weibar to tinybar by dividing by 10 million and thus losing precision for all fields that store their amounts in weibars. As a consequence, it’s impossible for clients to reconstruct the RLP form of the EVM transaction using the data in the REST APIs. This HIP will propose a change to directly store the weibar amounts to unblock such cases. It will expose a new query parameter called `hbar` to enable this new format.
```
Reccetech: It says divide by 10 million. I think its 10 billion. 
public static final Long WEIBARS_TO_TINYBARS = 10_000_000_000L;  // 10 BILLION

HIP/hip-1485.md
```diff
@@ -0,0 +1,972 @@
...
+	"timestamp": {
+	    "from": "1651560386.060890949",
+	    "to": "1651560386.661997287"
+	}
+}
+```
+
+The endpoint `/api/v1/blocks` will be similarly affected.
+
+#### Storing weibars
```
Reccetech, line 100: I don't fully understand this. What about historical transactions?

HIP/hip-1485.md
```diff
@@ -0,0 +1,972 @@
...
+4. `only_top_call` - if enabled the `callTracer` includes information only about the root top level call
+5. `stack` - if enabled `opcodeLogger` includes information about the EVM stack at each executed opcode
+6. `storage` - if enabled `opcodeLogger` or `prestateTracer` include information about the contract storage
+7. `timeout` - if enabled the execution will timeout with `HTTP 408 REQUEST_TIMEOUT` error, so that no more resources are waisted
+8. `tracer` - points the tracer type that should be used for the tracing. This field determines the output populated as a response.
+
+If the request contains mismatched parameters e.g. enabling `diff` with `callTracer` or enabling `storage` with `prestateTracer`, there will be a `400 BAD_REQUEST` response.
+
+
+#### debug_traceCall
```
Reccetech, line 156: In some examples the tracer field is inside tracer_config, and in others it is outside at the top request level. Is this desired?

HIP/hip-1485.md
```diff
@@ -0,0 +1,972 @@
...
+        "only_top_call": true,
+        "stack": false,
+        "storage": false,
+        "timeout": "30s",
+        "tracer": "callTracer"
+    }
+}
+```
+
+Example response is:
```
Reccetech, line 455: does not match the Geth callTracer format used in the debug_traceCall callTracer response.

{

HIP/hip-1485.md
```diff
@@ -0,0 +1,972 @@
...
+    ],
+    "block_number": "0x40",
+    "transaction_index": -1,
+    "timeout": "30s",
+    "tracer": "prestateTracer",
+    "tracer_config": null
+}
+```
+
+Example response is:
```
Reccetech, line 696: In debug_traceCall, the prestate result is wrapped in the top-level response envelope. 
In this example the debug_traceCallMany, the prestate result is returned bare at the top level:

HIP/hip-1485.md
```diff
@@ -0,0 +1,972 @@
...
+        {
+            "calls": [
+                {
+                    "from": "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045", 
+                    "max_fee_per_gas": "0xf", 
+                    "to": "0x014d023e954bAae7F21E56ed8a5d81b12902684D", 
+                    "value": "0x1"
+                }
+            ], 
+            "state_overrides": [
```
Reccetech, line 760: This section defines state_overrides as an array of objects with an explicit address field.
The Ethereum standard (stateOverride parameter in eth_call, debug_traceCall, eth_simulateV1) uses a map keyed by address:

"stateOverride": {
  "0x...": { "balance": "...", ... }
}

Is this something we should consider?

HIP/hip-1485.md
```diff
@@ -0,0 +1,972 @@
...
+	    "to": "1651560386.661997287"
+	}
+}
+```
+
+The endpoint `/api/v1/blocks` will be similarly affected.
+
+#### Storing weibars
+
+RLP encoded EVM transactions utilize `gasPrice`, `maxFeePerGas`, `maxPriorityFeePerGas`, and `value` amounts in weibar, which have more decimal places than tinybar. In order not to lose precision by converting the weibars into tinybars, these fields will be directly saved in weibar denomination without any conversion. In addition, all contract result REST APIs that return these fields will be enhanced with a boolean `hbar` query parameter. If the value is set to true, then the amount will be converted to tinybars as it was before and if the value is set to false, the saved amount will be directly populated. If the value is omitted, then it will default to true to ensure backwards compatibility with existing clients.
```
Reccetech, line 102: so:
hbar=true (default): return tinybars (the old behavior)
hbar=false: return weibars (the new behavior)

You would think if your setting HBAR to true it would return in HBAR not tinybars.

HIP/hip-1485.md
```diff
@@ -0,0 +1,972 @@
...
+{
+	"count": 3,
+	"gas_used": 300000,
+	"hapi_version": "0.11.0",
+	"hash": "0x3c08bbbee74d287b1dcd3f0ca6d1d2cb92c90883c4acf9747de9f3f3162ad25b999fc7e86699f60f2a3fb3ed9a646c6b",
+	"logs_bloom": "0x00000020002000001000000000000000000000000000000000000000000010000000000004000000000000000000000000108000000000000000000080000000000004000000000000000000000000880000000000000000000101000000000000000000000000000000000000008000000000000400000080000000000001000000000000000000000000000000000000000000002000000000100000100000200000040000100000001000000000000000000000000000000001001000004000000000000000000001000000000000000000100000000000100000000000000000000000000000000000000000000000080000100800000000000000120080",
+	"name": "2022-05-03T06_46_26.060890949Z.rcd",
+	"number": 77,
+	"previous_hash": "0xf7d6481f659c866c35391ee230c374f163642ebf13a5e604e04a95a9ca48a298dc2dfa10f51bcbaab8ae23bc6d662a0b",
+	"receipts_root": "0x56e91f271bcc54a6ff8345e692c0f86e5b48e21b936caed001722fb5e363b425"
```
Reccetech: missing comma

HIP/hip-1485.md
```diff
@@ -0,0 +1,972 @@
...
+
+Since the `debug_traceTransaction` utilizes the `GET /api/v1/contracts/results/{transactionIdOrHash}/opcodes` endpoint, which is scoped to tracking opcodes only, there will be a new endpoint for supporting the `prestateTracer`. It will be `GET /api/v1/contracts/results/{transactionIdOrHash}/prestate` , which will accept several query parameters - `diff`, `code`, `storage`. If the `diff` is set to `true`, the endpoint will return two different arrays - `pre` and `post` for account traces before and after the transaction execution. If the `diff` is set to `false`, only the `pre` array will be populated with account traces using the state before the transaction execution. If `code` is set to `true`, the `code` field will be populated if we have a contract. The same applies for `storage`.
+
+An example request and response in `prestate` mode:
+
+`GET /api/v1/contracts/results/0xe044554a0a55067caafd07f8020ab9f2af60bdfe337e395ecd84b4877a3d1ab4/prestate?code=true&&storage=true`
```
Reccetech: double ampersand.

HIP/hip-1485.md
IvanKavaldzhiev: Previously when an EthereumTransaction is ingested inside the importer in mirror-node, the `gasPrice`, `maxFeePerGas`, `maxPriorityFeePerGas`, and `value` amounts which are originally in weibars were converted to tinybars by diving with WEIBARS_TO_TINYBARS. This resulted in lost precision of these fields in the database. With the new changes, the original weibar value is stored.

For historical transactions there was a DB migration that restored the fields using the original calldata, which is rlp-encoded.

HIP/hip-1485.md
IvanKavaldzhiev: You're correct, fixed.

HIP/hip-1485.md
IvanKavaldzhiev: Correct, I changed the tracer_config to be the same in all places.

HIP/hip-1485.md
IvanKavaldzhiev: It seems it was a wrong example, indeed. I added a new structure for debug_traceCall and debug_traceCallMany that corresponds to the static scheme design for REST APIs and allows recursive multiple nested call actions to be returned.

HIP/hip-1485.md
IvanKavaldzhiev: Fixed.

HIP/hip-1485.md
IvanKavaldzhiev: On mirror-node we don't follow strictly the official JSON-RPC format. The change from map of overrides to list of overrides is purely cosmetic but it would be good to see the point of view of the relay team, so that they confirm whether this format using lists will suit their needs.

HIP/hip-1485.md
IvanKavaldzhiev: I think it depends. If the value is very big, the tinybars might amount >= 1 HBAR, so using hbar as a flag can refer to both tinybars and hbars.

HIP/hip-1485.md
IvanKavaldzhiev: Fixed

HIP/hip-1485.md
IvanKavaldzhiev: Fixed

HIP/hip-1485.md
```diff
@@ -0,0 +1,1126 @@
...
+---
+
+
+## Abstract
+
+To enhance the EVM compatibility of the network, additional debug and trace APIs will be added to the Mirror Node and JSON-RPC Relay. Additionally, existing APIs will be enhanced to support additional tracers and options to aid in debugging EVM transactions.
+
+## Motivation
+
+When users encounter an issue with their contract, it is challenging to debug since it is executed on a remote machine outside their control. For this reason, various debug and tracing APIs were added to the JSON-RPC specification to improve the user’s troubleshooting experience. To improve the EVM compatibility related to block support, single call, and multiple calls debugging functionality, we need to perform some changes to the API endpoint definitions on the Mirror Node.
```
Ferparishuertas, line 26: Shouldnt we add support accesslist in the mentioned mirror node, rpc endpoints eth_call simulatev1, traceCall, many?

IvanKavaldzhiev: Added.

HIP/hip-1485.md
```diff
@@ -0,0 +1,1396 @@
...
+	    "to": "1651560386.661997287"
+	}
+}
+```
+
+The endpoint `/api/v1/blocks` will be similarly affected.
+
+#### Storing weibars
+
+RLP encoded EVM transactions utilize `gasPrice`, `maxFeePerGas`, `maxPriorityFeePerGas`, and `value` amounts in weibar, which have more decimal places than tinybar. In order not to lose precision by converting the weibars into tinybars, these fields will be directly saved in weibar denomination without any conversion. In addition, all contract result REST APIs that return these fields will be enhanced with a boolean `hbar` query parameter. If the value is set to true, then the amount will be converted to tinybars as it was before and if the value is set to false, the saved amount will be directly populated. If the value is omitted, then it will default to true to ensure backwards compatibility with existing clients.
```
steven-sheehy: We should list the APIs impacted so it's clear.

HIP/hip-1485.md
```diff
@@ -0,0 +1,1396 @@
...
+
+`debug_traceCall` and `debug_traceCallMany` will have a new nested object called `tracer_config`. It will contain all needed information for the tracing process, including:
+
+1. `code` - if enabled the `prestateTracer` includes information about the contract bytecode
+2. `diff` - if enabled the `prestateTracer` includes `pre` and `post` arrays comparing changes before and after the transaction execution
+3. `memory` - if enabled the `opcodeLogger` includes information about the EVM memory at each executed opcode
+4. `only_top_call` - if enabled the `callTracer` includes information only about the root top level call
+5. `stack` - if enabled `opcodeLogger` includes information about the EVM stack at each executed opcode
+6. `storage` - if enabled `opcodeLogger` or `prestateTracer` include information about the contract storage
+7. `timeout` - if enabled the execution will timeout with `HTTP 408 REQUEST_TIMEOUT` error, so that no more resources are waisted
```
steven-sheehy:
```suggestion
7. `timeout` - if enabled the execution will timeout with `HTTP 408 REQUEST_TIMEOUT` error, so that no more resources are wasted
```

HIP/hip-1485.md
```diff
@@ -0,0 +1,1396 @@
...
+
+Example response:
+
+```json
+{
+    "actions": 
+        [
+            [
+                {
+                    "calls": [
```
steven-sheehy: Why is there an array within an array?

HIP/hip-1485.md
```diff
@@ -0,0 +1,1396 @@
...
+            "s": "0x3333333333333333333333333333333333333333333333333333333333333333"
+        }
+    ]
+```
+
+#### debug_traceCall
+
+`debug_traceCall` simulates a transaction against the state at a specified block and returns a step-by-step execution trace. It accepts an optional tracer (e.g. callTracer, prestateTracer, or a custom JS/Go tracer) to shape the output,
+making it the standard tool for debugging contract logic, profiling gas, and inspecting internal calls before sending a real transaction. The call object accepts optional `access_list` and `authorization_list` fields as described in [Access List and Authorization List](#access-list-and-authorization-list). Depending on the `tracer` that was provided as an input parameter to the `debug_traceCall`
+RPC endpoint, a new Mirror Node endpoint `POST api/v1/contracts/call/debug` will be added. It will accept a different tracer type in its request body.
```
steven-sheehy: We should not have an API where there's an enum in the request that causes it to populate mutually exclusive fields in the response. Nor should we have inputs in the request that don't apply to certain tracers.

Instead, we should split this endpoint into:

* `POST /api/v1/contracts/call/actions`
* `POST /api/v1/contracts/call/keccak256`
* `POST /api/v1/contracts/call/opcodes`
* `POST /api/v1/contracts/call/prestate`

Drop generic tracer_config and set relevant fields in top level request object. Each endpoint should have its own sub-section in the doc.

HIP/hip-1485.md
```diff
@@ -0,0 +1,1396 @@
...
+```json
+{
+    "block": "latest",
+    "data": "0x81a73ad500000000000000000000000000000000000000000000000000000000000006b2",
+    "from": "0x00000000000000000000000000000000000003d8",
+    "gas": 45000,
+    "gas_price": 68000,
+    "to": "0x00000000000000000000000000000000000005c1",
+    "value": 0,
+    "access_list": [
```
steven-sheehy: sort all inputs and outputs. Sweep through the doc and do it consistently throughout

HIP/hip-1485.md
```diff
@@ -0,0 +1,1396 @@
...
+    }
+}
+```
+
+Example response:
+
+```json
+{
+	"actions": [],
+	"opcodes": {
```
steven-sheehy: With separate API can drop opcodes field wrapper.

HIP/hip-1485.md
```diff
@@ -0,0 +1,1396 @@
...
+    "opcodes": null, 
+    "preimage": null, 
+    "prestate": {
+        "pre": [
+            {
+                "address": "0xd26d9000ec8778ae3257305c9acdd79f16f87857", 
+                "balance": "0x9d4", 
+                "code": "", 
+                "nonce": 840, 
+                "storage": null
```
steven-sheehy: Show an example with storage

HIP/hip-1485.md
```diff
@@ -0,0 +1,1396 @@
...
+
+```json
+{
+    "pre": [
+          {
+              "address": "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045",
+              "balance": "0x0",
+              "code": "",
+              "nonce": 3,
+              "storage": null
```
steven-sheehy: Use empty object instead of null. Sweep through rest of doc to fix other areas.

steven-sheehy (changes requested):

HIP/hip-1485.md
```diff
@@ -0,0 +1,1396 @@
...
+        "post": []
+    }
+}
+```
+
+#### eth_simulateV1
+
+A new endpoint will be added in JSON-RPC Relay. It will allow testing multiple transaction in a specific block context and expecting their output.
+
+The list of transactions will be executed using the block context that is passed. If the `validation` property is set to `true` the transactions will be validated as if they are executed inside a real EVM. The only exception is validating whether the transaction sender is an EOA or a smart contract and the sender signature. If the `traceTransfers` property is set to `true`, all transferred values in tinybars will be captured by emitting ERC20 transfer event logs. The emitter contract will be `0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee` and this allow tracking internal transfers.
```
steven-sheehy: Do we have the necessary control within the CN library to disable validation?

HIP/hip-1485.md
```diff
@@ -0,0 +1,1396 @@
...
+                    "value": "0x1",
+                    "access_list": [
+                        {
+                            "address": "0xde0b295669a9fd93d5f28d9ec85e40f4cb697bae",
+                            "storage_keys": [
+                                "0x0000000000000000000000000000000000000000000000000000000000000003"
+                            ]
+                        }
+                    ],
+                    "authorization_list": [
```
steven-sheehy, line 908: I've not actually seen any examples for eth_simulateV1 that take an access list or authorization list. Are we sure we should support it? Do you have a source link?

HIP/hip-1485.md
```diff
@@ -0,0 +1,1396 @@
...
+               {
+                 "key": "0x0000000000000000000000000000000000000000000000000000000000000002",
+                 "value": "0x000000000000000000000000000000000000000000000000000000000000270f"
+               }
+             ]
+           }
+         ]
+       }
+    ], 
+    "trace_transfers": true
```
steven-sheehy, line 1032: Where is the validation flag?

HIP/hip-1485.md
```diff
@@ -0,0 +1,1396 @@
...
+}
+```
+
+An example response is:
+
+```json
+{
+    "result": [
+        [
```
steven-sheehy: Why is there an array within an array?

HIP/hip-1485.md
```diff
@@ -0,0 +1,1396 @@
...
+```json
+{
+    "actions": [],
+    "opcodes": null,
+    "preimage": null,
+    "prestate": {
+        "pre": [
+            {
+                "address": "0xd26d9000ec8778ae3257305c9acdd79f16f87857",
+                "balance": "0x9d4",
```
steven-sheehy: We should be consistent and encode weibars as numerics like in existing APIs.

HIP/hip-1485.md
```diff
@@ -0,0 +1,1396 @@
...
+                        "data": "0x0000000000000000000000000000000000000000000000000000000000000001",
+                        "log_index": "0x0",
+                        "removed": false,
+                        "topics": [
+                            "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef",
+                            "0x000000000000000000000000c000000000000000000000000000000000000000",
+                            "0x000000000000000000000000c000000000000000000000000000000000000001"
+                        ],
+                        "transaction_hash": "0xe7217784e0c3f7b35d39303b1165046e9b7e8af9b9cf80d5d5f96c3163de8f51",
+                        "transaction_index": "0x0"
```
steven-sheehy: We should be consistent and encode log_index, transaction_index, and block_number fields as numeric like in existing APIs.

steven-sheehy (changes requested):

HIP/hip-1485.md
IvanKavaldzhiev: Geth mentions that for calls passed in eth_simulateV1 they support list of GenericCallTransaction.
And a GenericCallTransaction supports accessList.

https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-eth
https://geth.ethereum.org/docs/interacting-with-geth/rpc/objects#transaction-call-object

Tenderly documents supporting accessList and authorizationList as optional parameters.

https://docs.tenderly.co/node-rpc/guides/eth-simulatev1#simulate-blocks-with-eth_simulatev1

It may be other providers also support them, but just don't show examples with these fields.

HIP/hip-1485.md
IvanKavaldzhiev: We can't support it and I might have deleted it in the request body, but forgotten to remove the text description in the beginning about this flag.

HIP/hip-1485.md
IvanKavaldzhiev: Added

HIP/hip-1485.md
IvanKavaldzhiev: Fixed

HIP/hip-1485.md
IvanKavaldzhiev: Refactored

HIP/hip-1485.md
IvanKavaldzhiev: Added

HIP/hip-1485.md
IvanKavaldzhiev: Fixed

HIP/hip-1485.md
IvanKavaldzhiev: Fixed

HIP/hip-1485.md
IvanKavaldzhiev: Added

HIP/hip-1485.md
IvanKavaldzhiev: Fixed

HIP/hip-1485.md
IvanKavaldzhiev: We don't. Forgot to delete the description related to the validation property.

HIP/hip-1485.md
IvanKavaldzhiev: Fixed

HIP/hip-1485.md
IvanKavaldzhiev: Fixed

HIP/hip-1485.md
IvanKavaldzhiev: Fixed

PR #1487, opened 2026-06-08
chore: Add clarification of Registered Node behavior

jsync-swirlds:
### Description
* Clarified that all endpoints for a single Registered Node should behave as though all are provided by a single logical entity.


xin-hedera (approved):
LGTM. Thanks for the clarification.

mgarbs (approved):

PR #1494, opened 2026-06-18
HIP 1494: Consensus Node Deployment Package Specification

leninmehedy:
**Description**:

This proposal defines the structure and schema of the Consensus Node Deployment Package (`build.zip`) — the
single artifact used to deploy and upgrade Hedera consensus nodes. The same package format serves all
operational contexts: network upgrades driven by the council freeze protocol, fresh node provisioning, and
network bootstrapping. The package bundles consensus node software, configuration files, and a `manifests/`
directory containing machine-readable specifications for container images, infrastructure software versions,
state sources, and large remote files.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


nathanklick:
## Proposal: an `archives:` section in `external-files.yaml` for verifiable archives with verifiable contents

### Motivation

`external-files.yaml` today can only describe a *single opaque file per entry*: `destination` names one path, and there is no unpack, extract, or member-selection directive anywhere in the entry schema. That does not cover a real and already-live case.

The concrete example is the WRAPS recursive-proof proving keys. `tss.wrapsEnabled=true` requires four universal proving-key artifacts to be staged under `TSS_LIB_WRAPS_ARTIFACTS_PATH` before the node starts, or the node logs that it cannot build recursive proofs and WRAPS never initialises. Those artifacts are published as **one ~2 GB `wraps-v1.0.0.tar.gz` containing four `.bin` members** (`decider_pp.bin`, `decider_vp.bin`, `nova_pp.bin`, `nova_vp.bin`).

Expressing that requirement in the current schema forces one of three bad options:

1. Publish the four members as four independently addressable URLs with four checksums — a demand on the artifact publisher, not on the package author, and it discards the single-artifact provenance the tarball already has.
2. Declare the tarball as a single `files[]` entry landing at some path, and rely on out-of-band tooling to unpack it — which puts the most security-sensitive step (writing N attacker-influenced paths to disk) entirely outside the manifest contract, and outside the `destination` marker-prefix policy that section exists to enforce.
3. Re-host the members ourselves, which forks provenance from upstream.

Archives are also the natural shape for the general case this section is for: large, versioned, multi-file payloads (proving keys, embedded datasets, model or circuit artifacts) are published as one archive far more often than as N loose files.

### Proposed shape

A new **top-level `archives:` key**, sibling to `files:`, rather than an `unpack:` block bolted onto `files[]`. Rationale for the separate key:

- `files[].destination` is a **file** path; an archive's destination is a **directory**. Overloading one field with two meanings — disambiguated by the presence of a sibling block — makes every consumer branch on entry shape before it can validate anything.
- The verification contract is genuinely different: one checksum, versus N+1 checksums *plus* a set-equality invariant over the archive membership.
- The failure modes are different and deserve their own error taxonomy (see below).
- It leaves `files[]` and its existing validation completely untouched.

```yaml
schemaVersion: 1

files:
  # ... unchanged ...

archives:
  - url: "https://builds.hedera.com/tss/hiero/wraps/v1.0/wraps-v1.0.0.tar.gz"
    format: "tar.gz"                              # tar.gz | tar | zip — explicit, never inferred
    algorithm: "sha256"                           # of the archive exactly as downloaded
    checksum: "abc123..."
    contentType: "application/gzip"               # optional
    destination: "HAPIAPP_DIR/data/keys/wraps"    # a DIRECTORY marker path
    optional: false                               # default; failure aborts, as with files[]
    phase:
      download: prepare
      install: freeze
    contents:                                     # AUTHORITATIVE and exhaustive
      - path: "decider_pp.bin"                    # member path within the archive
        algorithm: "sha256"
        checksum: "def456..."
        size: 2170538264                          # optional; see the disk pre-check below
      - path: "decider_vp.bin"
        algorithm: "sha256"
        checksum: "789ghi..."
      - path: "nova_pp.bin"
        algorithm: "sha256"
        checksum: "012jkl..."
      - path: "nova_vp.bin"
        algorithm: "sha256"
        checksum: "345mno..."
```

An optional per-member `destination` override (subject to the same marker-prefix policy) covers flattening or relocating members out of a nested archive; when absent the install path is `destination` + `/` + `path`.

### The authoritative-contents invariant

This is the part I would most like to see land as a hard rule rather than a recommendation: **`contents` is the complete and authoritative description of the archive. No extraneous file may be present in the archive, and nothing not listed in `contents` may ever reach the destination.**

That means set *equality*, not set *containment*, checked in both directions:

- A regular-file member present in the archive but absent from `contents` → **abort**. Not "extract and ignore" — the archive itself is not what the manifest says it is, and treating that as benign is how an extra payload rides along inside a signed package.
- A `contents[].path` absent from the archive → **abort**.

Directory entries are ignored for the equality check (they are structural). Any non-regular member — symlink, hardlink, device node, FIFO, socket — is **rejected outright and never extracted**, which also closes the classic tar-slip / zip-slip class of escape.

### Validation rules (parse time)

1. `url`, `format`, `algorithm`, `checksum`, `destination`, `phase.download`, `phase.install` are required and non-empty.
2. `format` must be one of a closed set (`tar.gz`, `tar`, `zip`), and is **never inferred from the URL** — a URL extension is attacker-influenced metadata, and content-sniffing an archive type is exactly the ambiguity worth designing out.
3. `destination` must begin with a recognised marker prefix followed by `/`, reusing the existing `HAPIAPP_DIR` / `SOLO_PROVISIONER_DIR` closed set and its enforcement.
4. `contents` is required and must declare at least one entry.
5. `contents[].path` must be relative: no leading `/`, no `.` or `..` path segment, no drive letter, no backslash separator. Paths must be unique within the archive.
6. `contents[].algorithm` and `contents[].checksum` are required and non-empty.
7. Every **resolved install path** — across all `archives[]` members *and* all `files[]` entries — must be unique. Two entries silently overwriting each other at install time is the same defect the existing `files[].destination` uniqueness rule already guards against; the archive case just widens the namespace it has to be checked in.

### Processing rules (apply time)

1. Verify the archive's own checksum **before extracting anything**.
2. Extract to a staging directory on the same filesystem as the final destination — never directly into the destination.
3. Enumerate members and enforce the set-equality invariant above before any checksum work.
4. Verify every member's checksum in staging.
5. Install **atomically as a unit**: move members into place only after every one of them verifies. A partially installed archive leaves the host in exactly the "partial state that is worse than a clean failure" the existing required-entry rationale calls out.
6. `optional: true` applies to the archive as a whole. There is no partial-member success.
7. The disk-space pre-check covers the archive size **plus** the uncompressed total. An optional `size` field on the archive and on each `contents[]` entry makes that computable ahead of time — without a trial extraction, and without trusting the archive's own header.

### Suggested JSONL / error classifications

Following the existing `FileDownloadFailed` / `FileHashMismatch` / `ManifestInvalid` naming:

| Entry | Condition |
|---|---|
| `ArchiveDownloadFailed` | Transport, HTTP, or disk-space failure fetching the archive |
| `ArchiveHashMismatch` | Archive checksum does not match `checksum` |
| `ArchiveContentsMismatch` | Archive membership is not exactly `contents` (either direction) |
| `ArchiveMemberHashMismatch` | An extracted member's checksum does not match its `contents[]` entry |
| `ArchiveUnsafeMember` | Non-regular member, or a member path escaping the destination |

### `schemaVersion` implications

By the HIP's own evolution rules this is an **additive** change — a new optional top-level key — and so should stay at `schemaVersion: 1`, with older tooling ignoring `archives:` under the required lenient parsing.

That holds only if lenient parsing is actually implemented. The reference parser in `hashgraph/solo-weaver` currently strict-decodes every manifest (`decodeStrictSingleYAMLDoc` sets `dec.KnownFields(true)`), so an unknown top-level key is a hard parse error today. As things stand, adding `archives:` would be breaking in practice for any component that has not upgraded — which is precisely the scenario §Backwards Compatibility's "MUST use lenient parsing" exists to prevent.

I'd rather see the parser brought in line with the stated MUST than see this section bumped to `schemaVersion: 2` to work around it, since the same problem will recur on the next additive field. Flagging it here because the two decisions are coupled.

### Related, while in this section

A few smaller things noticed in the same area of the draft:

- **`phase.install` value mismatch.** The example and the bullet list both document `frozen`, but the reference parser accepts only `freeze` (`InstallPhaseFreeze InstallPhase = "freeze"`) and rejects anything else with a validation error. A package authored against the published example is refused by the reference implementation. Whichever spelling wins, spec and implementation should agree.
- **Garbled paragraph** in the "`external-files.yaml` for Large Files" rationale — the text runs "…from the install timing (`frozen` — after `now_frozen.mf`, CN stopped). See window) from the install timing (`freeze` …", which looks like a merge artifact and is where the mismatch above is visible.
- **`url` scheme is unvalidated** on `files[]`, while `state-sources.yaml` enforces a closed `gcs://` / `s3://` scheme set on `bucket`. Two fields doing the same job — telling the downloader which client to use — get opposite treatment. Worth stating deliberately which one is intended; the WRAPS example above relies on `https://` being acceptable, which the current examples never show.

leninmehedy:
> ## Proposal: an `archives:` section in `external-files.yaml` for verifiable archives with verifiable contents


Thank you @nathanklick. I have incorporated your suggestions: [47f9b44](https://github.com/hiero-ledger/hiero-improvement-proposals/pull/1494/commits/47f9b441d447e26e49f9bd1cbfbfe6c3298fac87)

PR #1500, opened 2026-06-22
Add HIP: Periodic Holding Fee for Hedera Token Service Fungible Tokens

AdrianKBL:
## Summary

This HIP introduces `PeriodicHoldingFee` (PHF), a new top-level configuration on fungible
HTS tokens that lets an issuer periodically debit either a percentage of each non-exempt
holder's balance (fractional mode) or a flat amount per holder (fixed mode), denominated in
the token, crediting a configured collector. PHF reaches passive holders of record — unlike
custom fees, which only fire on transfers — for tokens with ongoing off-ledger costs
(custody, insurance, audit, compliance).

- **Type:** Standards Track   **Category:** Service   **Status:** Draft
- **Requested by:** Kabila / Bravalta
- **Requires:** HIP-18, HIP-336, HIP-540, HIP-904, HIP-1010, HIP-1261

## What it changes

- New `assessTokenPeriodicHoldingFee` HAPI op (issuer-driven, keeper-orchestrated, signed by a
  dedicated `holdingFeeKey`) plus atomic "touch settlement" on balance-changing operations.
- Integer-only per-cycle math, immutable cap + mutable current rate, one `last_assessed_at`
  timestamp per relationship — no holder registry, no autonomous consensus work, no HBAR reserve.
- PHF settlements are exempt from the token's custom-fee schedule, bounding holder exposure to
  `cap × cycles`, never more than balance, never a debt.
- Additive protobuf changes (new messages/fields, one RPC, new response codes, one
  `HederaFunctionality`); Mirror Node and SDK impacts specified.

## Notes for editors

- HIP number left as `0000` placeholder per HIP-1 (please assign).
- `Impact on Mirror Node` / `Impact on SDK` are kept as subsections of `Specification`, matching
  recent Service HIPs (e.g. HIP-1340, HIP-991).
- Open Issues are scoped deliberately (fee line items, throttle policy, EVM log surface for
  precompile-triggered settlement, HIP-551 composition, token-expiry behavior, final proto tags).

### DIFF
--- a/HIP/hip-1500.md
+++ b/HIP/hip-1500.md
@@ -1,5 +1,5 @@
 ---
-hip: 0000
+hip: 1500
 title: Periodic Holding Fee for Hedera Token Service Fungible Tokens
 author: Brandon Hargreaves (@BHargreaves-Twig), Jesse Whiteside (@DLTGroup)
 working-group: Adrian Marin (@AdrianKBL), Rafa (@raludi)
@@ -13,8 +13,6 @@ status: Draft
 created: 2026-04-29
 updated: 2026-06-22
 requires: 18, 336, 540, 904, 1010, 1261
-replaces:
-superseded-by:
 release:
 ---
 
--- a/HIP/hip-0000-periodic-holding-fee.md
+++ b/HIP/hip-0000-periodic-holding-fee.md
@@ -1,0 +1,654 @@
+---
+hip: 0000
+title: Periodic Holding Fee for Hedera Token Service Fungible Tokens
+author: Brandon Hargreaves (@BHargreaves-Twig), Jesse Whiteside (@DLTGroup)
+working-group: Adrian Marin (@AdrianKBL), Rafa (@raludi)
+requested-by: Bravalta / Kabila
+discussions-to: https://github.com/hiero-ledger/hiero-improvement-proposals/pull/1500
+type: Standards Track
+category: Service
+needs-hiero-approval: Yes
+needs-hedera-review: Yes
+status: Draft
+created: 2026-04-29
+updated: 2026-06-22
+requires: 18, 336, 540, 904, 1010, 1261
+replaces:
+superseded-by:
+release:
+---
+
+## Abstract
+
+This HIP adds `PeriodicHoldingFee` (PHF), a new top-level configuration on fungible Hedera Token Service (HTS) tokens that allows an issuer to periodically debit either a percentage of each non-exempt holder's balance (**fractional mode**) or a flat amount per holder (**fixed mode**), denominated in the token itself, and credit a configured collector account. PHF is time-scoped: a holder owes for the whole cycles elapsed since the relationship's last accrual reset (`last_assessed_at`), whether or not they have transferred.
+
+PHF can be attached to a token only at `TokenCreate`. The mode, the per-cycle cap (basis points in fractional mode, smallest-unit amount in fixed mode), the assessment cadence, and the first assessment time are immutable for the life of the token; the current rate (within the cap), the collector, and the exempt-account list are mutable. Assessment is explicit and signed: an authorized keeper submits `assessTokenPeriodicHoldingFee` transactions against batches of holders, and the balance-changing operations enumerated in this specification perform touch settlement atomically before proceeding. Each settlement is exempt from the token's custom-fee schedule, so a holder's debit is exactly the assessed amount — at most the cap multiplied by the whole cycles elapsed and never more than the balance. No holder registry, no autonomous network work, no cycle-boundary processing, no HBAR reserve, and no per-cycle history is introduced; the only per-holder state is one timestamp on the token relationship HTS already maintains.
+
+## Motivation
+
+Some fungible HTS tokens represent claims with ongoing off-ledger costs — custody, insurance, audit, administration, or regulatory reporting — borne by holders of record rather than only by holders who transfer. Existing custom fees are assessed only during transfers and therefore never reach passive holders.
+
+Without a native mechanism, issuers must use off-ledger billing, smart-contract overlays, or wipe-and-remint flows. Those approaches are hard to audit, do not work uniformly for self-custody, and require operational keys inappropriate for routine fee collection. PHF provides a native, transparent, capped mechanism for periodic holder-paid fees while preserving HTS association, transfer, freeze, pause, custom-fee, and treasury semantics.
+
+## Rationale
+
+- PHF is a top-level token feature with two assessment paths. Explicit assessment (Path A) is issuer-driven and requires a signature by a dedicated `holdingFeeKey`: the protocol provides the primitive — debit a holder under declared, immutable, capped terms — and the issuer (or any authorized keeper) provides scheduling, holder enumeration, and batching off-chain. Touch settlement (Path B) is the minimal residue that remains in consensus: it closes the evasion loop atomically with the operations that move a holder's balance, exactly as transfer-triggered custom fees already do, and performs no scheduling of its own.
+- PHF supports two complementary economic models, mirroring HTS custom fees (`FractionalFee` and `FixedFee`): **fractional mode** charges a percentage of the holder's balance per cycle; **fixed mode** charges a flat amount per holder per cycle. Per-cycle math is integer-only in both: `(balance × current_per_cycle_basis_points × cycles_owed) / 10000` for fractional, `current_per_cycle_amount × cycles_owed` for fixed — one multiplication and one integer division, with no exponentiation, no rational-number normalization, and no time-unit conversion constants in consensus, so every node computes byte-identical results. Annualized equivalents are derivable by Mirror Node and SDKs and are presentation-only.
+- Accrued cycles are settled together, valued **linearly** at the holder's balance and the current rate in effect at settlement time. Linear valuation keeps consensus math to one multiplication and one division per settlement. Per-cycle compounded valuation is deliberately not used: it would require an iterative loop or fractional exponentiation in consensus, neither of which can be guaranteed to evaluate byte-identically across heterogeneous node hardware and runtimes, and so would put consensus at risk. The clamp to the holder's balance bounds the worst case (see "Per-Cycle Math"). Credit-side touch settlement keeps that valuation honest: a credit carried by the triggering operation's transfer list settles accrued cycles at the pre-credit balance first, so ordinary deposits never inflate previously accrued cycles.
+- The mode and its cap are immutable for the life of the token (`max_per_cycle_basis_points` in fractional mode, `max_per_cycle_amount` in fixed mode). The current rate within the established mode (`current_per_cycle_basis_points` or `current_per_cycle_amount`) is mutable strictly within `[0, cap]`. A token cannot switch modes via `TokenUpdate`. The current rate is kept as protocol state — rather than supplied per assessment by a calling contract — because PHF targets bare HAPI tokens with no contract above the line; a single mutable scalar bounded by an immutable cap is the smallest mechanism that lets issuers move pricing without a schedule entity or rate history.
+- PHF keeps no list of holders and schedules no autonomous work. The only per-holder state is a single `last_assessed_at` timestamp on the token-relationship record HTS already maintains for every associated account; there is no holder registry, and no state survives for ex-holders, because dissociating (which requires a zero balance) removes the relationship and its timestamp together. A keeper enumerates current holders off-ledger and submits them to `assessTokenPeriodicHoldingFee` as an advisory batch; consensus never walks the holder set on its own. The per-holder fee component (HIP-1261) charges the cost of processing each submitted holder, so enumeration and iteration are paid for by the party that requests them. Accrual is tracked per relationship rather than once per token because holders associate, transfer, freeze, lose KYC, and become exempt at independent times, and each of those must advance or reset accrual for that holder alone (see "Rejected Ideas").
+- A PHF settlement is a holder → collector transfer of the assessed amount that is exempt from the token's custom-fee schedule. The exemption is a dedicated rule triggered by the transfer being a PHF settlement — not by the PHF collector belonging to the token's custom-fee collector set — so it is independent of `all_collectors_are_exempt` (HIP-573) and of the token's custom-fee configuration. Exemption bounds the holder's exposure to exactly the declared PHF terms — at most `cap × cycles_owed`, never more than the balance — and prevents a custom fee (for example a fixed HBAR fee) from debiting a holder on a settlement the holder may not have signed, since Path A settlements are triggered by the keeper rather than the holder.
+- Authorization for PHF lives in a dedicated `holdingFeeKey`.
+
+## User stories
+
+- As a token issuer, I want a capped periodic fee set at token creation so holder-funded operating costs can be collected without off-ledger billing or routine wipe authority.
+- As a token holder, I want to see the rate, the immutable cap, the cadence, the collector, and the token's custom fee schedule on-ledger before associating, and I want a guarantee that a PHF settlement can never debit more than the cap multiplied by the whole cycles elapsed, never exceed my balance, and never create a debt.
+- As an issuer's billing service (keeper), I want a single HAPI operation to assess a batch of holders, with the network refusing assessments that fall outside the declared terms, and a cost that scales with the holders processed so oversized or spam batches are uneconomic to submit.
+- As an auditor or compliance officer, I want every PHF assessment to produce an on-ledger record carrying the cycles covered, the applied rate, the pre-assessment balance, the resulting amount, and the collector at settlement time, so each debit is verifiable from the record itself and the issuer's fee collection can be reconciled against off-ledger disclosures.
+- As a wallet, exchange, or integrator, I want PHF metadata in `TokenInfo` and indexed in Mirror Node before exposing the token to my users, so balances, spendable balances, and projected fee exposure are displayed accurately.
+
+## Specification
+
+### Terminology
+
+A **cycle** is a period of `min_seconds_between_assessments` seconds. Cycles accrue per holder relationship starting at the relationship's `last_assessed_at`. `min_seconds_between_assessments` is therefore both the minimum spacing between two consecutive assessments of the same holder and the accrual quantum that multiplies the per-cycle rate.
+
+### Scope
+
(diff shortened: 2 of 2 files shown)


mgarbs (commented):
Thanks for a thorough and clearly-written proposal — the per-relationship `last_assessed_at` state minimization, the integer-only linear valuation, and the touch-settlement atomicity are well reasoned. I've left my detailed questions inline on the relevant sections; one point belongs at the document level rather than any single line:

A HIP does its job best as a tight, single-purpose design doc, and this one spans a lot of surface — a new HAPI op, touch settlement woven through the transfer paths, key lifecycle, association/dissociation, freeze/KYC/pause interactions, Mirror Node, and SDK. For a concrete sense of scale, at ~11,600 words this is the second-longest HIP in the repository (of 157), behind only HIP-1056 "Block Streams" — a network-wide, multi-subsystem infrastructure change. For a single new token fee type to approach that length suggests the scope or the prose could be tightened, or the proposal split into smaller, focused HIPs (e.g. the core fee primitive separate from the touch-settlement and lifecycle interactions). Length on this scale can work against getting the design read and understood.

HIP/hip-1500.md
```diff
@@ -0,0 +1,652 @@
...
+- allowance-based transfers where the allowance covers only the requested amount, including failure when the post-settlement balance no longer covers it; full-balance transfers failing while a cycle is accrued;
+- `TokenAirdrop` immediate-transfer and pending-airdrop classification, `TokenClaimAirdrop` sender-side settlement, `TokenCancelAirdrop` non-trigger; `TokenWipe` non-trigger with cycles continuing to accrue;
+- key lifecycle: PHF created without `holdingFeeKey` operating in touch-settlement-only mode; key removal via the HIP-540 sentinel being permanent; `TOKEN_HAS_NO_HOLDING_FEE_KEY` on assess and mutations thereafter;
+- HIP-904 auto-association `last_assessed_at` initialization; deterministic skip-code reporting for non-existent / deleted / expired / not-associated holders; pause accrual followed by post-unpause settlement.
+
+## Rejected Ideas
+
+- **Off-ledger billing, wipe-and-remint, smart-contract-only periodic fees, rebasing supply**: cannot reliably reach self-custody, contract, or DeFi-held balances; require operational keys inappropriate for routine fee collection; or change token accounting semantics for downstream integrations.
```
mgarbs, line 609: I'd like to understand where the smart-contract path actually falls short here. A contract-controlled token implementing the issuer's exact debit rules, driven by self-rescheduling long-term scheduled calls (HIP-423), looks like it could meet this without consensus-layer changes, a new HAPI op, or new per-relationship state — what does it fail to do? And is it clear why existing token authorities (e.g. wipe) can't be composed to the same end, since that's the justification given here for a new `holdingFeeKey`?

HIP/hip-1500.md
```diff
@@ -0,0 +1,652 @@
...
+
+## Motivation
+
+Some fungible HTS tokens represent claims with ongoing off-ledger costs — custody, insurance, audit, administration, or regulatory reporting — borne by holders of record rather than only by holders who transfer. Existing custom fees are assessed only during transfers and therefore never reach passive holders.
```
mgarbs, line 27: How does PHF behave when the token is held by an AMM pool or lending contract? There the contract is the holder of record, so touch settlement would debit the pool rather than the underlying users, and every pool address would need exempting in advance. Does that leave PHF reaching "DeFi-held balances" any differently than the smart-contract approach it's contrasted against?

HIP/hip-1500.md
```diff
@@ -0,0 +1,652 @@
...
+
+- A `CryptoTransfer` from a non-exempt holder of a PHF-bearing token — and one crediting a non-exempt holder with accrued cycles — incurs additional consensus work and additional internal transfers (touch settlement, debit-side and credit-side) before the user's transfer is applied, paid by the payer of the triggering transaction as HIP-1261 line items. Fee estimation, gas estimation for HSCS precompile callers, and throughput expectations MUST account for this, and integrators should expect settlement records on incoming transfers, not only outgoing ones.
+- A holder's effective transferable balance is `balance − accrued settlement`: a transfer of the full stored balance fails whenever at least one cycle has accrued with a non-zero amount. This breaks "send max" flows in wallets and sweep patterns used by exchanges and custodians. Integrators SHOULD compute spendable balance via Mirror Node or the SDK helper before constructing transfers.
```
mgarbs, line 558: I'm trying to gauge the composability cost of this. Two accounts holding the same token can have different spendable balances, and transfers can debit mid-operation — which lands on every wallet, exchange, custodian, and AMM touching any PHF token, not just the issuers who opt in. Is that blast radius weighed against the benefit anywhere?

HIP/hip-1500.md
```diff
@@ -0,0 +1,652 @@
...
+The following are **network-level parameters**, defined by network governance through node configuration and uniform across all tokens; they are not declared per token by the issuer. The "Initial value" column gives the proposed starting value, which governance MAY tune. Per-token values an issuer chooses at `TokenCreate` (the caps, the current rate, the cadence, `first_assessment_at`, the collector, and the exempt list) are listed in "Periodic Holding Fee Configuration" and are distinct from these.
+
+| Parameter | Initial value |
+| --- | ---: |
+| `MIN_ASSESSMENT_INTERVAL_SECONDS` | 86,400 |
+| `MIN_FIRST_ASSESSMENT_DELAY_SECONDS` | 604,800 |
+| `MAX_EXEMPT_ACCOUNT_IDS` | 64 |
+| `MAX_EXEMPT_LIST_DELTA_PER_TX` | 16 |
+| `MAX_ASSESS_BATCH_SIZE` | 100 |
+| `MAX_PER_CYCLE_BASIS_POINTS_CEILING` | 10,000 |
```
mgarbs, line 540: What's the reasoning behind a 100%-of-balance-per-cycle ceiling here? With the current rate mutable up to the cap and rate changes effective immediately, buyer-beware on the cap plus `TokenReject` are the only holder protections. Would a lower network ceiling — or a prospective-effect window for increases — be worth weighing?

HIP/hip-1500.md
```diff
@@ -0,0 +1,652 @@
...
+- **Auto-exempting custom-fee collectors via a flag on PHF**: PHF semantics are independent of custom-fee semantics; a flag reaching into the custom-fee schedule would introduce cross-feature coupling and force consensus to walk the custom-fee list at every assessment. Issuers enumerate the exempt set explicitly in `exempt_accounts`.
+- **Fees denominated in HBAR or another token**: introduces recurring insufficient-balance and allowance problems outside the held token.
+- **PHF for NFTs**: holding semantics for unique tokens require a separate design.
+
+## Open Issues
+
+- Final protobuf tag and enum-ordinal assignments must be made by protobuf maintainers, including the `SchedulableTransactionBody` entry and `HederaFunctionality` value.
+- HIP-1261 fee-schedule line items — the per-processed-holder component and base fee of `assessTokenPeriodicHoldingFee`, and the touch-settlement line item charged to the triggering transaction's payer — should be validated with Hedera governance and testnet benchmarks.
```
mgarbs, line 628: Does a cycle stay net-positive when a long tail of small holders each settle to zero (`ZERO_ASSESSMENT_AFTER_ROUNDING`) while the issuer still pays a per-holder fee to process them? Since this line item is still open and the model depends on it — and since Path A still needs an off-ledger keeper to enumerate and batch — what's the net gain over existing primitives beyond the atomic in-consensus debit and the capped-exposure guarantee?

HIP/hip-1500.md
```diff
@@ -0,0 +1,652 @@
...
+- Throttle policy specific to `assessTokenPeriodicHoldingFee` (whether it shares the existing token-service throttle bucket or has a dedicated one) should be confirmed with consensus-node maintainers.
+- Mirror Node consistency guarantees for keeper holder-enumeration queries (lag bounds, pagination semantics) should be documented in Mirror Node specifications.
+- The precise EVM log surface when touch settlement is triggered via precompile (events on the calling contract's behalf) should be confirmed with HSCS and Mirror Node implementers.
+- HIP-551 atomic batches: whether inner transactions trigger touch settlement individually, and how Path B atomicity composes with whole-batch revert, should be specified with consensus-node maintainers.
```
mgarbs, line 632: A few of these deferred items look determinism-critical rather than cosmetic — HIP-551 batch composition with touch-settlement atomicity, throttle policy, and the precompile EVM log surface. Do those need to settle before the HIP advances, given any of them could change the design?

AdrianKBL:
> Thanks for a thorough and clearly-written proposal — the per-relationship `last_assessed_at` state minimization, the integer-only linear valuation, and the touch-settlement atomicity are well reasoned. I've left my detailed questions inline on the relevant sections; one point belongs at the document level rather than any single line:
> 
> A HIP does its job best as a tight, single-purpose design doc, and this one spans a lot of surface — a new HAPI op, touch settlement woven through the transfer paths, key lifecycle, association/dissociation, freeze/KYC/pause interactions, Mirror Node, and SDK. For a concrete sense of scale, at ~11,600 words this is the second-longest HIP in the repository (of 157), behind only HIP-1056 "Block Streams" — a network-wide, multi-subsystem infrastructure change. For a single new token fee type to approach that length suggests the scope or the prose could be tightened, or the proposal split into smaller, focused HIPs (e.g. the core fee primitive separate from the touch-settlement and lifecycle interactions). Length on this scale can work against getting the design read and understood.

Thanks, the length is a fair concern and we agree the document should be easier to read end to end. Two things are inflating it that we can cut without losing normative content: the Reference Implementation test enumeration and the Response Codes section restate rules already fixed in Specification, and How to Teach overlaps with the Rationale. We can compress those materially.

On splitting: we considered it, and the coupling is the catch. Touch settlement (Path B) is what makes the fee non-evadable, so a "core primitive" HIP that shipped Path A only would be a weaker, separately-reviewable design rather than a clean subset. If the editors prefer a split, the natural cut is (1) the fee primitive: configuration, per-cycle math, Path A explicit assessment, and the state model, and (2) touch settlement plus the lifecycle interactions (association/dissociation, freeze/KYC/pause, TokenReject).

HIP/hip-1500.md
AdrianKBL: Good question. The gap is authorization, not automation.

A contract can only debit a third party's balance two ways, and neither reaches the target:


**1. Allowance.** Every holder would have to have granted the fee contract an allowance. Self-custody holders, P2P recipients, and AMM/lending pools that received the token by an ordinary transfer never granted one, so they are unreachable. Reaching passive holders of record is the entire objective (Motivation), so an allowance-gated debit does not meet it.

**2. Wipe Key.** This fails on regulatory and trust grounds before the accounting even matters.

Regulatory and custody posture (the decisive point). Routine fee collection this way forces the issuer to hold, and an automated keeper to exercise, a key that can zero any holder's entire balance at will. For an asset-backed token, the holder's balance representing a non-confiscatable claim on the underlying asset is central both to how the token is characterized and to why anyone holds it. Coupling ordinary periodic billing to confiscation-grade authority is not a posture an asset-backed or regulated issuer can take, and it is not delegable to a fee keeper under any reasonable least-privilege or key-management standard. (Whether an issuer separately elects to hold a dormant compliance key is a different decision; PHF neither requires nor precludes it. The objection is to making that power a prerequisite for billing).

Holder trust and adoption. An active wipe key is a widely recognized red flag: many holders and integrators screen out tokens whose issuer can wipe balances, which suppresses exactly the self-custody and DeFi holding PHF is meant to reach.
Accounting and mechanics. Even setting those aside, TokenWipe burns and reduces total supply, it does not move value to a collector. Collecting via wipe therefore means wipe-and-remint, which transiently changes total supply (breaking supply-as-reserve accounting for asset-backed tokens and any supply-based downstream integration). That is the wipe-and-remint idea already rejected, and it is why TokenWipe is explicitly not a touch-settlement trigger (Rejected Ideas: "TokenWipe as a touch-settlement trigger").





A token implemented purely as a contract (ERC-20-style) sidesteps both, but then it is not an HTS token: it forgoes HTS association, native custom fees, native Mirror Node indexing, and HTS-native DEX pools, and downstream integrations must special-case it. That is the "smart-contract-only" / "change token accounting semantics for downstream integrations" case in Rejected Ideas.

**HIP-423** does not close the gap. Long-term scheduling extends the signature-collection window (up to 62 days), it does not grant new authority. A scheduled CryptoTransfer that debits a holder still needs that holder's signature or an allowance, so self-rescheduling automates submission, not permission, and each schedule is bounded by the 62-day horizon.

On "no new per-relationship state": the comparison runs the other way. A contract reproducing PHF's time-scoped accrual (owe for whole cycles since the last reset, correctly handling association, freeze, KYC revocation, pause, and exemption at independent times per holder) has to maintain its own holder registry and per-holder timestamps in contract storage, and mirror relationship state it does not own and cannot observe natively. That is strictly more state than PHF's single last_assessed_at on the relationship record HTS already keeps. See "A single token-level `last_assessed_at` instead of one per relationship" for why per-holder accrual is required at all.

Why a dedicated key rather than composing wipe: `holdingFeeKey` grants exactly one capability, debit a holder under declared, immutable, capped, custom-fee-exempt terms to a fixed collector. It cannot mint, cannot alter supply, and cannot zero a balance outside those terms. Composing wipe would grant strictly more power than the operation needs (least privilege) and produce burn semantics where transfer semantics are required. Crucially, this lets an issuer collect fees without ever holding wipe authority, decoupling routine fee collection from confiscation-grade power, which is precisely what an asset-backed or regulated issuer needs. Reusing `feeScheduleKey` is rejected on the same principle (scope creep into an unrelated authority domain). The holder consents by associating to a token whose PHF cap and cadence are immutable and visible on-ledger before association.

Net: the contract path can automate the submission, but it cannot obtain authorization to debit a non-participating holder without either the wipe key (wrong power, wrong accounting) or a universal allowance (unreachable for self-custody and DeFi). PHF's primitive is the minimal authority that composes with native token state.

HIP/hip-1500.md
AdrianKBL: PHF debits the holder of record, so for a pooled position that is the pool or lending contract, not the underlying users and PHF does not disaggregate pooled positions. No per-balance mechanism does.

The contrast we intended is narrower than the current wording suggests. A native primitive can reach a balance wherever it sits, including a contract-held balance, without that holder's cooperation (Path A is signed by holdingFeeKey), which an allowance-based contract cannot do. It is not a claim that PHF pushes incidence down to LPs. In practice debiting a pool silently perturbs its internal accounting (constant-product reserves, share pricing), so issuers should exempt pool and lending-market addresses via exempt_accounts and handle end-user incidence through the integration or off-ledger, as with any transfer-time fee today.

HIP/hip-1500.md
AdrianKBL: PHF is opt-in per token at TokenCreate and cannot be added later, so no existing token changes behavior, and PHF presence is discoverable in TokenInfo and Mirror Node before association. The surface is limited to tokens that declare PHF, and integrators can detect it programmatically.

Spendable balance differing from stored balance is not itself new: fractional custom fees already require computing a net-of-fee amount for "send max" today. What is new is the timing, settlement on incoming credits and mid-operation, which we call out in Backwards Compatibility, with the mitigation being the Mirror Node / SDK spendable-balance helper.

HIP/hip-1500.md
AdrianKBL: The network ceiling (`MAX_PER_CYCLE_BASIS_POINTS_CEILING` = 10,000) only excludes encodings above 100% per cycle. It is deliberately not the holder protection. The binding protection is the per-token immutable cap chosen at creation and visible before association, plus the balance clamp (no negative balance, no debt) and TokenReject. We avoid a lower network-imposed economic ceiling because legitimate cadence and rate combinations vary, and the immutable per-token cap is the correct place to bound exposure (Parameters already states issuers SHOULD declare a cap that meaningfully bounds it). If governance prefers a more conservative outer bound, the ceiling is a network parameter and can be lowered without affecting the mechanism.

On a prospective-effect window for increases: this is evaluated and rejected in Rejected Ideas ("Prospective effect for rate increases"). A pending (value, effective_at) pair adds pending-rate state and a second effective-rate lookup on every assessment, while the immutable cap already bounds exposure and any outgoing transfer settles accrued cycles at the terms in effect at that moment.

HIP/hip-1500.md
AdrianKBL: Agreed, and thanks for separating these from the cosmetic ones. **HIP-551** batch composition (whether inner transactions each trigger touch settlement, and how Path B's atomic unit composes with whole-batch revert under BatchKey) and the precompile EVM log surface are design-affecting, not editorial.

Action: we will promote both from Open Issues into normative Specification before Last Call, with proposed semantics for review. Proposed: each inner transaction's touch settlement is part of that inner transaction's atomic unit and reverts with it, a whole-batch revert reverts all inner settlements, and the EVM log surface follows the existing HTS-to-EVM CryptoTransfer mapping. Final wording to be confirmed with consensus-node and HSCS maintainers.

Throttle policy is different in kind: it is a network-tuning parameter (which bucket assessTokenPeriodicHoldingFee uses) that does not change transaction semantics or determinism, so it can stay an operational item to confirm with consensus-node maintainers. We will state that distinction explicitly so it is not conflated with the design-affecting items.

HIP/hip-1500.md
AdrianKBL: Net-positivity per cycle is an issuer economic decision, not a protocol-correctness property. Because the per-holder component (**HIP-1261** base-plus-extras model) makes each processed holder cost something, the keeper is incentivized to assess only holders whose expected amount exceeds the per-holder cost and to process dust holders less frequently. That per-holder charge exists precisely so oversized or unproductive batches are uneconomic (Rationale, Cost). PHF does not require assessing every holder every cycle.

On net gain: the atomic in-consensus debit and the capped-exposure guarantee are already substantial, and PHF adds several properties that no composition of existing primitives provides together:


supply-neutral collection (no wipe-and-remint, so total supply keeps tracking reserve),
a least-privilege dedicated key,
reachability of self-custody holders with no allowance,
native exemption from the custom-fee schedule, bounding holder cost to exactly the PHF amount,
native discoverability in TokenInfo and Mirror Node,
touch settlement closing the evasion loop, so a holder cannot avoid the fee by never being assessed.


The keeper still enumerates and batches, but no keeper can perform the debit itself without wipe or a universal allowance, which is the authorization gap PHF closes.

PR #1507, opened 2026-06-24
Add source-aware HIP edit flow

mgarbs:
## Summary
- add source metadata for merged and PR-backed HIPs, including PR head/base branch details
- replace the static edit link with a local markdown editor that loads full source frontmatter and preview
- submit edits through a fork-first flow that targets the existing HIP PR branch when possible and falls back to main
- remove the manual GitHub token field entirely and use a Connect GitHub integration hook
- remove obsolete Netlify config from the repo; the site is deployed by GitHub Pages

## GitHub Connect
- GitHub Pages builds can set `VITE_GITHUB_AUTH_START_URL` from the `HIPS_GITHUB_AUTH_START_URL` repository variable
- production can also provide a runtime `window.HIPS_GITHUB_TOKEN_PROVIDER`
- if no provider is configured, the editor stays usable for copy/download/open-on-GitHub but does not ask users for tokens
- provider implementation, hosting, and secret handling are intentionally outside this PR and should be reviewed separately

## Testing
- `npm run build`
- `git diff --check`

### DIFF
--- a/.github/workflows/deploy-site.yml
+++ b/.github/workflows/deploy-site.yml
@@ -52,6 +52,7 @@ jobs:
         run: npm run build
         env:
           VITE_BASE: ${{ steps.pages.outputs.base_path }}/
+          VITE_GITHUB_AUTH_START_URL: ${{ vars.HIPS_GITHUB_AUTH_START_URL }}
 
       - name: Upload artifact
         uses: actions/upload-pages-artifact@fc324d3547104276b827a68afc52ff2a11cc49c9 # v5.0.0
--- a/.github/workflows/update-draft-hips.yml
+++ b/.github/workflows/update-draft-hips.yml
@@ -83,7 +83,22 @@ jobs:
                       title
                       number
                       url
+                      baseRefName
                       headRefOid
+                      headRefName
+                      headRepository {
+                        name
+                        owner {
+                          login
+                        }
+                      }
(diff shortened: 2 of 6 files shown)


.github/workflows/deploy-site.yml
```diff
@@ -52,6 +52,7 @@ jobs:
         run: npm run build
         env:
           VITE_BASE: ${{ steps.pages.outputs.base_path }}/
+          VITE_GITHUB_AUTH_START_URL: ${{ vars.HIPS_GITHUB_AUTH_START_URL }}
```
leninmehedy, line 55: We should ensure OAuth broker (the service at VITE_GITHUB_AUTH_START_URL) explicitly requests ONLY the public_repo (not admin:org, or delete_repo etc.). 

Even if current OAuth broker asks for sensitive permissions, user can still review and reject. But If we request unnecessary scopes, sensitive users may reject and may not use this feature.

This is a nitpick, can be checked/handled separately.

leninmehedy (dismissed):
A minor security hardening comment, otherwise LGTM. Great feature. Thanks!

.github/workflows/deploy-site.yml
leninmehedy: This is a minor and nitpick, can be addressed separately. No need to block this PR..

PR #1522, opened 2026-07-22
HIP: Account Staking Configuration via the Hedera Account Service

stoqnkpL:
The Hedera Account Service (HAS) system contract at address `0x16a`, introduced in [HIP-632](https://github.com/hiero-ledger/hiero-improvement-proposals/blob/main/HIP/hip-632.md) and extended in [HIP-906](https://github.com/hiero-ledger/hiero-improvement-proposals/blob/main/HIP/hip-906.md), lets accounts and contracts manage authorization and HBAR allowances from within the EVM. It does not expose an account's **staking configuration** — the `staked_node_id` / `staked_account_id` and `decline_reward` fields that today are settable only through the HAPI `CryptoUpdate` and `ContractUpdate` transactions. As a result a smart contract that custodies HBAR cannot direct its own balance to stake to a consensus node, or decline staking rewards, without an off-chain HAPI transaction signed by an admin key it may deliberately not hold. This HIP adds staking-configuration functions to the `IHederaAccountService` interface and its `IHRC632` account-address facade, so an account (EOA or contract) can set or clear its own staking target and reward preference directly from the EVM, gated by the existing HAS security model. The change is purely additive: no existing HAPI transaction, protobuf message, or system-contract selector is altered.


Reccetech:
@stoqnkpL Thanks for putting together this HIP. The Hashgraph Smart Contract team is still reviewing this HIP and will come back with suggestions, but in the meantime as promised I wanted to share some suggestions for discussion. 

### 1. Address the Contract Self-Configuration Blocker

In the codebase neither available dispatch path works for a contract updating its own staking configuration:

- `CryptoUpdateHandler` unconditionally rejects any account with `smartContract = true` with `INVALID_ACCOUNT_ID` — no bypass exists for system contract dispatches.
- `ContractUpdateHandler` throws `MODIFYING_IMMUTABLE_CONTRACT` in both `preHandle` and `validateSemantics` for any contract holding a self-referential `CONTRACT_ID` key — which is every standard EVM-deployed contract.

Both paths are independently blocked for the exact accounts HIP-1522 is designed to serve.

The HIP could address how this will be resolved. Maybe a targeted solution would be to condition the `smartContract` guard in `CryptoUpdateHandler` to allow staking-field-only updates dispatched from the system contract, since contracts are accounts in Hedera state and the staking fields are shared.

### 2. Reconsider `updateStakingConfig` Scope

`updateStakingConfig(int64 nodeId, bool decline)` (and its `IHederaAccountService` counterpart `updateStakingConfig(address account, int64 nodeId, bool decline)`) only covers node staking combined with reward preference — it has no equivalent for account staking, and per the current spec table a negative `nodeId` is used to clear the target (so it doubles as `unstake`, but still cannot express "stake to account X and set decline"). The name implies a complete combined setter but the function is partial.

The HIP could either expand the function to cover the account-staking case, rename it to reflect its actual scope (e.g., `stakeToNodeAndSetDeclineReward`), or explicitly acknowledge the limitation as a deliberate design choice.

### 3. Read Functions for Staking State

HIP-1522 adds write-only staking functions. It does not add any corresponding read functions, leaving a contract with no on-chain way to query the staking state it is managing. HAPI exposes the following staking fields via `CryptoGetInfo` that have no EVM equivalent today:

| Field | Meaning |
| --- | --- |
| `staked_node_id` | Which consensus node the account is staked to |
| `staked_account_id` | Which account the account is staked through |
| `decline_reward` | Whether the account has opted out of staking rewards |
| `pending_reward` | Accrued but not yet distributed reward (in tinybar) |
| `stake_period_start` | Start of the account's current active staking period |
| `staked_to_me` | Total balance of other accounts staking through this account |

Note: the current PR text already lists this under `Open Issues`.

The absence of read functions creates practical problems for the use cases the HIP is designed to enable. A governance or custody contract managing staking on behalf of users cannot:

- Verify the current staking target before deciding whether to switch nodes.
- Display or act on pending reward amounts without relying on off-chain mirror node data.
- Confirm that a prior `unstake()` call succeeded by reading back the cleared state.

`pending_reward` in particular is relevant to treasury and DeFi contracts that want to account for accrued staking yield on-chain, and it is notably absent from the HIP's `Open Issues` list of deferred read fields (`staked_node_id` / `decline_reward` / `stake_period_start`).

An established pattern exists in HAS for read functions: `hbarAllowance`, `getEvmAddressAlias`, `getHederaAccountNumAlias`, and `isValidAlias` all read directly from account state without dispatching a transaction and use a fixed `viewGasRequirement()` (100 gas floor) rather than dispatch-based pricing. Staking reads could follow exactly the same pattern.

Comments and feedback appreciated.

stoqnkpL:
@Reccetech Thanks — all three addressed in the commits above. 
1. `CryptoUpdate` is now dispatched for every target with a new Contract accounts section specifying the guard change.
2. `updateStakingConfig` is renamed to `stakeToNodeAndDeclineReward`.
3. `getStakingInfo` returns the six fields in one struct.

Two things for your team:

1. `SELF_STAKING_IS_NOT_ALLOWED (321)` and `STAKING_NOT_ENABLED (323)` are never returned by the node - they exist only in response_code.proto and the generated Solidity, and StakingValidator doesn’t reject self-staking at all. Should the HIP drop them, or should the node reject self-staking?
4. HIP-904’s `setUnlimitedAutomaticAssociations` may already hit the same guard in `CryptoUpdate` - identical dispatch shape, so a contract calling it on itself looks like it fails with `INVALID_ACCOUNT_ID`. Its e2e tests only cover an EOA and its unit tests stub the dispatch, so nothing would have caught it.

PR #1535, opened 2026-08-19
HIP-1535: CLPR

rbair23:


0xsims:
Read this as an operator running production attestation anchoring on HCS. Two questions from that vantage, both about applications that retain CLPR-delivered records rather than consuming them:

1. §9’s recovery table notes R5/R6 may leave messages UNRESOLVED, and §8.7 handles migration by opening a new Channel and closing the old. For message delivery this is complete. But for an application retaining CLPR-delivered records as durable evidence: once a Channel is CLOSED and its verifier deprecated, is a record verified under that Channel still independently re-verifiable afterward, or does verifiability end with the Channel lifecycle? A sentence in §8.7 on whether history survives migration would help anyone building an audit/retention layer.

2. The listed reference verifiers all authenticate a peer ledger’s consensus state (TSS, BLS, QBFT, Sei validator sets). Is there any intended scope for verifiers that authenticate application-level committed data (e.g. an attestation or document commitment proven into a ledger’s state), or is that firmly an application-layer concern above CLPR? Asking to understand the boundary, not to widen it.

edward-swirldslabs:
@0xsims, 

The CLPR Application API completely hides anything related to state proofs.  The state proofs are purely to attest to the state of the outbound message queues and what has been delivered and handled by the CLPR Services on either end of a channel.  

> 1. is a record verified under that Channel still independently re-verifiable afterward, or does verifiability end with the Channel lifecycle? A sentence in §8.7 on whether history survives migration would help anyone building an audit/retention layer.

The term `record` is not used in the HIP or CLPR Specification anywhere, so I am inferring your meaning to be bundles or messages.  Communication between CLPR Services along channels happens in `bundles`.  Communication between CLPR Applications happens in `messages`.  State proofs attest to the content of the bundles and are set aside once the verifier contract has validated the state proof.  The CLPR Service only gets the bundle payloads, including messages, and trusts that the channel was configured with the correct verifier contract.  Similarly, the application receiving a message intrinsically trusts the message was sent from the claimed source when received.   

The application has to be pre-configured with what channels to send and receive on.  It is that configuration to trust messaging on the channel that implies the application writer has vetted the CLPR Services involved and Verifier Contracts to be trust worthy. 

The history is in the block chain's transaction logs, event logs, and state changes.  CLPR, at a specification level, does not do anything more to manage or curate additional history of interaction on a channel.  The sequence of messages across a channel has a running hash which establishes order and that can be used to verify content of those messages if you know a preceding hash and the intervening messages. 

> 2. The listed reference verifiers all authenticate a peer ledger’s consensus state (TSS, BLS, QBFT, Sei validator sets). Is there any intended scope for verifiers that authenticate application-level committed data (e.g. an attestation or document commitment proven into a ledger’s state), or is that firmly an application-layer concern above CLPR? Asking to understand the boundary, not to widen it.

Channel setup is permissionless.  Anyone can write a custom verifier contract, setup a custom channel between two CLPR Services using the custom verifier contract, bind their own Connector and applications on either end and send messages back and forth. That custom verifier contract receives state proven bundles and can do whatever else it wants during the `verifyBundle` call, including emit events, keep a record of messages sent between applications on the channel, call other smart contracts, or whatever other side effects are desired.   This is a custom setup, however.  This is not a way to inject side effects on arbitrary channel communication.  If what you are wanting to do is track on-chain, then you can only do it with the applications that agree to use the channel setup with your custom verifier contract.  The difficulty with curating a message history on-chain is that the state is going to balloon in size with permanent storage of the messages.  

> Read this as an operator running production attestation anchoring on HCS. Two questions from that vantage, both about applications that retain CLPR-delivered records rather than consuming them:

If you are wanting to track arbitrary channel communication and arbitrary message histories, this is better done off-chain through monitoring the blockchain output or block streams and then feeding back to HCS from off-chain.  You can also construct state proofs from the block stream history (once TSS is released).  

> But for an application retaining CLPR-delivered records as durable evidence: once a Channel is CLOSED and its verifier deprecated, is a record verified under that Channel still independently re-verifiable afterward

There is nothing yet in the CLPR specification that associates one channel's message history with another.   That is left to human coordination.  

Is tracking the Block Stream off-chain for the state changes you are wanting to monitor sufficient?

0xsims:
Thanks Edward. Terminology confirmed: I meant messages/bundles — "record" on my end is a message an application retains as durable evidence after delivery.

To answer your question directly: yes, block stream tracking off-chain is sufficient for our model. It's what we run on Hedera today — observe, sign, anchor to HCS. And CLPR staying out of history curation is the right call from where I sit. Retention is our layer, not yours.

One clarifying question on the running hash, since it determines how an evidence layer gets built: is the channel running hash exposed on-chain (events or queryable state), or does it live inside the CLPR Service? If it's on-chain, a non-participant observer can checkpoint it straight from the block stream. If it's service-internal, checkpointing needs participant cooperation, which changes the trust story for anyone relying on the evidence later.

Either way, the retention pattern we'd build: while a channel is live, an observer periodically anchors (channel id, sequence number, running hash) to an external ledger. After teardown, anyone replays history against those checkpoints and re-verifies any message without trusting the original participants. The case where this earns its keep is a private network — when a Sphere is decommissioned at the end of a business relationship, its logs and state go with it, but externally anchored checkpoints survive to make the surviving copies of the history independently verifiable.

We plan to build this as a reference implementation (CLPR channel observer, checkpoints anchored to HCS, post-quantum signed). If useful, a sentence in §8.7 noting that post-lifecycle re-verifiability depends on externally retained running-hash checkpoints would save future implementers the archaeology.

